Issue #5, opened 2026-08-04
Prove the zellij launch seam


blooop:
## Question

Prove the zellij launch seam: from a full-screen TUI that currently owns the terminal, start a HITL agent session on a chosen ticket in a zellij pane/tab, in the right project directory, and return control sanely.

Concretely, prototype and answer:

- `zellij action new-pane` / `new-tab` from inside a running TUI — does focus hand off cleanly, and what happens when `wf` runs *outside* any zellij session (fall back to `exec`? refuse?)
- naming: pane/tab per ticket so a claimed ticket's session is findable again
- the invocation itself: `claude /wayfinder <map> <ticket>` with permissions flags, cwd set to the project checkout
- what `wf` sees afterwards: nothing (tracker is truth) — confirm that's livable, i.e. the pane closing needs no cleanup from `wf`

Deliverable: a working throwaway script or snippet proving the path, and the chosen semantics written up.

blooop:
Prototype done — branch [`prototype/zellij-seam`](https://github.com/blooop/wayfinder/tree/prototype/zellij-seam), see [`prototypes/zellij-seam/prove.sh`](https://github.com/blooop/wayfinder/blob/prototype/zellij-seam/prototypes/zellij-seam/prove.sh) (re-runnable, headless, uses a throwaway detached session) and [`FINDINGS.md`](https://github.com/blooop/wayfinder/blob/prototype/zellij-seam/prototypes/zellij-seam/FINDINGS.md) for exact invocations and raw output. zellij 0.44.3. HITL: reporting measured behavior only — nothing decided or closed here.

**The seam works.** Measured, per the ticket's questions:

1. **Outside → named session:** `zellij --session <name> action new-pane --name 'wf#6' --cwd <dir> -- <cmd>` works in one shot (same for `new-tab`); stdout returns a stable handle (`terminal_2` / tab id) usable later with `focus-pane-id` / `go-to-tab-by-id`. Commands run even with no client attached.
2. **Inside:** a process in a pane inherits `ZELLIJ_SESSION_NAME`, and a bare `zellij action new-pane ...` lands in the same session, same tab as the launcher. No explicit targeting needed when `wf` runs inside zellij.
3. **Findability:** named `wf#<ticket>` panes/tabs show up in `list-panes` / `list-tabs` / `query-tab-names` / `dump-layout`. No focus-pane-by-name exists, and duplicate names are allowed — the spawn-time ID is the reliable handle.
[...]

PR #33, opened 2026-08-06
Stop leaking wf's raw tty into the zellij server (#30)

### LINKED ISSUE #30: The cross-session handoff still leaves a broken terminal
## Question

A cross-session launch still leaves the terminal broken after
[#22's fix](https://github.com/blooop/wayfinder/issues/22) — `04d08ae`, *"Give the
terminal back when a launch hands over"*. **Which layer is actually leaking, and
what does `wf` change?**

### The repro, as reported

`wf` running **inside** zellij session A; launch a ticket whose project is a
**different** session B — i.e. the `Handoff::Quit` path. Then exit the launched
window. Reported symptoms, verbatim:

- *"it gets dropped back into a broken zellij"* — the client is **still alive**;
  this is not the outer shell after zellij exited, and not a dead pane in A.
- *"when I press enter the terminal jumps 2 newlines"*.
- and, separately, keys not echoing at a prompt.

**Those are two different terminal states, and that is the crux.** Leaked raw
mode means *no* echo. Enter double-advancing means echo is **on** and newline
translation is applied twice — `onlcr`/`icrnl` disagreeing, or two layers each
emitting one. So "the terminal is messed up" is currently covering at least two
distinct faults, possibly on two distinct ttys, and nothing can be fixed until
they are separated.

### Established from code — do not re-derive

Read during charting (`main.rs`, `launch.rs` @ `edac56b`; note `main.rs` line
numbers move, another session is editing it):

- `Handoff::Quit` fires only for `(Mode::Hitl, Host::Inside(current))` where
  `current != session` (`launch::handoff`). `focus_steps` emits
  `go_to_tab_argv(B, tab)` then `switch_session_argv(A, B)` — B is pointed at the
  tab first, then **your client** is moved off A.
- `run()` returns `Ok(Ending::HandedOver)` immediately, and `main` calls
  `ratatui::restore()` before printing. So restore **does** run, promptly.
- ratatui 0.29 `init`/`restore` are exactly
  `enable_raw_mode + EnterAlternateScreen` / `LeaveAlternateScreen + disable_raw_mode`,
  and `suspend`/`resume` in `main.rs` are symmetric with them. `wf` enables **no**
  mouse capture and **no** keyboard-enhancement protocol.
- crossterm restores the termios saved at the *first* `enable_raw_mode`, i.e. the
  pre-`wf` state of the pane's tty.

**Therefore: `wf`'s handling of the one tty it owns — its own pane in A — looks
correct.** The leak is on a tty `wf` never owned, or it is zellij state (scroll
region, alternate screen, pane termios) that `wf`'s exit sequence desynchronises.
`switch-session` may be the *trigger* without `wf` being the *leaker*.

### First act is measurement, not discussion

HITL, at the user's keyboard. Capture `stty -a` — and ideally `tput`-visible state
— at each step, on **each** tty separately (the outer terminal, A's pane, B's pane):
before launching, immediately after the client lands in B, after exiting the
launched window, and once "broken". Diff them. The diff names the layer; only then
is there a decision to make.

### Two things to settle once the layer is known

1. **Does the fix even belong in `wf`?** If the leak is the zellij client's, `wf`'s
   options are to avoid `switch-session` as the handoff gesture, or to sequence its
   own teardown *before* moving the client rather than after.
2. **Entangled with [Picker or supervisor: does wf exit on select?](https://github.com/blooop/wayfinder/issues/26).**
   `Handoff::Quit` *is* "wf exits on select" for the cross-session case. If #26
   decides `wf` always exits on select, the Stay/Quit/Suspend trichotomy collapses
   and this fix changes shape. **Not blocked on it** — the measurement and the
   leak-localisation hold either way, and a terminal broken on every cross-session
   launch is daily-use-breaking, which is the destination. But check #26 before
   committing to a fix shape, and expect this answer to constrain it.

### Also found while charting (cheap, unrelated to the leak)

On the `Quit` path the parting line — *"run `wf` in that session to pick another
ticket"* — is `println!`ed **after** the client has already moved to B, so it lands
on an abandoned pty in A and **nobody can ever read it**. Unreachable by
construction. Fix or drop it as part of this ticket.

blooop:
Fixes #30 — the "broken terminal after a cross-session launch".

## What was wrong

`wf` is a ratatui TUI, so its tty is **raw** for as long as the picker is up. A zellij client reads the termios it gives new panes **from its own stdin**, and the server then stamps that state onto **every pane it ever creates**, for its whole lifetime.

`tokio::process::Command::output()` sets *only* stdout and stderr to piped (`tokio-1.53.1/src/process/mod.rs:1065`) — it never touches stdin, so stdin stays **inherited**. That is a silent divergence from `std::process::Command::output()`, which nulls it. Both of `launch.rs`'s spawn sites used `output()`, so every session `wf` created was born raw and the agent in it ran on a tty with:

- `-echo` — keys did not echo
- `-isig` — `Ctrl-c` inert
- `-opost` — `onlcr` not applied, so `\n` drops a line without a carriage return and output staircases

Which is exactly the reported symptom set, including the two states #30 called mutually exclusive. They were never in conflict: the pane was raw **from birth**, not left raw by a teardown.

Two hypotheses are ruled out by measurement, not argument. `wf`'s own `init`/`restore` symmetry was never at fault, and the **EXITED-pane post-mortem is not the cause** — zellij 0.44.3 renders *no* exit hint on a dead pane (no such string anywhere in the byte stream) and `Enter` there produces zero output, so `Enter` is not a live binding on a corpse and #5's no-`--close-on-exit` rule stands untouched.

## How it was localised

Identical `zellij attach --create-background`, varying only the creating process's fds:

| stdin | stdout/stderr | resulting pane termios |
| --- | --- | --- |
| raw pty | raw pty | **raw** |
| `/dev/null` (ctty *still* the raw pty) | `/dev/null` | cooked |
| **raw pty** | pipes | **raw** |

The last row is exactly `output()`'s shape; the middle row rules out the controlling terminal. It is the **fd**, not the ctty.

Then end to end on the real binary: `wf` driven under a pty through a genuine `Handoff::Quit`, and both sessions probed afterwards from an unrelated **cooked** shell — the session `wf` created reported `-opost -isig -icanon -echo`, the session it did not create reported none of them. Post-fix, both report `opost isig icanon echo`, and the agent's own tty comes up `icrnl opost isig icanon echo`.

## The change

`.stdin(Stdio::null())`, on a shared `zellij_command()` so both spawn sites get it and the reason is written down once. The leak is a property of the tty `wf` holds, not of which subcommand runs, so it belongs on **every** invocation rather than just session creation.

`Handoff::Suspend` still inherits the terminal deliberately — it calls `suspend(terminal)` first, restoring cooked mode, and the child there *is* the thing that needs the screen.

**Wider than #30 assumed:** any session `wf` creates was affected, so AFK agent tabs in new sessions had been running on raw `0×0` ptys too — not just the cross-session HITL path.

## Also, the unreachable parting line

#30 asked for this to be fixed or dropped. It was unreachable by construction: `println!`ed after the client had already moved to the other session, into an abandoned pty in the session just left. Dropped, and `Ending::HandedOver` / `LaunchReport::HandedOver` lose their `String` — the type no longer claims a message exists that nobody can read. The README already carried that guidance where it *can* be seen, and now says so explicitly.

## Verification

- New `tests/live_zellij_termios.rs` makes the test process's stdin a genuinely raw tty (`openpty` + `cfmakeraw`, asserting the premise), goes through the real `ensure_session` / `create_or_focus_tab`, and asks the pane itself what termios it got. **Verified to fail without the fix** (`#30: pane was born with -echo …`) **and pass with it**. Its own file, because it mutates fd 0.
- `cargo test`: 141 tests across 9 binaries green, including the existing live zellij, fetch, poll, discovery and streaming-startup suites.
- `cargo clippy --all-targets -- -D warnings` clean.
- `libc` added as a **dev**-dependency only.

## Summary by Sourcery

Detach zellij subprocesses from wf's raw TTY to prevent leaking raw terminal settings into created sessions, remove an unreachable handover notice path, and add an integration test ensuring sessions created by wf have cooked terminal settings.

Bug Fixes:
- Ensure zellij subprocess stdin is null so wf's raw TTY state is not propagated to new panes and sessions.

Enhancements:
- Refactor zellij process launching through a shared helper that centralizes environment scrubbing and stdin configuration.
- Remove the unreachable handover parting message and simplify related Ending/LaunchReport variants and behavior.

Build:
- Add libc as a dev-dependency for test-time TTY manipulation.

Documentation:
- Clarify README behavior when wf hands over from inside another zellij session and explicitly note that no parting line is printed.

Tests:
- Add an integration test that runs with a raw stdin TTY and verifies that sessions and panes created by wf have cooked terminal termios settings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -673,6 +673,7 @@ version = "0.2.2"
 dependencies = [
  "anyhow",
  "dirs",
+ "libc",
  "nucleo-matcher",
  "ratatui",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,3 +21,6 @@ tokio = { version = "1", features = [
     "sync",
     "time",
 ] }
+
+[dev-dependencies]
+libc = "0.2.189"
--- a/README.md
+++ b/README.md
@@ -128,7 +128,9 @@ How `wf` hands over depends on where it is itself running, decided from its own
 - **inside another session** — zellij's session switcher gesture
   (`switch-session`) moves you over, and `wf` exits as it goes: the tab it was
   drawing in belongs to the session you just left, so nothing could reach it to
-  quit it. Run `wf` again in the session you land in to pick the next ticket.
+  quit it. It prints no parting line either, for the same reason — you are
+  already elsewhere, so the only pane it could print into is one you have left.
+  Run `wf` again in the session you land in to pick the next ticket.
 
 No new navigation keybindings: getting back is zellij's standard detach or
 tab/session switching. The project's session is created detached if it does not
--- a/src/launch.rs
+++ b/src/launch.rs
@@ -36,6 +36,7 @@
 use std::collections::BTreeMap;
 use std::fmt;
 use std::path::{Path, PathBuf};
+use std::process::Stdio;
 
 use anyhow::{bail, Context, Result};
 use tokio::process::Command;
@@ -107,18 +108,19 @@ pub fn host_from_env(zellij: Option<&str>, session_name: Option<&str>) -> Host {
 /// Probed by *running* it rather than by searching `$PATH`, because the question
 /// that matters is whether this module's invocations will work — a `zellij` that
 /// is on PATH but cannot start is no more usable than an absent one. `--version`
-/// is the one invocation that touches no session, and the inherited zellij
-/// variables are scrubbed as everywhere else so a stale `ZELLIJ_SESSION_NAME`
-/// cannot make even this hang (#5 findings).
+/// is the one invocation that touches no session, and it goes through
+/// [`zellij_command`] like every other, so the inherited zellij variables are
+/// scrubbed — a stale `ZELLIJ_SESSION_NAME` cannot make even this hang (#5
+/// findings) — and `wf`'s raw tty stays out of its stdin (#30).
 pub async fn zellij_available() -> bool {
-    Command::new("zellij")
-        .arg("--version")
-        .env_remove("ZELLIJ")
-        .env_remove("ZELLIJ_SESSION_NAME")
-        .env_remove("ZELLIJ_PANE_ID")
-        .output()
-        .await
-        .is_ok_and(|output| output.status.success())
+    let argv = ["zellij".to_string(), "--version".to_string()];
+    match zellij_command(&argv, None) {
+        Ok(mut command) => command
+            .output()
+            .await
+            .is_ok_and(|output| output.status.success()),
+        Err(_) => false,
+    }
 }
 
 /// The answer to [`detect_host`], computed at most once per process.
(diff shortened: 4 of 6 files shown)


blooop:
Closing as **moot, not wrong**. The fix and its measurements are correct — see the [#30 resolution](https://github.com/blooop/wayfinder/issues/30#issuecomment-5203015255) — but the destination has since been redrawn: **`wf` drops zellij entirely and becomes just a ticket selector** that execs the agent and exits. There will be no zellij subprocess to leak a tty into, so `zellij_command` and everything around it is being deleted rather than fixed.

Kept out of main deliberately: merging a fix to code that is about to be removed would only make the removal diff harder to read. The branch `fix/30-raw-tty-leak-into-zellij` stays on the remote if the reasoning or the `openpty` regression-test technique is ever wanted.

What survives is the *knowledge*, recorded on #30: `tokio`'s `Command::output()` leaves stdin **inherited** where `std`'s nulls it, and a zellij client reads new panes' termios from its stdin. Worth remembering for any future subprocess spawned from a raw-mode TUI.

Issue #28, opened 2026-08-06
Cache the map search: get data on screen as fast as the frame


blooop:
## Question

Since [Draw the picker before the fetch: parallel, streaming startup](https://github.com/blooop/wayfinder/issues/27),
`wf`'s screen is up in **0.05s** — but its *tickets* still land at ~3.5s, and that
number is now almost entirely one call:

- `gh api search/issues` with **two** `repo:` qualifiers — **2.53s**. One qualifier is
  0.50s, so this scales badly in exactly the direction `wf` grows.
- Each `fetch_map` — 0.37–0.54s, and since #27 they run concurrently, so N maps cost
  one round trip.

The search is the one genuinely serial step: without it there are no issue numbers to
fetch, so nothing can overlap it. It also returns the least volatile thing `wf` knows —
which repo has a `wayfinder:map` and its issue number. A map issue is created once and
essentially never moves.

`~/.cache/wf/projects.json` is already a per-machine cache holding path, repo slug, and
session name for every registered checkout. It holds everything *except* the map number.

## What to decide

- **Where it lives.** Extending the existing `Checkout` record is the obvious move, but
  the map number is a property of the *repo*, not the checkout, and two checkouts of one
  repo would then hold it twice — a duplicated fact that can disagree with itself. Model
  it so it cannot.
- **How a cached number is modelled at all.** A checkout can be: not yet searched, known
  to have no map, or known to have map #N. Three states, and the sentinel temptation
  (`0`, or `Option<u64>` doing double duty) is the same one #27's `Startup` had to
  remove. Also: is the cache read on a fast path that then *skips* the search, or is the
  search still fired and the cache only a head start? The second is strictly safer and
  costs nothing on screen.
- **Invalidation.** A stale number is worse than no number: it fetches a wrong or closed
  issue and the map renders as something it isn't. What happens when a cached issue is
  no longer a `wayfinder:map` — the fetch fails, so is failure enough to trigger a
  re-search, or does the fetch have to verify the label it assumed?
- **First run stays honest.** A cold cache must still show `searching for maps…` and
  behave exactly as it does today. The cache is a fast path, not the mechanism.

## Why now

[Picker or supervisor: does wf exit on select?](https://github.com/blooop/wayfinder/issues/26)
turns on the cost of a *start*. #27 already corrected its 0.47–0.54s figure once; this
ticket is what would make a start genuinely cheap, which changes what #26 is choosing
[...]

Issue #27, opened 2026-08-05
Draw the picker before the fetch: parallel, streaming startup


blooop:
## Question

`wf` shows nothing until every network call has finished. `main` does all of this
*before* `ratatui::init()`:

1. `discover_checkout(cwd)` — two git subprocesses, <10ms
2. `find_maps(repos)` — one `gh api search/issues`, measured **0.50s**
3. `fetch_map` per mapped repo, in a **serial `for` loop** — `gh api graphql`,
   **0.37–0.54s each**
4. `agent_tab_count` — `zellij list-sessions` plus a `query-tab-names` per live session

The first frame therefore lands at roughly **0.5 + 0.5×N** seconds, where N is every
mapped project you have ever registered, and the terminal is blank for all of it. Step
4 also puts a `zellij` call — the wedge-prone boundary of
[A wedged zellij call freezes the whole TUI](https://github.com/blooop/wayfinder/issues/21) —
on the path *before* there is any UI to report a problem in.

Load immediately instead: draw the picker first, then let discovery and every map fetch
stream into the running UI, in parallel.

## What to do

- **Terminal first.** `ratatui::init()` and the first `draw` happen before any network
  or zellij IO. Startup progress becomes in-frame state, not the `eprintln!`s that only
  exist today because there is no screen yet.
- **Parallel, not serial.** The per-repo initial fetches are independent `gh`
  subprocesses keyed by slug; `0.5×N` becomes `0.5`. The same applies to `refetch_all`,
  which has the identical serial loop on the `ctrl-r` path.
- **Stream.** The machinery already exists: `run()` drains an mpsc channel per frame and
  calls `replace_map(merge_maps(&maps))`. Startup should feed *that* channel rather than
  completing before it is created.
- **Model the pre-data state.** An empty list currently means "no tickets"; under
  streaming it would also mean "not loaded yet" — one value, two meanings, which is the
  sentinel this map's `/constructive-modeling` constraint rules out. Loading needs to be
  its own state, visible on screen, and the `bail!` on "every map fetch failed" has to
  become displayable state now that there is a UI up to display it.
- **Decide when the cwd focus applies.** `Scope::Project` is chosen today from
  `maps.contains_key(slug)`, which is only knowable after the fetch; the map *search*
  answers the same question 0.5s earlier. Applying it late must not silently override a
  `ctrl-g` the user pressed while waiting.

## Why it matters beyond the wait

[...]

Issue #26, opened 2026-08-05
Picker or supervisor: does wf exit on select?


blooop:
## Question

Is `wf` a **long-running supervisor** or a **short-lived picker**?

Today it is a supervisor: a TUI that stays up, polls the tracker every 4s, suspends itself to attach
to an agent session, resumes when you detach, and reconciles the auto-start invariant on every poll.
Builds 1–6 all assume that shape.

The alternative: run `wf`, see the tree, select the next piece of work, it spawns the tab and **exits**.

This is upstream of [A wedged zellij call freezes the whole TUI](https://github.com/blooop/wayfinder/issues/21),
which is why that ticket is blocked on this one. The freeze is a render thread parked on a wedged
subprocess — a picker that exits has no render thread to park, so the whole bug class becomes
**unrepresentable** rather than fixed. #21's drafted resolution (launch onto a task, outcome as a sum
type, `esc` cancels, `kill_on_drop`) is a correct fix *for a supervisor* and dead code for a picker.

### What is known

- **#5's rejection of `exec` is narrower than it reads.** It says: "never `exec` (detach would end the
  window)". But inside zellij — the common case — a picker never attaches at all: `action new-tab` +
  `go-to-tab`, then exit, and its own pane closes while native zellij carries you over. Outside zellij,
  ending on detach just returns a shell prompt.
- **Restart is cheap, and was not known when #5 chose.** One `gh api graphql` map fetch measures
  **0.47–0.54s**; `gh api` REST 0.47–0.75s; every `zellij` call 10–50ms. #5 traded a long-lived TUI for
  roughly half a second.
- **Long life exists for auto-start, and that is 19% of the code**: `refresh.rs` (314 lines) +
  `autostart.rs` (620) of 4948. Build 6 shipped it as "`wf` now works while you are away" (#18, #19).
- **The split is already latent.** `autostart::reconcile` is a *pure* function gated by a no-IO
  `any_candidate`; Build 6 built it separable without needing it to be.
- **Live refresh is already marked droppable.** #11 called it "droppable comfort", and a 0.5s cold start
  makes polling-to-stay-fresh close to pointless for something you re-run on demand.

### The shape of the answer

The candidate is that these are **two programs with two lifetimes**, and fusing them into one TUI is the
architectural mistake:

- **`wf`** — the picker. Fetch once, draw, select, spawn, exit.
- **the reconciler** — long-lived and **headless**. No render thread, so a wedged `zellij` cannot freeze
[...]

PR #46, opened 2026-08-06
Build 7 — delete zellij: wf is just a ticket selector

### LINKED ISSUE #34: Build 7 — delete zellij: wf is just a ticket selector
## Question

Make `wf` what [#26](https://github.com/blooop/wayfinder/issues/26) decided it is: **a ticket
selector.** Draw the list, select a ticket, run the agent right here, exit. **No interaction with
zellij at all.**

This ticket is the deletion. It is mostly subtraction, and the diff *is* the change — nothing about
launching needs inventing, because [#31](https://github.com/blooop/wayfinder/pull/31) already built
and proved the mechanism on the `Host::NoZellij` path: `claude --dangerously-skip-permissions
"/wayfinder <map> <n>"` as a child in the checkout, tested end to end under a pty with a fake
`claude`. What changes is that it stops being one host of four and becomes **the only** behaviour,
and that `wf` **exits** instead of resuming the picker.

### Goes

- **`Host`** (all four variants), **`Handoff`** (`Stay`/`Quit`/`Suspend`), `detect_host`, the
  `OnceCell` probe, `host_from_env`, `zellij_available`. No host to branch on.
- **The tab machinery**: `new_tab_argv`, `go_to_tab_argv`, `query_tab_names_argv`,
  `switch_session_argv`, `attach_argv`, `create_session_argv`, `delete_session_argv`,
  `ensure_session`, `session_state`, `create_or_focus_tab`, `find_tab`, `tab_exists`,
  `count_agent_tabs`, `tabs_by_session`, `TabKey`/`TabLabel` ([#20](https://github.com/blooop/wayfinder/issues/20)),
  `Opened`, `TabOutcome`, `zellij_command` — and `run()`, the whole verified-by-re-reading boundary.
- **`autostart.rs`** (~620 lines) and **`refresh.rs`**'s polling (~314) — long life existed for
  auto-start, and [#26](https://github.com/blooop/wayfinder/issues/26) dropped it. `ctrl-a`, the
  `agents: N tabs` slot, the two-tier ETag poll
  ([Build 5](https://github.com/blooop/wayfinder/issues/17)), three-state poll health, and
  `TicketType`'s role in gating auto-start all go with them. `TicketType` itself **stays** — it is
  displayed.
- **`tests/live_zellij_launch.rs`**, and the `libc` dev-dependency if nothing else wants it.

### Stays

Everything about *selecting*: the stack (#2), the GraphQL data plane (#3), accretive discovery (#4),
the grouped list (#8), the picker screen and its keybindings (#9), streaming startup (#27), the
map-search cache (#28), the projects cache, the **multi-checkout picker** (a repo can have several
checkouts and the agent must run in exactly one), `wf --version` and the packaging.

### The two things to settle while doing it

1. **`exec` or spawn-and-exit?** #26 judged `exec` the right shape — `wf`'s process *becomes* the
   agent, so no parent lingers with nothing to do and the agent inherits the terminal, the exit code
   and the signals directly. `#5`'s "never `exec`" is spent: it existed so `wf` could survive a
   detach, and nothing survives now. **Restore the terminal before the `exec`** — that ordering is
   the one thing that can go wrong, and it is exactly the `ratatui::restore()`-before-handover that
   `04d08ae` and #30 were both circling. If `exec` proves awkward under tokio (it replaces the
   process mid-runtime), spawn-and-wait-then-exit is the fallback; say which was used and why.
2. **Does the picker fetch once, or keep refreshing while it is up?** #26 deleted *polling*, but the
   picker is on screen for a few seconds and #27's streaming load must still work. Simplest reading:
   the load streams in as it does today, and there is no repeat poll afterwards — `ctrl-r` stays as
   the manual refresh. Confirm rather than assume, because it decides how much of `refresh.rs`
   survives.

### Constraints

- **`/constructive-modeling` applies** (map Notes). This is a deletion, so the win is types that
  stop existing: four hosts and three handoffs collapse into no enum at all, and `Opened`'s
  `Tab`/`Direct` split disappears with the tab. Watch for what the collapse *exposes* — a launch is
  now just (checkout, ticket, map) and should not be able to name a session it cannot use.
- **Keep the test suite honest.** It is 141 tests today and a large fraction test the seam being
  deleted. Deleting a test that guards deleted code is correct; quietly losing coverage of the
  *launch* path is not — #31's pty-driven end-to-end (real TUI, shimmed `claude`) is the pattern to
  keep, since it is the only thing that proves `enter` actually runs the agent.
- **A version bump and a README rewrite** are part of this, not follow-ups: the *Without zellij*
  section becomes the only path, and the handover bullets describing the three-way `$ZELLIJ` split go.

Done when `grep -ri zellij src/ tests/` is empty, `enter` runs the agent and `wf` is gone, and the
tests that remain all test things that still exist.

blooop:
Closes #34. Makes `wf` what [#26](https://github.com/blooop/wayfinder/issues/26) decided it is: **a ticket selector.** Draw the list, pick a ticket, run the agent right here, exit. **Net −2830 lines.** `grep -ri zellij src/ tests/` is empty.

## The two things #34 asked to settle

**1. `exec`, not spawn-and-exit.** It works fine under tokio, so the fallback was not needed. `run` returns `Ending::Quit | Handover(Launch)` and `main` restores the terminal and *then* calls `Launch::exec`, which returns only on failure — there is no `Ok` for it to produce, since on success this process *is* the agent. Restore-before-exec is the one thing that can go wrong, so it is two adjacent statements in `main` rather than anything buried in the loop.

**2. Load once, `ctrl-r` to ask again** (confirmed with the human, not assumed). A selector on screen for seconds and restarting warm in ~0.6 s has nothing to keep fresh. `Pollers` → `Loaders`: one shot per repo, still reconciled against the search's answer so a moved map number reaches the task doing the fetching. #27's streaming load and #28's cache head start are untouched.

## Goes

`Host` (all four variants), `Handoff`, the tab machinery (`TabKey`/`TabLabel`, `create_or_focus_tab`, `ensure_session`, `session_state`, `Opened`, `TabOutcome`, `run()` and its verified-by-re-reading boundary), `autostart.rs`, the two-tier ETag poll, `ctrl-a`, the `agents: N tabs` slot, `suspend`/`resume`.

Two things beyond #34's list, both because the collapse exposed them as unread:

- **`Checkout.session`** and `derive_sessions`/`path_slug`. The nickname existed only to name a session; nothing else ever read it. `Checkout` is `{path, repo}`, and the which-checkout modal distinguishes trees by **path**, which is what actually distinguishes them. Older cache files load straight past the extra key.
- **`Freshness`/`RefreshState`**, and `RefreshEvent`'s now-unreachable `Unchanged` (→ a two-state `MapFetch`). An age-since-last-poll counter with no poll is noise. A failed load now names the repo on the count line, which the global stale flag never did.

## Modelling

A launch is exactly **(checkout, ticket, map)**. Its fields are private with `plan` the only constructor, so #34's "should not be able to name a session it cannot use" is enforced rather than documented. Four hosts and three handoffs collapse into no enum at all; `Opened`'s `Tab`/`Direct` split goes with the tab.

`TicketType` **stays** — parsed once at the `gh` boundary, still asserted in `fetch` and `live_fetch`. One correction to #34 there: it is *not* displayed anywhere, only carried. Displaying it would mean adding a column to the minimal row [#9](https://github.com/blooop/wayfinder/issues/9) locked, so it stays as-is rather than reopening that decision inside a deletion. `auto_startable()` is deleted — auto-start is permanently out of scope, so its only reader is never coming back.

## Tests

141 → 92, all of the loss guarding deleted code.

[#31](https://github.com/blooop/wayfinder/pull/31)'s pty-driven end-to-end was a **manual** check, not a committed test. It is committed now as `tests/live_launch_exec.rs`, because after this deletion it is the only thing that proves `enter` actually runs the agent. Real binary, real pty (sized explicitly — 0×0 lays out nothing and the test would wait forever for a row), `claude` shimmed to a script reporting its argv, cwd and termios, scratch `XDG_CACHE_HOME` so the user's real projects cache is untouched. It pins three claims:

1. the agent gets `--dangerously-skip-permissions "/wayfinder 1 <n>"` **in the checkout**;
2. the spawned pid *is* the agent — an `exec` replaced the image, so there is no `wf` parent left. Spawn-and-wait would not satisfy this;
3. the tty it inherits is **cooked**. Verified to have teeth: inverting the restore/exec order fails it with exactly #30's `-echo -isig -icanon -opost`.

`cargo clippy --all-targets -D warnings` is clean. (`cargo fmt --check` has pre-existing diffs on main and is not a CI gate; left alone rather than churning unrelated lines.)

## Also

0.2.2 → 0.3.0. README rewritten — the *Without zellij* section is now the only path, the `$ZELLIJ` handover table and the `ctrl-a` row are gone, and a short **Working while you are away** section says the honest thing: that is another terminal session, not a feature. Recipe description and run-deps note updated.

## Summary by Sourcery

Simplify wf into a single-process ticket selector that loads maps once and execs the agent in-place, removing all zellij/tab/autostart machinery and background polling.

New Features:
- Run the selected ticket’s agent by execing `claude` in the chosen checkout, replacing the wf process.
- Load each project’s map once per run with a streaming startup that uses cached map numbers and a retrying label search.
- Add an end-to-end pty-based test that verifies `enter` execs the agent with the correct argv, cwd and cooked tty.

Enhancements:
- Refine launch modelling so a `Launch` is exactly (checkout, ticket, map) with private fields constructed only via `plan`.
- Simplify refresh/load infrastructure from long-lived pollers to one-shot per-repo loaders and a leaner startup state tracker.
- Clarify ticket type semantics while keeping `TicketType` as a purely interpretive axis without auto-start logic.
- Streamline the projects cache format to `{path, repo}` and keep checkouts sorted by path for a stable checkout picker.
- Tighten UI around a single launch key, removing AFK-specific chrome and making the which-checkout modal purely path-based.
- Clarify library and main binary docs to describe wf as a short-lived ticket selector that hands the terminal over.

Build:
- Bump crate and recipe version from 0.2.2 to 0.3.0 and adjust the recipe description and test expectations to match the new behaviour.

Documentation:
- Rewrite README and package metadata to describe wf as a ticket selector that execs the agent, remove zellij/auto-start docs, and add a brief note on working while away.

Tests:
- Remove obsolete tests tied to zellij sessions, auto-start and freshness, and add/update tests for the new launch, load and cache behaviours, including live startup and exec tests.


blooop:
## Adversarial review round

Two reviewers — one on regressions from the deletion, one on exec/terminal/process semantics. **Seven confirmed defects, five of them mine.** Fixed in `7e6689e`.

### The bad one

**The cursor was never made visible again before the exec.** Nothing in the picker positions a cursor, so every `Terminal::draw` writes `ESC[?25l`, and the only thing that writes `?25h` back is `Terminal`'s `Drop` — `ratatui::restore()` is *only* raw-mode-off plus leave-alternate-screen. On the quit path `Drop` runs at the end of `main`; on the handover path `exec` replaces the image first and it never runs. So the cursor was restored in exactly the two cases where it does not matter and left hidden in the one where `wf` hands the terminal to somebody else. DECTCEM outlives the alternate screen (verified against tmux), so it genuinely reaches the agent — a healthy Ink `claude` masks it by emitting `?25h` on mount, a `claude` that dies before mounting does not. **This is the line the deleted `suspend()` had**, and the PR dropped it.

### The rest

- **A failed fetch read as "no projects".** `Freshness::Stale` was drawn every frame; I replaced it with `app.notice`, which `handle_key` clears on the next keypress. Fetch fails → search succeeds → `Startup` reports loaded → heading becomes *no projects — run wf inside a checkout to register it*, i.e. the exact lie `Startup` exists to prevent, sending the user to fix the one thing that is not broken. Now `App::failed` is a **set of slugs, drawn every frame** — names the repo in the heading when nothing loaded, and a persistent count-line segment for the partial case (four maps on screen, one missing), which is the one that hides best. Strictly stronger than the flag it replaced, which could not say *which* repo.
- **`ctrl-r` was clobbered by the load it raced.** `refetch_all` wrote `maps` directly while a one-shot load for the same repo was in flight; the older snapshot landed second and won, while the line said `refreshed`. The 4 s poll used to heal this, so deleting the poll made it terminal. `ctrl-r` now goes through `Loaders::restart` — one channel, send order, newest write wins by construction. `refetch_all` is deleted, taking an `.expect("slug is owner/name")` panic with it.
- **In-flight `gh` children survived the exec.** `▶` needs only one map and the search runs ~2.5 s alongside, so pressing enter early leaves a live child at `execve`. Verified inherited by the agent, in its foreground process group, holding its tty on fd 0, zombie for the agent's lifetime. Fixed with `stdin(Stdio::null())` (tokio's `Command::output()` pipes only stdout/stderr and leaves stdin **inherited** — the same divergence from `std` that caused #30) plus `kill_on_drop` and an awaited `Loaders::shutdown` before handover. `abort()` alone would not do it: the child dies when the task's `Child` drops, and there is no next poll before an exec.
- **`Launch::exec` resolved the agent on PATH *after* chdir.** An empty `$PATH` entry — leading/trailing/doubled `:`, an everyday `.bashrc` accident — resolved `claude` out of **the checkout**. Cloning a repo and running `wf` in it was enough to run its `./claude` with `--dangerously-skip-permissions`. Resolution now happens before the chdir and drops empty entries. Pre-existing on main via `current_dir` + spawn; fixed here because this is where it now lives. Also splits the two `ENOENT`s that shared one message.

### The test was worse than the code

All three self-inflicted:

- **claim 2 was vacuous.** The shim exits 0, and so would a spawn-and-wait `wf` — `status.success()` never distinguished the two designs. Now asserts the shim's `$$` equals the spawned pid.
- **claim 3 was blind to the cursor bug it existed to catch.** DECTCEM and the alternate screen are DEC private modes; `stty` cannot see either. Now also asserts `?1049l` and `?25h` in the pre-handover byte stream. Verified to have teeth: removing the `show_cursor` line fails it.
- **the reader could mangle its own needle.** It decoded each 4096-byte chunk separately, so a read boundary inside `▶` (`E2 96 B6`) became U+FFFD. ratatui writes diffs, so no redraw — the test would have hung its full 60 s. Reads bytes now.

### Left unfixed, deliberately

Both PLAUSIBLE, both recorded on #34: a signal landing in the tens of microseconds between restore and `execve` loses the launch (no corruption, and pre-existing guard code); and keystrokes already parsed into crossterm's queue when `enter` is read are dropped rather than reaching the agent (one `read()` wide).

Also cleared by the reviewers, with reasoning rather than assertion: legacy `{path,repo,session}` cache entries still deserialize (no `deny_unknown_fields`), `Startup::is_loaded` is reachable on every path, and no assertion about *surviving* behaviour was lost with the deleted tests.

97 tests, clippy `-D warnings` clean.

Issue #50, opened 2026-08-06
Build: every open map renders as its own cluster


blooop:
## Question

Land the multi-map model: every open map renders as its own cluster, including several on one repo.

Decided while charting, waiting only on the prototype's chosen rendering:

- Map identity becomes **(repo, map issue number)** — `find_maps` stops collapsing to lowest-number-per-slug (the rule that currently hides [Map #35](https://github.com/blooop/wayfinder/issues/35)), the projects cache and `Pollers` key on the pair, and `merge_maps` gives way to whatever cluster composition the prototype picked.
- The **full edge set becomes first-class** (`/constructive-modeling`): keep closed-blocker edges rather than discarding them in `parse_map`, derive unblocks-counts locally, and let `Status::Blocked { needs }` and the new edges coexist without restating each other.

Blocked by the prototype ticket: the cluster header and in-cluster layout are its answers.

blooop:
**breadcrumb:** claimed; model settled — `MapId { repo, number }` becomes the identity everywhere a map is named (find_maps returns every open map, cache field renamed `maps` → `open_maps` so old caches keep their checkouts and just lose the head start, Loaders/Startup/failed key on MapId); `merge_maps` deleted in favour of `clusters: BTreeMap<MapId, Map>` rendered as ▌-headed clusters with per-status counts; `Ticket.blocked_by` carries the full edge set (closed blockers kept) with `unblocks` derived by local inversion, `Status::Blocked{needs}` still the open subset parsed at the gh boundary. Launch's "repo has no map" notice becomes unrepresentable — a row always sits in a cluster that knows its map.

blooop:
**breadcrumb:** the pty end-to-end test caught a real regression the unit tests could not: with map #1's 29 done tickets rendered as the first cluster, the cursor row sat below the fold and the body never scrolled — the old global group-major order had only ever kept the cursor on screen by accident. Added scroll-to-cursor to the body draw (view-agnostic; #51's views need it too) and relaxed the e2e prompt assertion to `/wayfinder <map> <ticket>` shape, since which cluster the cursor lands in is the tracker's business now.

blooop:
## Resolution

Landed as [PR #53 — Build 8: every open map renders as its own cluster](https://github.com/blooop/wayfinder/pull/53) (CI green; 91 unit tests + all live integration tests, including the pty end-to-end launch, pass against the real tracker with this repo's three open maps).

What the build settled, beyond what the ticket already specified:

[...]

Issue #47, opened 2026-08-06
Map: the selection view — every open map, its structure, one screen


blooop:
## Destination

wf's picker renders **every open map** — across projects, and several per repo — as visible clusters, with the **blocking DAG legible inside each cluster** and **PR status on tickets**, so that choosing the next ticket is informed by structure (what unblocks what), not status alone. Shipped on prefix.dev, installed via pixi-global, and in daily use. The focused single-project view is the design center; the all-projects view (wf opened outside any checkout) must not degrade as project count grows.

## Notes

- **Execution belongs in this map** (same override as [Map #1](https://github.com/blooop/wayfinder/issues/1)): the route runs to a released `wf`, not a spec.
- **Rendering + navigation only**: enter-launches-the-agent-and-exits, discovery, and claiming are untouched by this effort.
- **`/constructive-modeling` on all work here** (constraint carried from Map #1): sum types over tag-plus-parallel-fields, no sentinels, exhaustive match, parse-don't-validate at the `gh` boundary.
- Skills for sessions here: `/grill-me`, `/constructive-modeling`, `/prototype`, `/research`.

Decisions locked while charting (2026-08-06):

- **Blocking edges are the relationship** this view is about — the DAG of blocked-by / unblocks. "A ticket with many children" means *many tickets depend on it* (high leverage). Sub-issue nesting below tickets is not adopted; no cross-map edges (maps stay per-repo, per the wayfinder model).
- **Maps render as visible clusters** — each open map is a unit with its tickets and structure inside it; "relationship between maps" means side-by-side shape and progress, not edges between maps.
- **Every open map renders, including several on one repo** — the lowest-number-per-slug rule in `find_maps` goes. (It currently hides [Map #35](https://github.com/blooop/wayfinder/issues/35) entirely.)
- **Structure vs status grouping is decided by prototype, not argument** — [Starmap legibility #8](https://github.com/blooop/wayfinder/issues/8) rejected DAG/tree for the grouped flat list, and this effort deliberately revisits that with a new purpose (selection support), by reacting to real renderings of live maps.
- **PRs are in scope, as status on their ticket** — evidence of a ticket's progress shown inline, not first-class rows. (Absorbs Map #1's last patch of fog, "PRs in the picture".)
[...]

Issue #35, opened 2026-08-06
Map: devpod-per-branch dev environment for wayfinder


blooop:
## Destination

`blooop/wayfinder` has a working, committed devpod + devcontainer setup where **one branch = one devpod workspace**, so several branches — and the agents working them — run at once without colliding, and the isolation-vs-startup tradeoff has a decided answer: always container-per-branch, with "fast" delivered by a warm cargo cache rather than a warm container.

Reached when: `.devcontainer/` is committed, `devpod up` on two branches at once gives two usable containers, `cargo test` passes inside one, the image is prebuilt in CI, and there is a documented host-side command to go from "work on branch X" to "agent running isolated on branch X".

## Notes

**Domain:** Rust dev tooling / container dev environments. The repo is `wf` — a pure cargo project (no pixi, no `rust-toolchain.toml`, no Dockerfile today), 91 locked deps, `target/` currently 2.1 GB.

**This map carries execution.** A devcontainer design can only be judged by booting it, so `task` tickets here genuinely build and commit files rather than only deciding. Decisions still land as resolution comments.

**Settled in charting (constraints, not open questions):**
- ~~Isolation is for **blast-radius containment** (every `wf` agent runs `--dangerously-skip-permissions`)~~, **concurrent working trees**, and **resource/port conflicts**. Toolchain parity with CI is explicitly *not* a driver.
  **Blast-radius containment struck 2026-08-07** by [#41](https://github.com/blooop/wayfinder/issues/41), following [#73](https://github.com/blooop/wayfinder/issues/73) on [map #67](https://github.com/blooop/wayfinder/issues/67): the repos this serves want `network=host`, X11 and device passthrough, so the container is not a confidentiality boundary and was never going to be one. The other two drivers stand, and they are what the container actually delivers. The human's call is explicit — *"an unsafe but ergonomic version first; once I understand the workflows I will harden it"* — so hardening is a later effort, not a gap here.
- **devpod workspace per branch** is the topology. This sidesteps the git-worktree trap (a worktree's `.git` is a file pointing into the parent repo, so mounting only the worktree breaks git) and gives per-branch start/stop/delete for free. `devpod` is already on PATH.
- Image lineage: **own `.devcontainer/Dockerfile` FROM `mcr.microsoft.com/devcontainers/rust`**, plus `gh` + `claude` + `zellij`. Not python_template's prebuilt image (carries pixi/python this repo doesn't want, and couples release cadences), not pixi-managed rust.
[...]

Issue #51, opened 2026-08-06
Build: leverage view default, structure forest on toggle, grouped view retires


blooop:
## Question

Land the chosen view model ([#48 resolution](https://github.com/blooop/wayfinder/issues/48)): the leverage view as the one default screen, the structure forest on a toggle, the grouped view retired.

- Default: per cluster, takeable tickets (frontier + claimed) sorted most-direct-dependents-first, each with its unblocks-subtree of open tickets; `● N done (hidden)` per cluster; `⊘ N blocked deeper down` where applicable; idle maps dropped from the body with a `· N idle maps hidden` footer count.
- Toggle (key TBD in build, e.g. `tab`): full structure forest — whole DAG, done dimmed in place, parent = lowest-numbered blocker, `⤷ also needs #n` for extra edges.
- Filtering: a live query flattens to a nucleo-score-ordered flat list (rows name their repo); empty query restores the tree. The 2a groups-survive-typing behavior retires with the groups.
- Row anatomy: `<glyph> #n <title> [type]` — ticket types displayed for the first time.
- `/constructive-modeling`: the view state (tree vs flattened-by-query vs toggled-forest) should be a sum type, not booleans.

Blocked by the multi-map cluster/model build. Prototype reference: branch `prototype/selection-views`.

blooop:
**breadcrumb:** claimed; build starting on branch `build9-leverage-view`. Design decisions: toggle key = `tab`; view state is a derived sum type — `Screen::Structured(Lens::{Leverage,Forest}) | Flattened{query}` with only the `Lens` stored, so flattened-without-a-query is unrepresentable; leverage roots sort by *open* direct dependents (a root whose dependents are all done has no leverage left); `⊘ N blocked deeper down` counts blocked tickets *not rendered in any subtree* (truthful, vs the prototype's direct-blocker approximation); idle-maps count rides the bottom count line.

blooop:
**breadcrumb:** build landed as [PR #55](https://github.com/blooop/wayfinder/pull/55) — new `src/view.rs` plans the body as one item list shared by cursor and draw; grouped view and 2a filtering deleted; 113 tests green including the live pty e2e. Awaiting CI + merge.

blooop:
## Resolution

Landed as [PR #55](https://github.com/blooop/wayfinder/pull/55), squash-merged to main. The #48 view model is `wf`'s screen now:

[...]

PR #53, opened 2026-08-06
Build 8 — every open map renders as its own cluster (#50)

blooop:
Lands the multi-map model for [#50](https://github.com/blooop/wayfinder/issues/50), the first build slice of [Map #47](https://github.com/blooop/wayfinder/issues/47).

## What changed

- **Map identity is now `MapId { repo, number }`.** `find_maps` returns every open `wayfinder:map` — the lowest-number-per-slug rule that hid [Map #35](https://github.com/blooop/wayfinder/issues/35) is gone. The projects cache, `Loaders`, `Startup`, and the failure set all key on the pair, so two maps on one repo are two loads, two arrivals, two possible failures.
- **`merge_maps` is deleted.** The screen holds one cluster per open map: a `▌ repo · map title  ○n ◐n ⊘n ●n` header with the map's tickets grouped beneath it (frontier / claimed / blocked / done). The repo column is gone — the header carries it. The body scrolls to keep the cursor visible, since several clusters can be taller than the terminal (the live 3-map repo hits this immediately).
- **The full blocking edge set is first-class** (`/constructive-modeling`): `parse_map` keeps closed-blocker edges on `Ticket::blocked_by` (structure) while `Status::Blocked { needs }` remains the open subset parsed at the `gh` boundary (status); `Map::unblocks` derives reverse edges by local inversion. Zero extra API points — the query always carried them.
- **Launch reads the map from the row's cluster**, so the "repo has no map" notice became unrepresentable and was deleted.
- **Cache migration:** the field is renamed `maps` → `open_maps`, so an old cache file loads with its checkouts intact and just loses the head start once.

## Testing

- 91 unit tests (multi-map fixtures throughout: two clusters on one repo, same ticket under two maps, moved map numbers, cache migration, scroll-to-cursor).
- All live integration tests updated and passing against the real tracker, including the pty end-to-end launch test — which is what caught the missing scroll.

## Summary by Sourcery

Render every open wayfinder map as its own on-screen cluster keyed by MapId, update startup/loading and discovery to work per-map instead of per-repo, and keep the full blocking edge set in the model while preserving cursor and launch behavior across multi-map refreshes.

New Features:
- Support multiple open maps per repo with each rendered as an individual cluster including per-status ticket counts.
- Allow focusing a project while still showing all of its open maps rather than a single map per repo.

Bug Fixes:
- Ensure the body view scrolls to keep the cursor-visible row on screen even when clusters exceed terminal height.
- Prevent cursor jumps when the same ticket appears on multiple maps by anchoring selection to (map, ticket) identity across refreshes.

Enhancements:
- Refactor map identity, cache, loaders, and failure tracking to use MapId-based sets instead of per-repo map numbers.
- Make blocked-by relationships first-class by preserving closed-blocker edges and deriving reverse unblocks edges locally from the full DAG.
- Simplify launch planning to read the map number from the cursor row’s cluster instead of a separate per-repo map table, removing the "repo has no map" path.
- Update UI layout and copy to remove the repo column, introduce map cluster headers, and adjust headings and failure notes to reflect per-map state.

Documentation:
- Document the per-machine cache as storing all open maps and describe the new per-map cluster rendering behavior in the README.

Tests:
- Extend unit, UI, and live integration tests to cover multi-map scenarios, cache migration, cursor anchoring, per-map failures, body scrolling, and the new MapId-based discovery and loading logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,8 @@ to open on all of them.
 
 The registry is a per-machine cache at `~/.cache/wf/projects.json`
 (`$XDG_CACHE_HOME` respected) holding `{path, repo}` per checkout, plus every
-open map the last search found. Deleting it is safe — projects re-accrete as
-you open them, and the maps are re-found on the next start.
+open map that the last search found. Deleting it is safe — projects re-accrete
+as you open them, and the maps are re-found on the next start.
 
 Only repos with an open `wayfinder:map`-labelled issue show tickets; other
 checkouts stay cached but hidden.
--- a/README.md
+++ b/README.md
@@ -37,12 +37,17 @@ known project, `ctrl-f` re-focuses the row's project). Run outside a checkout
 to open on all of them.
 
 The registry is a per-machine cache at `~/.cache/wf/projects.json`
-(`$XDG_CACHE_HOME` respected) holding `{path, repo}` per checkout, plus the map
-issue number the last search found for each repo. Deleting it is safe — projects
-re-accrete as you open them, and the map numbers are re-found on the next start.
+(`$XDG_CACHE_HOME` respected) holding `{path, repo}` per checkout, plus every
+open map the last search found. Deleting it is safe — projects re-accrete as
+you open them, and the maps are re-found on the next start.
 
 Only repos with an open `wayfinder:map`-labelled issue show tickets; other
 checkouts stay cached but hidden.
+
+**Every open map renders as its own cluster** — a `▌ repo · map title` header
+carrying the per-status counts (`○` frontier `◐` claimed `⊘` blocked `●` done),
+with the map's tickets grouped beneath it. A repo can keep several maps open at
+once and each gets its own cluster; focusing a project shows all of them.
 
 ## Startup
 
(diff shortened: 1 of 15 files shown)


blooop:
Thanks @sourcery-ai — took two of the three.

**Named `Row` and `RowKey` structs** (done). This was the right call and lands squarely on this map's standing `/constructive-modeling` constraint: the two were same-shaped tuples whose *difference* is the whole point — `Row { map, index }` is positional and only valid against the clusters it was read from, `RowKey { map, ticket }` is the durable identity that survives a refetch. Named fields say that at every call site where `.0`/`.1` could not, and `App::row_key` is now the single place the conversion happens (three hand-rolled constructions collapsed into it). `preserve_cursor` stays generic over `K`, with its two-maps-one-ticket test rerun against the real `RowKey`.

**README grammar** (done).

**Caching per-cluster derived data** — declining, and noting why so it isn't silently dropped. The condition you attached ("if performance becomes an issue with larger maps") isn't met and isn't close: the largest live map here is 30 tickets across 3 clusters, redrawn on a 250 ms tick, and `wf` is on screen for seconds. The `MapId` clones and the re-scans in `visible`/`unblocks` are microseconds against a `gh` round trip measured in hundreds of milliseconds. A derived-data cache would add an invalidation edge — cache versus clusters, disagreeing after a refetch — which is exactly the kind of state this codebase spends its modelling budget deleting. If [#51](https://github.com/blooop/wayfinder/issues/51)'s leverage view makes the DAG walks hot, that's the moment to profile and revisit.

Issue #62, opened 2026-08-06
Prototype: the typed two-step launch — filter, select, staged prompt


blooop:
## Question

What does the typed two-step launch feel like — filter, select, staged prompt — and which existing chords retire to typed words?

Prototype it: enter on a node opens a second step where empty-enter launches interactive (default), a word selects a mode (defer), and free text becomes a steering prompt for the agent. React to a real rendering: what the prompt line shows (node name, resolved skill, mode), how esc backs out, how hjkl/arrows coexist with typing (j/k type mid-query today — does the two-step change that?), and whether ctrl-f / ctrl-g / ctrl-r / tab survive as chords or become typed words. Minimise shortcuts; keep the workflow simple. Prototype on a throwaway branch, linked as an asset.

blooop:
## Resolution

**agent-decided:** resolved by judgment on the human's explicit instruction ("use best judgement from here and implement") — this HITL prototype ticket is converted rather than reacted-to; the build PR is the artifact to react to, and this design can be overturned there cheaply.

**The two-step launch:**

1. **Step one is unchanged**: typing always filters, arrows/hjkl move the cursor (h/l reveal/close per the depth-aware cursor, taken as given from #57/#58).
2. **Enter on a launchable node opens the launch line** — a one-line input replacing the count line, showing the resolved route: `→ /tdd · #65 Author the /tdd skill`. Then:
   - **empty + enter** — interactive launch (today's behavior, the default)
   - **`defer`** — deferred subtree mode
   - **`defer <text>`** — deferred + steering prompt
   - **any other text** — steering prompt on an interactive launch
   - **esc** — back to the list, query and cursor intact
3. **A `done` or blocked node doesn't open the line** (not launchable; the enter is a no-op with a count-line notice).

**Stage rendering** (handed here by #61): the row keeps #51's anatomy; the status glyph column *becomes* the stage glyph: `○` ready · `◐` building/in progress · `◍` in review · `!` needs attention · `●` done, with `⊘` blocked overriding (a blocked node's stage is unactionable). Rollups on collapsed rows reuse the glyphs as counts (`◐2 !1`).

[...]

Issue #10, opened 2026-08-04
Teach the /wayfinder skill the breadcrumb protocol


blooop:
## Question

Teach the /wayfinder skill the breadcrumb protocol — the writer side of [Re-entry breadcrumbs](https://github.com/blooop/wayfinder/issues/6).

Task (execution belongs in this map): update the skill files in the dotfiles repo (chezmoi source; SKILL.md and GITHUB_TRACKER.md) so every wayfinder session:

- posts `**breadcrumb:**` comments at decision-grade moments (sub-decision settled, direction change, question parked)
- posts a `### handoff` comment (where we are / open thread / first move on resume) on deliberate exit
- runs the re-entry ritual when resuming a claimed ticket: body → last handoff → later breadcrumbs → Decisions-so-far; first write is a resumption breadcrumb; first words confirm the open thread
- closes tickets found effectively complete, with a comment saying why
- skips claimed tickets only when choosing work autonomously (claims never gate the human)

Done when the skill text implements the protocol and the dotfiles cheatsheet/README reflects any user-facing change.

blooop:
## Resolution

The /wayfinder skill now implements the writer side of the [re-entry breadcrumb protocol](https://github.com/blooop/wayfinder/issues/6), in dotfiles commit `e0273f898e2d32c0800dedc241364043b68865c5`:

- **SKILL.md** — new **Breadcrumbs, handoffs, re-entry** section: `**breadcrumb:**` comments at decision-grade moments (sub-decision settled, direction change, question parked — not narration; never edited or deleted); a `### handoff` comment on deliberate exit (*where we are / the open thread / first move on resume*, crash = no handoff, trail is the fallback); the re-entry ritual for resuming a claimed ticket (body → last handoff → later breadcrumbs → Decisions-so-far; first write a resumption breadcrumb, first words confirm the open thread; idempotent). The **Work through the map** flow references all three, adds handoff-before-deliberate-exit, and closes tickets found effectively complete with a comment saying why. The claim semantics in **Tickets** and the ticket-choice step now state that claims gate only *autonomous* choice — a human naming any open ticket is always valid.
- **GITHUB_TRACKER.md** — new **Journal on a ticket** section: `gh issue comment` for breadcrumbs and the handoff, `gh issue view N --comments` to read the trail (last `### handoff`, then what follows). The local-markdown fallback gains a matching `## Journal` convention.
[...]

PR #75, opened 2026-08-06
Build 12 — stage-aware tree + typed two-step launch (#74)

blooop:
## What landed

wf's tree is stage-aware and its launch is a typed two-step, per the resolutions of #61 (stage lattice + routing) and #62 (two-step launch + glyphs). Built test-first at the ticket's four pre-agreed seams:

- **Stage derivation** (`model::stage`) — a pure function (linked-PR facts, ticket state) → `Stage`, a five-value lattice (`Ready → Building → InReview → NeedsAttention → Done`) whose derived `Ord` *is* the max-over-open-PRs order. Per-open-PR fixed table; no open PRs: merged → done, else ticket state; PR dominates when present. Table-driven tests cover every row of the #61 table.
- **Routing** (`launch::route`) — `(TicketType, Stage) → Option<Route>`, exhaustive on both axes with no wildcard, so a new stage or type without a route fails to compile. Done is the total-function refusal; blocked is refused on status before routing is consulted. The vocabulary grows exactly one type: `wayfinder:build` (ranked below the decision types in the multi-label tie-break — it is the most autonomous type).
- **Launch-line state machine** (`app`) — enter on a launchable node opens the launch line where the count line was, showing the resolved route (`→ /tdd · #65 …`); text accumulates on the line, never in the query; esc restores the list with query and cursor intact; enter launches with the typed mode (`LaunchMode::parse`: empty/`defer`/`defer <text>`/steering). Done and blocked nodes are a no-op with a count-line notice. The mode survives the which-checkout picker.
- **View planning** (`view`/`ui`) — the glyph column is the stage glyph (`○ ◐ ◍ ! ●`, `⊘` blocked overriding) via one `RowGlyph` sum type; shut group rows carry a stage rollup as glyph+count pairs (`Fold::Shut { rollup }` — a rollup on an open group is unrepresentable), counted once per held node.

Exec'd commands: build @ ready/building/needs-attention → `claude "/tdd <n>"`; build @ in-review → `claude "/review <n>"`; decision types → `claude "/wayfinder <map> <n>"`; the launch-line mode rides any route as ` defer` / ` defer: <text>` / ` steer: <text>`.

One recorded judgment (flagged on #74): decision types route to /wayfinder at every non-done stage — PR-dominant derivation can put a prototype at in-review, and the #61 table's ready/in-progress row is read as "any unfinished stage".

The live e2e (`tests/live_launch_exec.rs`) now drives the two-step (enter → launch line → enter) and accepts any of the three route shapes.

[...]

Issue #14, opened 2026-08-04
Build 2 — fuzzy filter + keybindings


blooop:
## Question

Wire in fuzzy filtering and the keybinding skeleton: nucleo matching with groups-surviving-typing (2a — non-matching rows drop, groups and order stay), input line at the bottom, keys `enter` (no-op until the launch seam) `ctrl-f` `ctrl-g` `ctrl-r` `esc`. `tab` peek is deferred from v1. Done when typing narrows the live list per the main-screen resolution.

blooop:
**breadcrumb:** claiming Build 2. Plan: `app.rs` owns main-screen state (query, cursor over visible rows, All/Project scope, notice line) with `handle_key -> Outcome::{Continue,Quit,Refresh}` — the Refresh outcome is the seam #17's poller can reuse for ctrl-r force-refresh without us touching fetch internals. `filter.rs` scores nucleo (`nucleo-matcher 0.3`, mined from prototype/mainscreen) against `repo #num title`; render is 2a per the #9 resolution — non-matching rows drop, all four group headers persist as `matched/total`. Enter is a visible no-op notice until Build 4. Keys: esc clears query then quits; q quits only on empty query; ctrl-f/g wire scope state (near-no-op with one repo); arrows + ctrl-j/k move over ticket rows only.

blooop:
**breadcrumb:** integration moment with #17, which landed mid-build (a168638, background ETag poll). Resolved the rebase by composing the layers rather than picking one: the run loop drains the poller channel into `App::replace_map`, `draw` takes both the App and the freshness indicator on the count line, and ctrl-r force-refetches in place (also marking data verified-fresh). Follow-up e916900 makes `App::replace_map` delegate cursor preservation to `refresh::preserve_cursor` — identity wins, vanished ticket falls back to same-position-clamped, never teleports. Both tickets' test suites green together (37 unit + 2 live).

blooop:
## Resolution

Build 2 is on main: **8a070e2** (fuzzy filter + keybinding skeleton) and **e916900** (cursor preservation pinned to `refresh::preserve_cursor`). Typing narrows the live list per the main-screen resolution (#9).

**What was built**

- `src/filter.rs` — nucleo matching (`nucleo-matcher 0.3`, mined from `prototype/mainscreen`) scored against `"repo #num title"`, so typing a repo name narrows too; empty query matches all.
[...]

PR #85, opened 2026-08-07
Build 13 — one glyph vocabulary: stage counts in cluster headers, rollups on branch roots (#78)

### LINKED ISSUE #78: Build: one glyph vocabulary — stage counts in cluster headers, rollups on branch roots
## What to build

Two spec gaps the [Build 12 review](https://github.com/blooop/wayfinder/pull/75#pullrequestreview) found and accepted rather than closed, so the tree now speaks two glyph vocabularies at once.

**1. The cluster header still counts statuses, not stages.** [#61](https://github.com/blooop/wayfinder/issues/61) settled that "the status glyph column *becomes* the stage glyph", but `cluster_header` still renders `○n ◐n ⊘n ●n` from `map.counts()` (frontier/claimed/blocked/done) with its own hardcoded glyph array and parallel colour table. The same characters now mean stages on the rows below it, so **a ticket drawn `!` in its row is counted under `○` in the header** — and `◍`/`!` can never appear in a header at all. The header should count stages, through the same `RowGlyph` mapping the rows use, with no second copy of the table.

**2. Branch-root rows carry no rollup.** [#62](https://github.com/blooop/wayfinder/issues/62) settled that "collapsed **and branch-root** rows carry stage counts of what's beneath them". Only shut groups got them. Branch roots are also the only place the double-count hazard actually lives — a takeable dependent renders under its root *and* as its own root — so this is where "counted once per node via the rendered tree" has to be earned rather than asserted (the review killed a test that claimed it against a fixture that could not exercise it).

## Seams (pre-agreed — test at these, no others)

1. **View planning** (`src/view.rs`) — the body item list: rollups present on branch-root items, counted once per node when a node is reachable twice in the DAG. Same seam the existing view-planning tests use.
2. **Header rendering** (`src/ui.rs` + whatever count function replaces `Map::counts`) — stage counts derived through `RowGlyph`, asserted on real content.

## Acceptance criteria

- [ ] Cluster headers count stages; `◍` and `!` can appear in a header; no second copy of the glyph or colour table exists
- [ ] A ticket whose stage is `needs attention` is counted as `!` in its cluster header, not `○`
- [ ] Branch-root rows carry stage rollups of their subtree
- [ ] A node reachable twice in the DAG is counted once in a branch-root rollup — with a fixture that genuinely renders it twice, so the test can fail
- [ ] `cargo test`, `cargo clippy --all-targets`, `cargo fmt --check` green

## Out of scope

- The review's non-blocking cleanups (`hidden`/`rollup` dual representation, `Route`'s repeated switch, `Ticket::stage()` data clump, `route`'s totality leaning on a caller) — worth doing, but they are refactors, not this behavior gap

blooop:
Closes #78.

The tree spoke two glyph vocabularies at once. Both halves of that are now one.

## 1. Cluster headers count stages

`cluster_header` kept its own four-status tally — `map.counts()` through
`Status::group()`, with a hardcoded `['○', '◐', '⊘', '●']` array and a parallel
colour table beside it. The same characters meant stages on every row below, so
a ticket the row drew `!` was counted under `○` in the header above it, and
`◍`/`!` could never appear in a header at all.

The header now tallies through `RowGlyph` — the mapping the rows are already
drawn from — and colours through the same `glyph_style`. `Map::counts` and
`Status::group` are gone; `RowGlyph::tally` is the one place counts-by-glyph is
computed, shared with the shut-group rollups that had been doing it inline.

Glyphs the map has nobody in drop out rather than showing a zero, and the order
is the glyph display order (`○ ◐ ◍ ! ● ⊘`) rather than the old status order —
so `⊘` now trails the stages instead of sitting third.

## 2. Branch-root rows carry rollups

Only shut groups had them; #62 settled that branch roots do too, and branch
roots are where the double-count hazard actually lives. The leverage walk draws
a dependent under every root that unblocks it, so a diamond in the DAG — two
dependents of one root that both unblock the same ticket — genuinely renders
that ticket twice inside a single branch.

`Branch::{Plain, Root { rollup }}` follows the shape `Fold::Shut { rollup }`
established last build: a row that heads nothing cannot carry a rollup at all.
`attach_rollups` fills it in as one pass over the finished item list rather than
something the walks carry down — "what is beneath this row" is a fact about the
*rendered* tree, and the rendered tree is exactly that list, so leverage, forest
and flattened all get the same answer from the same code and there is no second
notion of a subtree to drift from the one on screen. Each node is kept once
however many times the walk drew it.

```
▌ wayfinder · Map: wf  ○1  ◐1  ●1  ⊘2
  ▶ ○ #6 Re-entry breadcrumbs [task] (beneath) ⊘2
    ├─  ⊘ #7 Supervising AFK agents [task]
    └─  ⊘ #14 Breadcrumb markers [task]
    ◐ #9 Main screen design [task] (beneath) ⊘1
    └─  ⊘ #14 Breadcrumb markers [task]
    ▸ ● 1 done (hidden) ●1
```

## Tests, at the two seams the ticket pre-agreed

**View planning** (`src/view.rs`)

- `a_branch_root_carries_the_stage_rollup_of_its_subtree` — catches a rollup
  that never arrives, one computed over the wrong set, and a rollup appearing on
  a row that heads nothing. Claimed #9 is drawn inside #6's branch *and* as its
  own root: #6 counts it, #9-as-root carries nothing.
- `a_node_the_branch_renders_twice_is_counted_once` — the diamond, inside one
  root's branch. It pins the premise first (`#10` really is drawn twice) so the
  test can fail, and it does: with the dedup removed it reports `⊘4` against the
  expected `⊘3`. This is the test the last review killed for asserting the rule
  against a fixture that could not exercise it.

**Header rendering** (`src/ui.rs`)

- `the_cluster_header_counts_stages_not_statuses` — a ready ticket pushed to
  needs-attention by a failing-checks PR and a claimed one pushed to in-review
  by an approved one. Catches exactly the reported bug: counting statuses puts
  the first back under `○` and gives the second no glyph at all.
- `the_cluster_header_names_the_map_and_carries_the_counts` (existing) updated
  to the glyph display order.
- `the_default_screen_is_the_leverage_view_not_the_groups` (existing) gained the
  two branch-root rollups, so the plan's rollup reaching the screen is asserted
  rather than assumed.

`RowGlyph::tally`'s existing model test moved to the new return shape.

## Notes

- Rebased onto `main` after #83 landed the same `items_after_statements` hoist
  this branch had independently; the duplicate commit dropped clean.
- README's screen documentation described the header as per-status counts and
  said in as many words that they were *not* the stage the rows drew below —
  the exact split this closes. Updated, along with the branch-root rollup.
- `group_line` still hardcodes `⊘`/`●` for the two group *kinds*. That is a
  label for what the group is, not a stage tally, and `GroupKind::Done` draws
  its `●` undimmed on purpose — left alone rather than folded in on the way past.

Verified locally at the CI flags (`RUSTFLAGS=-D warnings`): `cargo fmt --all
--check` clean, `cargo clippy --all-targets --all-features --locked` clean,
`cargo test --locked --lib --bins --examples` 159 + 5 passed, live tests
compile, docs clean.

## Summary by Sourcery

Unify glyph-based stage counting across cluster headers and rollups, and add stage rollups to branch-root rows based on the rendered tree structure.

New Features:
- Display stage rollups on branch-root ticket rows showing the composition of their subtrees.

Bug Fixes:
- Ensure cluster headers count ticket stages using the same glyph vocabulary as row glyphs, fixing mismatches between header counts and row stages.

Enhancements:
- Share a single glyph tallying mechanism between headers and rollups, and attach branch rollups in a post-planning pass over the rendered item list.

Documentation:
- Update README to describe stage-based counts, branch-root rollups, and clarify colour usage for stage glyphs.

Tests:
- Add and update tests around branch-root rollups, glyph-based tallies, and cluster header rendering to cover the new behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,10 +75,11 @@ Only repos with an open `wayfinder:map`-labelled issue show tickets; other
 checkouts stay cached but hidden.
 
 **Every open map renders as its own cluster** — a `▌ repo · map title` header
-carrying the per-**status** counts (`○` frontier `◐` claimed `⊘` blocked `●`
-done). Those are the ticket's own state, not the stage its rows draw below.
-A repo can keep several maps open at once and each gets its own cluster;
-focusing a project shows all of them.
+carrying the map's **stage** counts, in the same glyph vocabulary the rows
+below it use (`○ ◐ ◍ ! ● ⊘`, described under the leading glyph). A stage the
+map has nobody in is left out rather than shown as a zero. A repo can keep
+several maps open at once and each gets its own cluster; focusing a project
+shows all of them.
 
 **Clusters are ordered by activity, with finished maps last**: the map issue
 touched most recently sits at the top, and a map whose every ticket is done
@@ -130,7 +131,7 @@ descend and the depth axis is something you can see:
 ```
 
 Orange because it is the one hue the screen does not otherwise spend — cyan is
-cluster headers and the prompt, green/yellow/red the status glyphs and counts,
+cluster headers and the prompt, green/yellow/red the stage glyphs and counts,
 magenta the PR badges, dim everything settled — so the selection never competes
 with something that means something else. The branch furniture stays uniformly
 dim: it is structure, not status.
@@ -145,6 +146,13 @@ the cursor on one and `→` opens it in place (`▾`), listing what it held as r
 you can select and launch, and `←` shuts it again — so nothing is ever merely a
 number you cannot reach. A map with nothing takeable
 leaves the body entirely, counted on the bottom line as `· N idle maps hidden`.
+
+**A row that heads a branch carries the same rollup of what is beneath it**,
+written `(beneath) ⊘2` at the end of the line — so a takeable ticket says the
+shape of what taking it unlocks without your reading the subtree. A node is
+counted once however many times the tree drew it: the leverage view draws a
+dependent under every root that unblocks it, so a diamond in the DAG genuinely
+renders one ticket twice inside a single branch.
 
 **`tab` shows the structure forest** instead: the whole blocking DAG, done
 tickets dimmed in place. A ticket's tree parent is its lowest-numbered in-map
--- a/src/model.rs
+++ b/src/model.rs
@@ -54,19 +54,6 @@ pub enum Status {
     Done,
 }
 
-impl Status {
-    /// Position of this status within the cluster header's counts
-    /// (frontier / claimed / blocked / done).
-    pub fn group(&self) -> usize {
-        match self {
-            Status::Frontier => 0,
-            Status::Claimed => 1,
-            Status::Blocked { .. } => 2,
-            Status::Done => 3,
-        }
-    }
-}
-
 /// Where a node stands in its lifecycle — one five-value lattice for every
 /// node (#61), **derived, never declared**: from its linked PRs when any are
 /// open or merged, from its ticket state otherwise. Blocked is deliberately
@@ -110,6 +97,22 @@ impl RowGlyph {
                 RowGlyph::Stage(stage(&ticket.prs, &ticket.status))
             }
         }
+    }
+
+    /// Tally a set of tickets by glyph: pairs in display order, one entry per
+    /// ticket, and glyphs nothing is in left out entirely.
+    ///
+    /// The one place counts-by-glyph is computed. Cluster headers and rollups
+    /// are the same question asked of different sets of tickets, so they are
+    /// the same call — a second tally would be a second glyph vocabulary, which
+    /// is exactly what #78 exists to remove.
+    pub fn tally<'a>(tickets: impl IntoIterator<Item = &'a Ticket>) -> Vec<(RowGlyph, usize)> {
+        let mut counts: std::collections::BTreeMap<RowGlyph, usize> =
+            std::collections::BTreeMap::new();
+        for ticket in tickets {
+            *counts.entry(RowGlyph::of(ticket)).or_default() += 1;
+        }
+        counts.into_iter().collect()
     }
 
     /// The character drawn: `○` ready · `◐` building/in progress · `◍` in
@@ -481,14 +484,11 @@ impl Map {
             .any(|t| !matches!(t.status, Status::Done))
     }
 
-    /// Ticket counts by status group (frontier / claimed / blocked / done) —
-    /// the cluster header's `○n ◐n ⊘n ●n`.
-    pub fn counts(&self) -> [usize; 4] {
-        let mut counts = [0; 4];
-        for t in &self.tickets {
-            counts[t.status.group()] += 1;
-        }
-        counts
+    /// The whole map tallied by glyph — the cluster header's counts (#78).
+    /// Stages, through the same [`RowGlyph`] the rows are drawn from, so a
+    /// ticket drawn `!` is counted under `!`.
+    pub fn tally(&self) -> Vec<(RowGlyph, usize)> {
+        RowGlyph::tally(&self.tickets)
     }
 }
 
@@ -1050,7 +1050,7 @@ mod tests {
     }
 
     #[test]
-    fn counts_tally_the_four_status_groups() {
+    fn a_map_tallies_by_glyph_in_display_order() {
         let mut done = ticket(2, false, vec![]);
         done.status = Status::Done;
         let mut claimed = ticket(9, true, vec![]);
@@ -1062,6 +1062,14 @@ mod tests {
             last_activity: None,
             tickets: vec![ticket(6, true, vec![]), claimed, blocked, done],
         };
-        assert_eq!(map.counts(), [1, 1, 1, 1]);
+        assert_eq!(
+            map.tally(),
+            vec![
+                (RowGlyph::Stage(Stage::Ready), 1),
+                (RowGlyph::Stage(Stage::Building), 1),
+                (RowGlyph::Stage(Stage::Done), 1),
+                (RowGlyph::Blocked, 1),
+            ]
+        );
     }
 }
--- a/src/ui.rs
+++ b/src/ui.rs
@@ -15,7 +15,7 @@ use ratatui::Frame;
 use crate::app::{App, Overlay, Scope};
 use crate::launch::Launch;
 use crate::model::{Checks, Map, MapId, PrLink, PrStatus, Review, RowGlyph, Stage, Status, Ticket};
-use crate::view::{Fold, GroupKind, Item, Plan, Screen};
+use crate::view::{Branch, Fold, GroupKind, Item, Plan, Screen};
 
 /// One colour per glyph meaning, shared by the row column and the rollup
 /// pairs: calm colours for the flowing stages, red for the two that demand
@@ -56,6 +56,28 @@ fn cluster_header(id: &MapId, map: &Map) -> Line<'static> {
         ));
     }
     Line::from(spans)
+}
+
+/// A rollup's trailing spans: a dim word for what the counts are *of*, then
+/// the glyph+count pairs in display order, each in the colour its glyph means.
+///
+/// One function, because a collapsed group and a branch root are asking the
+/// same question of different sets of rows — what is under here, by stage — and
+/// two renderings of one answer is how the screen ended up speaking two glyph
+/// vocabularies in the first place (#78).
+fn rollup_spans(label: &str, rollup: &[(RowGlyph, usize)]) -> Vec<Span<'static>> {
+    let mut spans = vec![Span::styled(
+        format!(" ({label})"),
+        Style::new().add_modifier(Modifier::DIM),
+    )];
+    for &(glyph, count) in rollup {
+        spans.push(Span::raw(" "));
+        spans.push(Span::styled(
+            format!("{}{count}", glyph.char()),
+            glyph_style(glyph),
+        ));
+    }
+    spans
 }
 
 /// The cursor marker, which sits **after** a row's tree furniture and directly
@@ -138,10 +160,16 @@ fn pr_badge(ticket_repo: &str, pr: &PrLink) -> Vec<Span<'static>> {
 /// extra blocking edges the tree position cannot show (`⤷ also needs #n`). The
 /// flattened screen has no cluster header above the row, so the row names its
 /// repo itself.
+///
+/// A row that heads a branch closes with the stage rollup of it (#62), so the
+/// shape of a subtree reads off its root without walking the subtree — the
+/// same glyph+count pairs a collapsed group carries, because they mean the
+/// same thing.
 fn ticket_line(
     ticket: &Ticket,
     prefix: &str,
     also_needs: &[u64],
+    branch: &Branch,
     name_repo: bool,
     under_cursor: bool,
 ) -> Line<'static> {
@@ -183,6 +211,9 @@ fn ticket_line(
             Style::new().add_modifier(Modifier::DIM),
         ));
     }
+    if let Branch::Root { rollup } = branch {
+        spans.extend(rollup_spans("beneath", rollup));
+    }
     let style = match ticket.status {
         Status::Done => Style::new().add_modifier(Modifier::DIM),
         _ => Style::new(),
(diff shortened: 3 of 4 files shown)


blooop (commented):

Reviewed at merge-base `50f4992` → head `49945ca` (`git diff 50f4992...HEAD`), 320 insertions across `README.md`, `src/model.rs`, `src/ui.rs`, `src/view.rs`. Two independent axes, run in parallel contexts and reported unmerged: code that follows every standard while implementing the wrong thing must fail Spec and pass Standards, and the reverse.

Preflight: fixed point resolves, diff non-empty, `CI` and `Package` both success on the full-40-char head SHA.

## Reviewer's own verification

The PR claims criterion 4 was mutation-checked. **Verified rather than trusted**, in a detached worktree off `49945ca`: removing the dedup guard `if !beneath.contains(row)` (`src/view.rs:246`) and running `cargo test --lib` fails **exactly one** test —

```
---- view::tests::a_node_the_branch_renders_twice_is_counted_once ----
assertion `left == right` failed: three nodes beneath #6, not the four rows drawn for them
  left:  [(6, Root { rollup: [(Blocked, 4)] })]
  right: [(6, Root { rollup: [(Blocked, 3)] })]
test result: FAILED. 158 passed; 1 failed
```

The `⊘4` vs `⊘3` claim in the PR body is accurate. The premise assertion `stops(&plan, &m) == vec![6, 7, 10, 8, 10]` is an exact `Vec` equality carrying both `10`s, so it cannot pass vacuously — the fixture genuinely renders the node twice, and the test can genuinely fail. **This is the substantive difference from the test #75's review killed**, and it is earned.

Also confirmed independently: `Map::counts` and `Status::group` have zero remaining references in `src/`, `examples/` or `tests/`; three call sites now share one `RowGlyph::tally`; no added line exceeds 100 columns.

## Standards

### Blocking

**`src/ui.rs:241` — a second copy of the colour table survives, and it is the one arm with no defence.**

```rust
GroupKind::BlockedDeeper => ('⊘', "blocked deeper down", Color::Red),
GroupKind::Done          => ('●', "done", Color::Reset),
```

The `BlockedDeeper` arm re-derives by hand exactly the pair `glyph_style(RowGlyph::Blocked)` already owns (`src/ui.rs:30` — Red). Same glyph, same meaning, same colour, drawn adjacent to a count of blocked tickets. The PR's defence — "a label for what the group *is*, not a stage tally" — is **rejected for this arm**: there is no semantic daylight between this `⊘` and the `⊘` on the rows below it, and a future recolour of blocked in `glyph_style` leaves this line stale. That is precisely the drift #78 exists to close, shrunk from four rows to one but not removed. Acceptance criterion 1 is written as an *existence* check ("no second copy of the glyph or colour table exists"), not a regression check; waiving this makes it unfalsifiable.

The `GroupKind::Done` arm is **accepted**: `glyph_style(Stage::Done)` is `DIM`, and drawing the group's own `●` undimmed is a deliberate and correct difference. (Minor: expressing "not dim" as `Color::Reset` conflates two style axes and only works because nothing else colours those spans.)

One edit, no behaviour change, asymmetry stays visible:

```rust
let (glyph, label, style) = match kind {
    GroupKind::BlockedDeeper => (RowGlyph::Blocked, "blocked deeper down", glyph_style(RowGlyph::Blocked)),
    GroupKind::Done => (RowGlyph::Stage(Stage::Done), "done", Style::new()),
};
```

### Non-blocking

- **`src/model.rs:83-89` — the six-glyph display order is load-bearing but not pinned.** `RowGlyph::tally` orders through `BTreeMap`, i.e. derived `Ord`, i.e. declaration order; three render sites now depend on it. Verified by experiment: swapping `Stage::Building` and `Stage::InReview` in the declaration leaves 158 of 159 tests green — the only failure is `the_node_takes_the_max_over_its_open_prs`, a PR-semantics test. The `◐`-before-`◍` link in the documented `○ ◐ ◍ ! ● ⊘` order is held only by accident, by `Stage::Ord`'s *other* job. One assert closes it: sort all six `RowGlyph`s, map `.char()`, `assert_eq!(s, "○◐◍!●⊘")`. Adjacent swaps elsewhere *are* caught (`model.rs:1065`, `ui.rs:630`, `ui.rs:674`).
- **`src/view.rs:260` — bare `Index` against the file's own idiom.** `maps[&row.map]` and `.tickets[row.index]` panic mutely; `view.rs:366/441/565` all carry `expect("… is one of map.tickets")`. `attach_rollups(items, clusters)` takes the two as independent parameters, so "these items were planned from these clusters" is comment-enforced. Total in practice — `plan()` is the only caller — so an `expect` message, not a refactor.
- **`src/view.rs:261` — `if let Item::Ticket { branch, .. } = &mut items[i]`** silently no-ops on an index that cannot be a non-Ticket. Swallowed impossible case; `else { unreachable!() }`, or carry `&mut Branch` instead of `usize`.
- **`src/ui.rs:168-175` — data clump.** `prefix`, `also_needs`, `branch` are three of `Item::Ticket`'s four non-`row` fields, destructured at `ui.rs:290` and re-passed one by one; this PR added the fourth member. Under clippy's 7-arg threshold, so nothing enforces it. A `TicketRow` struct inside `Item::Ticket` collapses it.
- **`src/model.rs:490` — `Map::tally` is a Middle Man.** One line, one caller, and `RowGlyph::tally(&map.tickets)` is no longer at the call site. Mildly against the PR's own "one place" thesis: a reader now has two names for one computation.
- **`README.md:150` — overclaim.** "A row that heads a branch carries the same rollup of what is beneath it" is false on the forest screen: probed, forest `#7` at depth 1 with `#14` drawn beneath it gets `Branch::Plain` — only depth-0 rows get rollups. The `Branch` doc comment (`view.rs:148`) is accurate ("sits at the top of its cluster"); the README's bolded lead is looser than the code. "A **top-level** row…" fixes it.
- **The forest screen gained unasserted rendered content.** `tab_shows_the_structure_forest_with_done_in_place` (`ui.rs:713`) now renders `(beneath) ⊘2` on `#6` and asserts nothing about it. Outside the ticket's two pre-agreed seams, so not required — but a behaviour change reaching a second screen with no test pinning it.
- **`src/ui.rs:217-221` — a done branch root dims its own rollup.** `Line::from(spans).style(DIM)` applies under the rollup's span colours. Only reachable on the forest screen, where a done ticket can be a root. Cosmetic.

### Affirmative findings (checked, clean)

- **`Branch` does achieve the `Fold::Shut` shape.** `Branch::Plain` carries no rollup slot at all — no nullable, no parallel field, no sentinel. `Root { rollup: vec![] }` is representable but never constructed (`view.rs:257` skips empties) — and `Fold::Shut { rollup: vec![] }` has the identical hazard guarded identically at `push_group`'s `held.is_empty()` early return. A `NonEmpty` would close it; per the skill's "when NOT to apply", that is precision-chasing with no panic behind it. **Explicitly judged not worth fixing.**
- **The two-pass build is legitimate, not a patch-later defect.** `Branch::Plain` is an inhabited terminal meaning most rows keep, not an `Unknown`/`Option` placeholder. The pass runs inside `plan()` (`view.rs:219`), the module's only builder, so no caller observes an intermediate. Residual (low): `Plan` derives `Default` with `pub items`, so nothing in the type obliges a future direct constructor to run `attach_rollups`.
- **No other surviving glyph or colour site.** All of `src/` grepped for `○ ◐ ◍ ! ● ⊘` and every `Color::`/`Modifier::DIM`: outside `RowGlyph::char` and `glyph_style`, the only production hits are `ui.rs:241-242` (above) and `ui.rs:141/143` (`✗`/`✓` PR-verdict badges — a genuinely different vocabulary).
- **`attach_rollups` is correct on all three screens**, probed empirically. Flattened → every row depth 0, all `Plain`, no rollups. Forest → depth-0 roots only, orphan-sweep roots included. Leverage → correct; an *expanded* group's depth-1 held rows hit `open == None` and are dropped, which is the documented intent (they hang from the group's own `Fold`). A depth>0 row with no open branch is reachable, but only in that one case, and it is the right answer rather than a silent wrong one. `O(n²)` `beneath.contains` is bounded by one subtree — non-issue.
- **House rules pass.** `cargo fmt --check`, `clippy --all-targets -D warnings`, and `cargo doc --no-deps` under `RUSTDOCFLAGS=-D warnings` all silent in a clean copy — no broken intra-doc links. All new tests in `#[cfg(test)]` modules in `src/`; nothing added under `tests/`.

## Spec

| # | Criterion | Verdict |
|---|---|---|
| 1 | "Cluster headers count stages; `◍` and `!` can appear in a header; no second copy of the glyph or colour table exists" | **PARTIAL** |
| 2 | "A ticket whose stage is `needs attention` is counted as `!` … not `○`" | **MET** |
| 3 | "Branch-root rows carry stage rollups of their subtree" | **MET** |
| 4 | "A node reachable twice in the DAG is counted once … with a fixture that genuinely renders it twice, so the test can fail" | **MET** |
| 5 | `cargo test`, `cargo clippy --all-targets`, `cargo fmt --check` green | **MET** |

**1 — PARTIAL.** The header half is fully done: `cluster_header` iterates `map.tally()` and styles via `glyph_style` (`ui.rs:52-73`), `Status::group` and `Map::counts` are deleted, `RowGlyph::tally` (`model.rs:109`) is the single tally and is reused by `push_group` (`view.rs:329`). The clause *"no second copy of the glyph or colour table exists"* is not satisfied — see the Standards blocking finding at `ui.rs:241`. Both axes reached this line independently; they differ on whether it blocks, and I side with Standards for the reason given there.

**2 — MET.** `the_cluster_header_counts_stages_not_statuses` (`ui.rs:636`). #6 is `Frontier` + `Open{Failing, Required}` → `Signal::Red` → `Stage::NeedsAttention` (`model.rs:143-165`), so the fixture genuinely produces `!`. Against the old code the header rendered `○1 ◐1 ⊘2 ●1`; the assertion demands `◍1  !1  ●1  ⊘2`, so it fails on the old code for exactly the reported reason. The `contains` on `"wf  ◍1"` also pins zero-suppression — a leading `○0` breaks the substring.

**3 — MET, on screen not merely in the plan.** `Branch::{Plain, Root}` (`view.rs:164`), rendered at `ui.rs:215`, asserted through `render()` at `ui.rs:697` and `ui.rs:701` (`(beneath) ⊘2` / `⊘1`).

**4 — MET.** `walk_unblocks` (`view.rs:422-467`) filters only on `path`, never on `reached`, so `#6→{7,8}` each `→#10` pushes `#10` twice. Confirmed by execution — see the mutation result above. `a_branch_root_carries_the_stage_rollup_of_its_subtree` (`view.rs:991`) is **not** redundant with it: it alone covers a multi-glyph rollup and the negative case (`#9` drawn as its own root carries `Branch::Plain`).

**5 — MET.** Green on CI at this head; independently reproduced locally in a clean copy.

### Declared deviations

**(a) Glyph order changed to display order, zero counts dropped — acceptable, not scope creep.** Criterion 1 requires counting *"through the same `RowGlyph` mapping the rows use, with no second copy of the table"*. `RowGlyph`'s `Ord` **is** display order (`model.rs:83-88`); preserving the old status order or padding zero counts would require a header-specific tally — the second copy the criterion forbids. The deviation falls out of the criterion rather than around it, and README is updated to match.

**(b) `group_line`'s hardcoded `⊘`/`●` — not acceptable for the `BlockedDeeper` arm.** The Spec axis read the ticket's clause as scoped to `cluster_header`'s "own hardcoded glyph array and parallel colour table" and called it acceptable-but-narrowly, noting the residual honestly. The Standards axis rejected the `BlockedDeeper` arm outright. The checkbox is written unscoped and as an existence test, and this arm has no semantic distinction from the row glyph to stand on. Blocking, per Standards.

### Scope

No creep found. Tests sit at exactly the two pre-agreed seams; the one model-level change (`model.rs:1050`) is an existing test adapted to the new return type, not a new seam. Nothing added under `tests/`. The README edit was made mandatory by the removed text ("not the stage its rows draw below").

Out-of-scope items from #78 (`hidden`/`rollup` dual representation, `Route`'s repeated switch, `Ticket::stage()` data clump, `route`'s totality) were not assessed.

## Verdict

1. **`src/ui.rs:241`** — `GroupKind::BlockedDeeper => ('⊘', …, Color::Red)` is a surviving second copy of the glyph *and* colour table, duplicating `glyph_style(RowGlyph::Blocked)` exactly. It fails acceptance criterion 1's *"no second copy of the glyph or colour table exists"*, which is an existence check rather than a regression check. The `GroupKind::Done` arm beside it is fine and should stay as it is — the fix is the one arm, one edit, no behaviour change.

This is a narrow block on otherwise strong work. The hard part of this ticket — criterion 4 — is genuinely earned: the fixture renders the node twice, the premise is pinned, and the test fails when the dedup is removed. `Map::counts` and `Status::group` were deleted rather than wrapped, and `Branch::{Plain, Root}` really does make a rollup on a row that heads nothing unrepresentable. Everything in the Non-blocking list can be left or folded into a follow-up; only finding 1 stands between this and approval.

Issue #20, opened 2026-08-04
Readable agent tab names: title in the label, identity in the key


blooop:
## Question

Make agent tab names readable: `<repo>#<n> <capped title>` instead of the bare `<repo>#<n>` from [#5](https://github.com/blooop/wayfinder/issues/5)/[#7](https://github.com/blooop/wayfinder/issues/7). Surfaced while resolving [Auto-start AFK frontier tickets](https://github.com/blooop/wayfinder/issues/18); independent of auto-start, and an improvement to the Build 4 seam on its own.

The shape is decided (see #18's resolution); the work is the split it requires.

- **Identity must not be the label.** `tab_exists` is called with the *generated* name (`launch.rs:467`), so a title riding inside that name means a retitled issue generates a different name, misses its own tab, and duplicates it. Split into `tab_key(ticket)` → `<repo>#<n>`, used for focus-or-create lookup and `is_agent_tab`, never changing; and `tab_label(ticket)` → key + capped title, passed only to `new-tab --name`. The prefix-tolerance already in `tab_exists` (built for zellij's activity markers) is the right matcher — pass it the key.
- **Repo prefix stays unconditional.** It does not earn its width in the bar — a bar is one session and a session is one project (#5), so the repo can never disambiguate there — but it makes the name self-describing in `zellij list-sessions`, in `ps`, and in logs, and it keeps one matcher. A prefix conditional on launch scope was considered and rejected: `ctrl-g` flips scope mid-session (#9), so a scope-at-creation-time label drifts against the scope being viewed.
- **Cap the title** (~18 chars, word boundary). Zellij truncates anyway, so a name that only reads well untruncated does not read at all.

Open to decide in the build, small: whether `agent_tab_count` needs any change (it queries per session and knows the repo from the session, so probably not).

blooop:
**breadcrumb:** three decisions before writing code.

**1. `launch.rs:467` is exactly right** — `create_or_focus_tab`'s `if tab_exists(&query_tab_names(session).await?, tab)` is on line 467 of `src/launch.rs` at 3c5af65, and `tab` there is the same string that later goes to `new-tab --name`. Today key == label so nothing is broken yet; the defect is latent and would land the moment a title enters that string. No line-number drift to report.

[...]

PR #89, opened 2026-08-07
Keep an untouched cursor on the top row as maps stream in

### LINKED ISSUE #88: Opening wf lands the cursor on the second cluster, not the first
## What to build

**Reported against 0.5.0**: opening `wf` inside a checkout lands the cursor on the *second* cluster, not the first.

The default cursor position is treated as a user's choice. `App::replace_clusters` anchors on `cursor_key()` and hands it to [`refresh::preserve_cursor`](https://github.com/blooop/wayfinder/blob/main/src/refresh.rs), whose contract is *"identity wins over position — a map arriving must never teleport the selection just because rows moved."* That is right for a selection someone made, and wrong for the position the cursor merely starts at.

During streaming startup the clusters arrive one fetch at a time. The first to land takes the cursor at index 0. When a map with **newer** activity arrives, [#82](https://github.com/blooop/wayfinder/pull/82)'s ordering (work before history, newest first) sorts it above — and the cursor dutifully follows its anchored row *downwards*. On this repo, with five open maps streaming in, the cursor ends up on whichever map's fetch returned first, which is generally not the one drawn at the top.

Neither ingredient is wrong on its own. The identity-preserving cursor ([#50](https://github.com/blooop/wayfinder/issues/50)/[#57](https://github.com/blooop/wayfinder/issues/57)) and activity ordering (#82) simply collide, because nothing distinguishes *"the cursor has not been moved"* from *"the user chose row 0"*.

### Failing test (verified against `a3f88c8`)

Drop into `src/app.rs`'s test module — it fails today with `left: 1, right: 0`:

```rust
#[test]
fn an_untouched_cursor_stays_on_the_top_row_as_maps_stream_in() {
    let mut app = App::new(BTreeMap::from([cluster(
        "blooop/slow", 1, Some("2026-08-01T00:00:00Z"), true,
    )]));
    assert_eq!(app.cursor_pos(), 0);

    let mut both = BTreeMap::new();
    both.extend([cluster("blooop/slow",  1, Some("2026-08-01T00:00:00Z"), true)]);
    both.extend([cluster("blooop/fresh", 2, Some("2026-08-07T00:00:00Z"), true)]);
    app.replace_clusters(both);

    // the fresher map now renders first
    assert_eq!(app.cursor_pos(), 0);   // actual: 1, still on blooop/slow
}
```

## The modelling question this turns on

A sentinel will not do: `cursor == 0` cannot mean "untouched", because the user can deliberately sit on row 0 and must then be pinned like any other choice. Per this map's `/constructive-modeling` constraint the distinction wants to be a **sum type** — something along the lines of

```rust
enum Cursor {
    /// Never moved. Always means "the top of the list", whatever the list becomes.
    Untouched,
    /// Put here deliberately; identity is preserved across reorders.
    Chosen(StopKey),
}
```

so that `replace_clusters` preserves identity only in the `Chosen` case, and the untouched cursor re-derives to the first stop every time. Note `main.rs` already has the matching precedent for *scope*: a `focus` that is dropped the moment the user changes scope themselves, precisely so a pending default cannot overwrite a real choice.

Whether that is the right shape — and whether "the top" means the first stop or the first *takeable* stop — is the ticket's decision.

## Acceptance criteria

- [ ] Opening `wf` in a checkout leaves the cursor on the **first** row of the **first** cluster, however many maps stream in and in whatever order
- [ ] A cursor the user has moved is still pinned by identity across a reorder or refetch (the #50/#57 behaviour does not regress — its existing tests keep passing)
- [ ] Deliberately selecting the top row and then refetching keeps the user pinned to that row, not silently re-defaulted — no sentinel standing in for "untouched"
- [ ] The failing test above passes, and a test covers the user-chosen case reordering underneath them
- [ ] `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` green

## Out of scope

- Changing the cluster ordering from #82, or the streaming-load design. The ordering is right; what is wrong is what the cursor does while the list is still filling.

blooop:
Opening `wf` in a checkout landed the cursor on the **second** cluster.

Two correct behaviours collided. The cursor is pinned by identity across a refetch (#50/#57), and clusters render work-first, newest-active-first (#82). Maps stream in one fetch at a time, so the first to land took the cursor at index 0 — and when a busier map arrived and sorted above it, the anchored row carried the selection downwards.

`refresh::preserve_cursor` is untouched. Its contract — *identity wins over position* — was never wrong; it was being asked the wrong question. That contract is right for a row someone picked and meaningless for the position the cursor merely starts at, and nothing distinguished the two.

## The shape

The cursor now carries **how it got where it is**:

```rust
enum Cursor {
    Untouched,      // means "the first stop on screen", whatever the screen becomes
    Chosen(usize),  // put here deliberately; pinned by identity across a reorder
}
```

`Untouched` stores no position, because it does not have one — it is a *rule*, re-read from whatever list has arrived, so `cursor_pos` answers it from the current stops rather than from anything stored. Only `Chosen` is anchored in `replace_clusters`.

A flag beside the index would have left "untouched, and also remembering a position" representable — exactly the state that made a default indistinguishable from a choice. `Chosen` holds a position rather than a `StopKey` (the ticket's sketch) because the key is already re-derived from the live clusters at swap time, and the index is what carries `preserve_cursor`'s fallback for a stop that vanished outright; holding both would let the two disagree, which is a worse defect than the one being fixed.

"The top" is taken literally: **the first stop drawn, not the first *takeable* one** — as decided on the ticket — so an untouched cursor never skips past a cluster whose top rows happen to be blocked or done.

`tab`, `ctrl-f` and `ctrl-g` rebuild the list under the cursor and keep it on its stop. They now ask for an anchor only when there is a choice to anchor, so they cannot quietly promote a default into one — otherwise the bug walks back in through a key that was never about the selection.

## Tests

Four, all in `app.rs`'s existing `#[cfg(test)]` module:

- **`an_untouched_cursor_stays_on_the_top_row_as_maps_stream_in`** — the ticket's test, verbatim. Failed `left: 1, right: 0` before the change.
- **`a_chosen_row_stays_with_its_ticket_when_a_map_sorts_above_it`** — the #50/#57 behaviour that must survive: a row someone moved to slides down the screen rather than losing the selection.
- **`choosing_the_top_row_pins_it_rather_than_re_defaulting_to_the_top`** — the case a sentinel cannot serve. This cursor sits at position 0 exactly like a fresh one and must behave like its opposite. Verified discriminating: implementing the naive `Cursor::Untouched | Cursor::Chosen(0) => None` makes this test — and only this test — fail.
- **`toggling_the_lens_does_not_turn_an_untouched_cursor_into_a_choice`** — guards the scope/lens door. Verified discriminating: anchoring unconditionally makes this test, and only this test, fail.

## Checks

`cargo fmt --all --check`, `cargo clippy --all-targets --all-features --locked`, `cargo test --locked --lib --bins --examples` (163 passed), `cargo test --locked --all-targets --no-run` and `cargo doc` all green locally under `RUSTFLAGS=-D warnings`.

Cluster ordering and the streaming-load design are untouched, as the ticket scoped.

Closes #88

## Summary by Sourcery

Distinguish between an untouched and a user-chosen cursor to keep the initial selection on the top row while preserving identity-based cursor pinning as maps stream in.

Bug Fixes:
- Prevent the cursor from being dragged to the second cluster when new, busier maps stream in and reorder the list.

Enhancements:
- Introduce a Cursor enum to track whether the cursor is untouched or user-chosen, and derive its position accordingly.
- Ensure scope and lens toggles only anchor the cursor when it represents an explicit user choice, preserving default-top behavior for untouched cursors.

Tests:
- Add tests covering untouched cursor behavior as maps stream in, identity-pinned cursor behavior under reordering, pinning of a deliberately re-chosen top row, and lens toggling without promoting an untouched cursor into a choice.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -102,6 +102,40 @@ pub enum StopKey {
     Group(GroupId),
 }
 
+/// Where the cursor is, and — the part that matters — **whether anyone put it
+/// there** (#88).
+///
+/// Two facts used to be one `usize`, and index `0` had to stand for both of
+/// them: the position a fresh screen starts at, and a row the human deliberately
+/// picked. [`App::replace_clusters`] cannot serve both. A chosen row is pinned by
+/// identity, so a map arriving above it must never teleport the selection
+/// (#50/#57); a starting position has no identity to pin, and pinning it anyway
+/// is what dragged the cursor down to the second cluster as the first-arrived map
+/// was outranked by a fresher one.
+///
+/// A sum type rather than a flag beside the index, so the state "untouched, and
+/// also remembering a position" — the one that made a default indistinguishable
+/// from a choice — cannot be written down. [`Cursor::Untouched`] carries no
+/// index because it does not have one: it *means* the top of the list, and the
+/// top is re-read from whatever list is on screen now.
+///
+/// The chosen arm holds a position rather than a [`StopKey`] because the key is
+/// re-derived from the live clusters at the moment they are swapped, and the
+/// index is what carries `preserve_cursor`'s fallback for a stop that vanished
+/// entirely. Holding both would let the two disagree, which is a worse defect
+/// than the one this fixes.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Cursor {
+    /// Never moved. Always means "the first stop on screen", whatever the screen
+    /// becomes — taken literally: the first stop drawn, not the first *takeable*
+    /// one, so an untouched cursor never skips past a cluster whose top rows
+    /// happen to be blocked or done.
+    Untouched,
+    /// Put here deliberately, at this position in [`App::stops`]. Identity is
+    /// preserved across a reorder or a refetch.
+    Chosen(usize),
+}
+
 #[derive(Debug)]
 pub struct App {
     /// The clusters on screen: every open map that has arrived, keyed by id.
@@ -146,7 +180,7 @@ pub struct App {
     /// [`GroupId`], so an expansion survives a refetch, a query and a lens
     /// toggle — it is a choice about a *map*, not about a frame.
     expanded: Expanded,
-    cursor: usize,
+    cursor: Cursor,
 }
 
 impl App {
@@ -164,7 +198,7 @@ impl App {
             overlay: Overlay::None,
             lens: Lens::Leverage,
             expanded: Expanded::new(),
-            cursor: 0,
+            cursor: Cursor::Untouched,
         }
     }
 
@@ -295,9 +329,14 @@ impl App {
         }
     }
 
-    /// Cursor position clamped into the stop list.
+    /// Cursor position clamped into the stop list. An untouched cursor is not a
+    /// remembered position but a *rule* — the top of the list — so it is
+    /// answered from the stops on screen now rather than from anything stored.
     pub fn cursor_pos(&self) -> usize {
-        self.cursor.min(self.stops().len().saturating_sub(1))
+        match self.cursor {
+            Cursor::Untouched => 0,
+            Cursor::Chosen(pos) => pos.min(self.stops().len().saturating_sub(1)),
+        }
     }
 
     /// What the cursor is on, if anything is on screen.
@@ -339,13 +378,25 @@ impl App {
         self.cursor_stop().map(|stop| self.stop_key(&stop))
     }
 
+    /// The stop to hold on to while the list is rebuilt underneath the cursor —
+    /// a scope change, a lens toggle. `None` for an untouched cursor, which has
+    /// no stop to hold: it means the top of whatever list results, and anchoring
+    /// it to the row that merely happens to be first would turn a default into a
+    /// choice the human never made.
+    fn cursor_anchor(&self) -> Option<StopKey> {
+        match self.cursor {
+            Cursor::Untouched => None,
+            Cursor::Chosen(_) => self.cursor_key(),
+        }
+    }
+
     /// Point the cursor at a specific stop if it is on screen.
     fn point_at(&mut self, key: &StopKey) {
         let pos = self
             .stops()
             .iter()
             .position(|at| &self.stop_key(&at.stop) == key);
-        self.cursor = pos.unwrap_or(0);
+        self.cursor = Cursor::Chosen(pos.unwrap_or(0));
     }
 
     /// `↑`/`↓`: the next stop at the cursor's own depth, else simply the next
@@ -368,7 +419,9 @@ impl App {
     fn move_sibling(&mut self, delta: isize) {
         let stops = self.stops();
         if stops.is_empty() {
-            self.cursor = 0;
+            // Nothing on screen, so there is no stop to have chosen — pressing a
+            // key over an empty list is not a selection.
+            self.cursor = Cursor::Untouched;
             return;
         }
         let pos = self.cursor_pos();
@@ -383,7 +436,7 @@ impl App {
             let i = i as usize;
             adjacent.get_or_insert(i);
             if at.depth == depth {
-                self.cursor = i;
+                self.cursor = Cursor::Chosen(i);
                 return;
             }
             if at.depth < depth {
@@ -394,7 +447,7 @@ impl App {
         // `None` only when there is nothing at all in that direction, which is
         // the one case where holding still is the honest answer.
         if let Some(next) = adjacent {
-            self.cursor = next;
+            self.cursor = Cursor::Chosen(next);
         }
     }
 
@@ -414,7 +467,7 @@ impl App {
         }
         let pos = self.cursor_pos();
         if pos + 1 < self.stops().len() {
-            self.cursor = pos + 1;
+            self.cursor = Cursor::Chosen(pos + 1);
         }
     }
 
@@ -434,28 +487,43 @@ impl App {
         };
         if depth > 0 {
             if let Some(parent) = (0..pos).rev().find(|&i| stops[i].depth == depth - 1) {
-                self.cursor = parent;
+                self.cursor = Cursor::Chosen(parent);
                 return;
             }
         }
         if pos > 0 {
-            self.cursor = pos - 1;
+            self.cursor = Cursor::Chosen(pos - 1);
         }
     }
 
-    /// Swap in freshly fetched clusters, keeping query/scope/expansions intact
-    /// and the cursor pinned to stop identity (falling back to the same
-    /// position, clamped, if the stop vanished — see `refresh::preserve_cursor`).
+    /// Swap in freshly fetched clusters, keeping query/scope/expansions intact.
+    ///
+    /// What happens to the cursor depends on how it got where it is (#88). A
+    /// **chosen** stop is pinned by identity, falling back to the same position,
+    /// clamped, if it vanished — see `refresh::preserve_cursor`. An **untouched**
+    /// cursor has nothing to pin: it means the top of the list, so it re-derives
+    /// to the first stop of whatever just arrived. Maps stream in one fetch at a
+    /// time and the order leads on activity, so anchoring the start position too
+    /// would let the first map to land drag the cursor downwards the moment a
+    /// busier map sorted above it.
     pub fn replace_clusters(&mut self, clusters: BTreeMap<MapId, Map>) {
-        let anchor = self.cursor_key();
-        let old_index = self.cursor_pos();
+        let pinned = match self.cursor {
+            Cursor::Untouched => None,
+            Cursor::Chosen(_) => Some((self.cursor_key(), self.cursor_pos())),
+        };
         self.clusters = clusters;
-        let new_order: Vec<StopKey> = self
-            .stops()
-            .iter()
-            .map(|at| self.stop_key(&at.stop))
-            .collect();
-        self.cursor = crate::refresh::preserve_cursor(anchor.as_ref(), old_index, &new_order);
+        if let Some((anchor, old_index)) = pinned {
+            let new_order: Vec<StopKey> = self
+                .stops()
+                .iter()
+                .map(|at| self.stop_key(&at.stop))
+                .collect();
+            self.cursor = Cursor::Chosen(crate::refresh::preserve_cursor(
+                anchor.as_ref(),
+                old_index,
+                &new_order,
+            ));
+        }
     }
 
     /// The first enter (#62): stage a launch of the cursor's ticket by opening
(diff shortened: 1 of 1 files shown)


blooop (commented):

Reviewed in a fresh context that did not write the code. Fixed point `a3f88c8...5b5a230` (one file, `src/app.rs`, +209/−31); spec is #88 including its comments. Preflight: refs resolve, diff non-empty, `CI` + `Package` green on the full head SHA. The two axes below are **not merged or reranked** — each stands on its own.

## Standards

Primary standard is this map's `/constructive-modeling` constraint; Fowler's 12 smells applied as judgement on top.

**BLOCKING — `src/app.rs:399` (`point_at`): the diff fabricates a choice out of a miss.**

```rust
self.cursor = Cursor::Chosen(pos.unwrap_or(0));
```

This is a line the diff changed. The author had the new sum type in hand and encoded "the stop I was told to point at is not in this list" as `Chosen(0)` — a *deliberate* row 0. That is precisely the state the PR exists to abolish, and the PR's own doc comment eleven lines above says so: *"anchoring it to the row that merely happens to be first would turn a default into a choice the human never made."*

Reachable in two keystrokes, and I reproduced the full symptom rather than arguing it. The forest lens emits no group stops at all, so a chosen `Done`/`BlockedDeeper` group line has no counterpart across `Tab`:

```
chose:              cursor=Chosen(1)  (the Done group line)
after tab:          cursor=Chosen(0)  ← a default, recorded as a choice
after stream-in:    cursor=Chosen(1)  map=blooop/slow
```

The last line is #88's exact signature — a fresher map sorts above and drags the highlight off the top row — re-entered through the lens door. `ctrl-f`/`ctrl-g` share the shape.

What makes this blocking rather than a scope argument is that the codebase already contains the correct reasoning, four hundred lines away and in this same diff. `src/app.rs:422-424`, on the empty-list branch of `move_sibling`: *"Nothing on screen, so there is no stop to have chosen — pressing a key over an empty list is not a selection"* → `Cursor::Untouched`. A vanished anchor is the same fact and gets the opposite answer. The fix is one line — `self.cursor = pos.map_or(Cursor::Untouched, Cursor::Chosen);` — plus a guard, and `Untouched` is now available precisely so this function no longer has to invent a value.

Not a regression from `a3f88c8` (the old `pos.unwrap_or(0)` behaved identically). It blocks because the diff's whole thesis is contradicted on one of its own changed lines, and because the model that makes it nameable is the model being landed.

**non-blocking — `src/app.rs:134-136`: the `Chosen` doc claim is unqualified and false on one live path.** *"Identity is preserved across a reorder or a refetch."* Only usually. `src/main.rs:273`/`:279` mutate `app.scope` outside `handle_key`, taking no anchor, and `:293` calls `replace_clusters` immediately after — so `cursor_key()` at `app.rs:512` is re-derived from the **already-rescoped** list and pins by identity a row the user never selected. Verified:

```
chose:                        Chosen(1)  map=bbb/two
after out-of-band rescope
        + replace_clusters:   Chosen(0)  map=aaa/one   ← their row is gone
```

This is the one place the parent's suspicion lands: the author's stated justification for `Chosen(usize)` over the ticket's sketched `Chosen(StopKey)` was *"the key is already re-derived from the live clusters at swap time"* — and here re-deriving at swap time derives the wrong key, because the list moved for a reason other than the swap. A `StopKey` would have carried `bbb/two` through. Pre-existing on base, so not a merge blocker, and I do **not** think it justifies reopening the payload decision: the clamp-fallback argument for `usize` is real, `preserve_cursor`'s `vanished → same index, clamped` behaviour genuinely cannot be expressed by a key, and every other list-mutating path is sound (expansion toggles only ever insert/remove strictly *after* the stored index; query edits go to `Untouched`; `cursor_pos` clamps on every read). Narrow the doc claim, or anchor at the `main.rs` call site.

**non-blocking — `src/app.rs:510-513`: `pinned` is `Option<(Option<StopKey>, usize)>`.** A nested option whose middle state means "chosen, but the outgoing screen was empty". Representable, handled, harmless — but it is a small sentinel-shaped wrinkle inside the change that removes one.

**non-blocking — `src/app.rs:736` and `:725`: two stale doc comments.** The comment above `Tab` still promises *"the cursor stays on its ticket if the other screen shows it"*, which is now false for an untouched cursor. `ctrl-g` is the more reachable case, since `wf` opens scoped.

**Positive findings, checked rather than assumed.** No existing assertion was weakened or deleted: diffed against `a3f88c8`, the only test edits are the three mechanical wrappings at `:1238`, `:1242`, `:1301`, and the diff contains no removed `assert`. `refresh::preserve_cursor` is byte-identical and its `None` arm is still live (`cursor_key()` is `None` whenever the outgoing list was empty), so `no_prior_selection_clamps_index` still tests reachable code — the author's "right contract, wrong question" claim holds. `Chosen(usize)` cannot index out of range. No line exceeds 100 columns; no item-after-statement introduced; all four tests are in `#[cfg(test)]` in `src/` with nothing added under `tests/`. The 22-line enum doc is idiomatic for this file (cf. `Startup`, which likewise states the rejected alternative), though "untouched means the top of the list" is now asserted in five places (`:118`, `:129`, `:332`, `:381`, `:501`) — one canonical statement would leave one place to change.

## Spec

Spec is #88 and its comments. Each criterion quotes the ticket.

| # | Criterion | Verdict |
|---|---|---|
| 1 | *"leaves the cursor on the **first** row of the **first** cluster, however many maps stream in and in whatever order"* | met |
| 2 | *"still pinned by identity across a reorder or refetch (the #50/#57 behaviour does not regress)"* | met |
| 3 | *"keeps the user pinned to that row, not silently re-defaulted — no sentinel standing in for 'untouched'"* | met |
| 4 | *"The failing test above passes, and a test covers the user-chosen case reordering underneath them"* | met |
| 5 | *"`cargo fmt --check`, `cargo clippy --all-targets`, `cargo test` green"* | met |

Criterion 1 is met *structurally*, not incidentally: `Untouched` stores no index (`:133`), `cursor_pos` answers it from the live list (`:336`), and `replace_clusters` never writes it (`:510`) — so "in whatever order" cannot fail by construction. Every `Untouched → Chosen` promotion is a nav key that actually moved.

Criterion 3 reaches `Chosen(0)` through a **real gesture** — `Down` then `Up`, resolving through `move_sibling`'s depth-0 sibling branch at `:439` — not a poke at `app.cursor`, and it asserts *both* identity and position after the refetch. Mutation-verified below.

The settled human decision is implemented: *"'the top' means the first stop in the list"* — probing `knotty_app` in forest, an untouched cursor sits on a **done** ticket, the first row drawn rather than the first takeable one. Out-of-scope items are untouched: `scoped_clusters` (#82) and `refresh.rs` (streaming) carry no changes.

**non-blocking — `src/app.rs:535`/`:461`: acting on the top row is not recorded as choosing it.** `Enter` on the top row stages a launch; a map then arriving re-defaults the cursor off the row the human just staged. Same for `Right` opening a group at row 0, and `Up` at the top. Spec: *"Deliberately selecting the top row and then refetching keeps the user pinned to that row, not silently re-defaulted."* Held non-blocking because `Staged` is index-free, so nothing launches wrongly — only the cursor after `esc` — and because "the cursor did not move, so nothing was chosen" is a defensible reading. It is worth naming as a follow-up, since staging *is* the most deliberate gesture in the app.

**non-blocking — criterion 1's "in whatever order" has no test.** Both untouched tests stream the same single order. Verified two further orderings pass and the guarantee is type-level, so this is a coverage gap rather than a defect.

**non-blocking — the query-edit change is in mandate, and confirmed so.** Spec: *"The default cursor position is treated as a user's choice."* Typing already discarded the choice by resetting to 0 pre-diff; the only delta is that a refetch landing on a live query re-derives to the first visible row instead of pinning whichever row happened to be first. Same bug, same fix, applied to the filtered list — not creep.

**non-blocking — `cursor_anchor()` on `tab`/`ctrl-f`/`ctrl-g` is in mandate, but the trade was made silently.** It costs #51's "cursor stays on its ticket across a lens toggle" for the untouched case (`knotty_app`: `#20` before `Tab`, `#1` after; pre-diff it stayed on `#20`). Anchoring there readmits the bug mid-stream, as `toggling_the_lens…` proves, so the trade is right — but no test names the loss and the doc comment still promises the old contract.

## Mutation verification of criterion 3

Run by the reviewer in a detached worktree at the head SHA, not taken on the author's word. Baseline: 163 passed.

| Mutation | Result |
|---|---|
| `replace_clusters`: `Cursor::Untouched \| Cursor::Chosen(0) => None` (the sentinel) | **162 passed, 1 failed** — only `choosing_the_top_row_pins_it_rather_than_re_defaulting_to_the_top` |
| `cursor_anchor()` → `self.cursor_key()` unconditionally | **162 passed, 1 failed** — only `toggling_the_lens_does_not_turn_an_untouched_cursor_into_a_choice` |

Both author claims reproduce exactly: each guard discriminates precisely what it says it does, and criterion 3's test genuinely separates "chose row 0" from "never moved". I also re-ran the ticket's test against `a3f88c8` and got `left: 1, right: 0` — the stated red — while the chosen-row guard is green on base, as a non-regression guard must be.

Two mutations of my own **survived**, which the author did not report:

| Mutation | Result |
|---|---|
| query edits (`:781`, `:787`, `:794`) set `Chosen(0)` instead of `Untouched` | **163 passed** — unguarded |
| empty-list nav (`:424`) sets `Chosen(0)` instead of `Untouched` | **163 passed** — unguarded |

Three of the four `Untouched` writes in the diff are therefore untested. The behaviour is right (see the Spec note above) and this is not blocking on its own, but on a startup path where the user is typing during the ~2.5 s search, `:781`/`:787`/`:794` *is* the real-world instance of the reported bug, and nothing currently holds it red.

## Verdict

1. `src/app.rs:399` — `point_at` records a missed anchor as `Cursor::Chosen(0)`, fabricating a choice from a default and reproducing #88's exact symptom through `Tab`/`ctrl-f`/`ctrl-g`. One-line fix (`pos.map_or(Cursor::Untouched, Cursor::Chosen)`) plus a guard.

Everything else is non-blocking. All five acceptance criteria are met, the two discriminating-test claims survived independent mutation, no existing assertion was weakened, and `Chosen(usize)` is the right payload for the reasons the author gave — the model is sound and the diff is close. It is held only because the one line that contradicts its own thesis is a line this diff wrote.

Recommended alongside the fix, not gating: a guard on the query-edit `Untouched` writes, and narrowing the `Chosen` doc claim at `:134-136` so it does not promise identity preservation that the `main.rs` rescope path does not deliver.

Issue #96, opened 2026-08-07
Three skills, one launch word: does `defer` retire and routing grow a mode axis?


blooop:
## Question

Three wayfinder skills now exist in dotfiles, sharing one `GITHUB_TRACKER.md` and one `LIFECYCLE.md`: `/wayfinder` (interactive, planning-only, 170 lines), `/wayfinder-auto` (charts and drives a map alone — guiding principles in priority order instead of a grilling, execution in scope, 114 lines), and `/wayfinder-one` (single-ticket map for known work: no charting, no fanning, same `/tdd` → gate → fresh-context `/review` lifecycle, 50 lines). They split on **who decides** and **how much fog**.

wf still reaches unattended work the old way: the launch line's `defer` word routes to `/wayfinder`, whose deferred-mode section (settled on #63) is now a weaker duplicate of `/wayfinder-auto` — self-answering grilling, four stop conditions, as-you-go claiming, all of it superseded.

To settle:

- **Does `defer` retire in favour of routing to `/wayfinder-auto`,** and what does the launch grammar become — does the word stay `defer` and change target, or become `auto` to match the skill? (#62 froze the chords; this is the typed second step, not a chord.)
- **Does routing grow a mode axis?** `route` is `(TicketType, Stage) -> Skill` today, and the launch mode is a third, independent axis: a decision node at `ready` routes to `/wayfinder` or `/wayfinder-auto` depending only on what was typed. Per this map's constructive-modeling constraint the function stays exhaustive on every axis, so a new mode or stage is a compile error.
- **Is `/wayfinder-one` reachable from wf at all?** It *creates* its map and ticket rather than launching on an existing row, so there may be nothing to put the cursor on — it may belong at the shell, not in the picker. Related: a one-ticket map is a cluster whose header is not a cursor stop (`Item::Header` returns `None` from `stop_at`), which is why it files two issues rather than one.
- **What happens to `/wayfinder`'s deferred-mode section** — deleted outright, or kept as a pointer at `/wayfinder-auto`? Deleting it is ~20 of the lines that made that skill feel too detailed.

Skill edits land in blooop/dotfiles (as #10, #65 and #66 did); the routing change lands here.

blooop:
**breadcrumb:** the routing half is settled and implemented in [PR #99](https://github.com/blooop/wayfinder/pull/99). The four questions, answered:

[...]

Issue #59, opened 2026-08-06
Map: the general dev-process tree — stage-routed launches, deferred subtrees


blooop:
## Destination

wf is the tree interface to the whole dev process: every node carries a **stage derived from its PR state** (deciding → building → in review → done), enter routes to the skill for that node's (type, stage) — /wayfinder, /tdd, /review — and launching is a **typed two-step**: filter, select, then an optional line that defers judgement to the agent (whole subtree, unattended) or steers it. The skill suite wf routes to exists and is in daily use. wf remains exec-and-exit: it never supervises — a node's lifecycle runs unattended as **one launched agent orchestrating a fresh subagent per stage**, with real gates between.

## Notes

- **Execution belongs in this map** (same override as [Map #1](https://github.com/blooop/wayfinder/issues/1) and [Map #47](https://github.com/blooop/wayfinder/issues/47)): the route runs to a released wf and a working skill suite, not a spec.
- **This map changes the /wayfinder skill itself** (deferred subtree mode relaxes "never resolve more than one ticket per session"). Skill edits land in blooop/dotfiles, as [teach-the-skill #10](https://github.com/blooop/wayfinder/issues/10) did; tickets stay here.
- **Builds on [Map #47](https://github.com/blooop/wayfinder/issues/47)**: the PR data plane ([#49](https://github.com/blooop/wayfinder/issues/49)'s `closedByPullRequestsReferences`), PR badges ([#52](https://github.com/blooop/wayfinder/issues/52)) and the depth-aware cursor ([#57](https://github.com/blooop/wayfinder/issues/57)) are taken as given, not re-decided.
- **`/constructive-modeling` on all work here** (constraint carried from Map #1): sum types, no sentinels, exhaustive match, parse-don't-validate at the `gh` boundary — the stage lattice especially.
- Skills for sessions here: `/grill-me`, `/constructive-modeling`, `/prototype`, `/research`.

Decisions locked while charting (2026-08-06):

- **wf stays a selector** — [#26](https://github.com/blooop/wayfinder/issues/26) stands. It draws the tree, you pick, it execs one `claude` and exits. Everything new lives in what the tree shows and what that one process is told to do.
- **Deferring = one launch resolves a whole subtree unattended.** The agent decides instead of grilling, under a standing default prompt plus an optional per-launch steering line. Chosen **per launch in wf**, not per ticket label — a judgement made where you're standing, not weeks earlier.
[...]

PR #109, opened 2026-08-08
The search prompt leads the screen, and enter opens a launch picker

blooop:
Two screen changes, driven by using it: the thing you type into was at the far
end of the screen from the title, and the launch it stages was a word you had to
already know.

## The search prompt leads the screen

The `>` prompt and its count line were the last two lines before the key hints,
so typing happened at the bottom while the rows re-sifted upward, away from the
cursor. Both move above the body — prompt on the first line inside the border,
count under it, tree beneath, hints still anchored to the bottom. `fzf
--reverse`'s order, and the count line stays against the query whose matches it
counts. The body's own leading blank became the gap between the chrome and the
first cluster, so nothing needed inserting.

## `enter` opens a launch picker

#62 staged the launch on a one-line prompt and #96 gave that line a mode axis,
which made the modes *words*: first `defer`, then `auto`, typed blind at the
front of the line. Three problems. The words were undiscoverable — nothing on
screen said `auto` existed, or that it meant nobody would be asked anything.
Typing one looked exactly like a typo until an unattended agent started. And the
grammar had to defend itself against ordinary English: `automate the release`
was special-cased into *not* meaning unattended.

So `enter` opens a centred overlay instead:

```
┌ launch #42 Make cargo test pass inside the container ────────────────────┐
│                                                                          │
│  ▶ interactive /wf       you are in the loop; it grills you              │
│    auto        /wf-auto  the agent decides alone and drives it to done    │
│                                                                          │
│    steer  █                                                              │
│                                                                          │
│  enter launch · ↑/↓ mode · type to steer · esc cancel                    │
└──────────────────────────────────────────────────────────────────────────┘
```

Every mode is a row naming the skill it routes *this* node to and who ends up
deciding — the difference between `/wf` and `/wf-auto` is the choice being made,
so both are on screen rather than one being echoed back after the fact. Arrows
or tab move; `enter` runs the row you are on, so the common case is still `enter
enter`. Anything typed lands in an always-visible steer field and steers
whichever mode is selected.

**No string can move the selection**, so a launch is unattended only because you
selected it — and steering text starting `auto` needs no special case, because
it was never a mode.

### What that changes in the code

- `Overlay::LaunchLine { staged, text }` → `PickLaunch { staged, mode, steer }`:
  the mode is state, not a re-parse of the text on every frame.
- `LaunchMode::parse` → `LaunchMode::picked(mode, steer)`. Still the same
  product of two axes; the mode half just arrives already decided.
- The picker lists `Mode::all()`, derived from an exhaustive `Mode::after`
  cycle, so a third mode joins the overlay without a second place to edit.
- The count line keeps its slot — a modal covers the rows, and does not need the
  chrome under it to move out of the way.

### The ruling this revisits

#96 settled two halves: `auto` → `/wf-auto` (routing), and `auto` as the word on
the launch line (input). **The routing half is untouched** — `route` is still
`(&Aim, Mode) -> Route`, total on every axis. Only the input half changes: the
mode is picked, not typed. #62's chords are otherwise intact; `enter` still
stages and a second `enter` still launches.

## Testing

208 unit + 18 integration tests green, including `live_launch_exec`, which
drives a real PTY and actually execs an agent — updated to wait for the picker
instead of the old `→ /wf` line. Driven for real via `wf-next`.

## Summary by Sourcery

Rework the launch flow to use a modal launch picker and move the search prompt to the top of the screen.

New Features:
- Introduce a launch picker overlay that lets users choose the launch mode and steering text before running an agent.

Enhancements:
- Anchor the search prompt and match-count line at the top of the screen so typing happens at the top while results sift below.
- Make launch modes explicit `Mode` values with labels and blurbs, and derive the picker options from an exhaustive mode cycle.
- Ensure launch mode selection is driven by cursor movement rather than parsing typed text, so steering input can no longer implicitly change who decides.

Tests:
- Extend unit, integration, and live PTY tests to cover the new launch picker behaviour and the repositioned search chrome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ and the rows themselves are right there. Both are ordinary cursor stops: put
 the cursor on one and `→` opens it in place (`▾`), listing what it held as rows
 you can select and launch, and `←` shuts it again — so nothing is ever merely a
 number you cannot reach. A map with nothing takeable
-leaves the body entirely, counted on the bottom line as `· N idle maps hidden`.
+leaves the body entirely, counted on the count line as `· N idle maps hidden`.
 
 **A row that heads a branch carries the same rollup of what is beneath it**,
 written `(beneath) ⊘2` at the end of the line — so a takeable ticket says the
@@ -226,7 +226,10 @@ blocker; edges the tree cannot show are annotated `⤷ also needs #n`.
 
 **Typing sifts**: a live query prunes whichever screen `tab` had toggled down to
 the rows that matched, keeping the tree they sit in; clearing it restores that
-screen whole. Matching is fuzzy but **tight** — every character the query lands
+screen whole. The `>` prompt sits at the **top**, directly under the title, with
+the count line under it and the rows beneath both — you type at the top of the
+screen and watch the tree sift below it, as `fzf --reverse` does. Only the key
+hints stay anchored to the bottom. Matching is fuzzy but **tight** — every character the query lands
 on has to start a word or sit against another matched one, so `map` finds "the
 **m**anager-**a**gent **p**rotocol" and sub`tree` matches mid-word, while
 letters picked out of the middles of three unrelated words do not: `tree`
--- a/README.md
+++ b/README.md
@@ -292,25 +292,42 @@ agent inside it — see
 [Isolation](#isolation-the-agent-runs-in-the-repos-devcontainer).
 
 **Launching is two steps.** `enter` on the cursor's node does not launch: it
-opens the **launch line** where the count line was, showing where `enter` will
-go — `→ /wf-tdd · #65 Author the /wf-tdd skill`. What you type lands on that line, not
-in the query, and *is* the mode:
+opens the **launch picker** over the list, showing what is about to happen and
+the one thing still undecided — who resolves the node:
 
-| the line says | what launches |
-| --- | --- |
-| *(empty)* | interactive — the default |
-| `auto` | the agent decides alone and drives the rest of the lifecycle unattended |
-| `auto <text>` | the same, with `<text>` as the steering prompt |
-| *anything else* | interactive, with what you typed as the steering prompt |
+```
+┌ launch #65 Author the /wf-tdd skill ────────────────────────────┐
+│                                                                 │
+│  ▶ interactive /wf-tdd   you are in the loop; it grills you     │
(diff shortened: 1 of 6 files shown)


blooop:
## Self-review

Read back as a reviewer rather than as the author. Three findings, all now fixed
in the follow-up commit; one design note that stands as-is; two things
deliberately not done.

### Fixed

**A keybinding was hiding in a helper.** `stepped(mode, key.code)` took a
`KeyCode` and decided direction from it (`Up` → back, everything else →
forward), so half of a binding lived somewhere the rest of the bindings cannot
be read from — and the helper accepted keys that never mean a step. It takes a
distance now, with `next_mode`/`previous_mode` naming what the handler wants.

**`draw_overlay` read as if it *were* the checkout modal.** It destructured one
variant into pre-declared bindings, returned early for the other, then fell
through into the checkout modal's body — which still carried that modal's doc
comment, now describing a function that dispatches two of them. Split:
`draw_overlay` is a three-arm exhaustive match, `draw_checkout_picker` owns its
body and its comment.

**The wrap was untested.** `previous_mode` had no test, so nothing pinned that
`↑` from the first row lands on the last rather than sticking, that `tab` steps
forward in the picker rather than toggling the view behind it, or that moving
the mode leaves typed steering text alone. That last one is the real risk of a
two-axis overlay — one axis quietly resetting the other. Pinned now.

### Checked and correct

`enter` returns before the overlay is put back, which matters: `resolve_launch`
may have opened the which-checkout modal, and falling through would overwrite it
with the launch picker the launch had just left. Covered by
`the_picked_mode_survives_the_checkout_picker`, and the early return now says
why it is early.

### Design note

`tab` moves the mode inside the picker while `tab` toggles leverage ⇄ forest
outside it. That is not an accident and not a conflict — the modal owns every
key while it is up, as the checkout picker already does with `j`/`k` — but it is
a second meaning for the same key and worth knowing about.

### Not done

- **No ticket.** #96 froze `enter`'s chords and settled the typed-word input in
  the same breath; this changes the input half without a decision node saying
  so. The routing half is untouched. Filing it against map #59 after the fact is
  the honest sequence if this wants a paper trail.
- **`Mode::all()` allocates** on each keypress and once per picker frame. Two
  elements, drawn on a human keystroke — measuring it would cost more than it
  saves.

PR #113, opened 2026-08-08
A plain launch mode: enter a node with a bare claude session

### LINKED ISSUE #112: Build: a plain claude session as a third launch mode
## What to build

The launch picker offers two modes today — `interactive` (`/wf`, `/wf-tdd`, `/wf-review`) and `auto` (`/wf-auto`) — and both hand the agent a skill. There is no way to reach a node with **no skill at all**: a plain `claude` session, in the same workspace and container the skill launches would have got, that you drive yourself. Today that means quitting `wf`, remembering the workspace spec and typing `dl` by hand.

Add `plain` as a third launch mode. It routes every aim — a ticket or a whole map — to a session that execs `claude --dangerously-skip-permissions` with **no slash command**. Everything else about the launch is unchanged: same checkout resolution, same per-node `owner/repo@wayfinder/<repo>-<n>` workspace, same devlaunch detection. Only the prompt differs.

Steering text typed in the picker becomes that session's opening prompt **verbatim** — there is no skill for a ` steer: …` suffix to steer, so the suffix form would be talking to nobody. With nothing typed there is no prompt argument at all, so `claude` opens on its own prompt rather than on an empty string.

## Seams (pre-agreed — test at these, no others)

1. **`route` (`src/launch.rs`)** — `(&Aim, Mode) -> Route`, still total on every axis: `plain` routes both aims, every type and every stage, to the new plain route.
2. **`Launch::agent_argv` (`src/launch.rs`)** — the exec'd argv: no prompt argument when nothing was typed, typed text verbatim as the single prompt argument, and the isolated `dl` form carrying the same command through its shell-quoting unchanged.
3. **The mode list and the picker's cycle (`Mode::all` in `src/launch.rs`, mode stepping in `src/app.rs`)** — the third mode exists in picker order and `↑`/`↓` reach it and wrap through it.

## Acceptance criteria

- [ ] The picker lists a third mode, naming what it execs and what picking it means
- [ ] `↑`/`↓` in the picker reach the new mode and still wrap
- [ ] `plain` on a ticket and `plain` on a map both exec `claude` with no slash command
- [ ] With no steering text, no prompt argument is passed at all
- [ ] Typed steering text is the prompt verbatim — no `steer:` prefix, no skill name
- [ ] The devlaunch form is unchanged apart from the prompt: same workspace, same quoting
- [ ] `route` stays exhaustive — a new mode, type or stage is still a compile error
- [ ] The README's launch section documents the mode
- [ ] `cargo fmt --check`, `cargo clippy --all-targets --all-features`, `cargo test` green

## Out of scope

- Retiring or renaming `interactive`/`auto`
- A plain session outside the per-node workspace (host-vs-container detection is untouched — a plain session lands wherever a skill launch would have)
- Any change to what the skills themselves do

blooop:
Closes #112.

A third row in the launch picker: **plain** — a session with no skill in it.

```
  ▶ interactive /wf-tdd   you are in the loop; it grills you
    auto        /wf-auto  the agent decides alone and drives it to done
    plain       claude    no skill; a bare session on the node's branch
```

`wf` already resolves a node to a place to work — a checkout, a per-node branch,
a clone and a container. Until now that work was only reachable with a skill
bolted to the front of it, so reading the code a ticket is about, or a fix too
small to be a lifecycle, meant quitting `wf` and retyping the workspace spec by
hand. `plain` is the same launch with nothing invoked in it.

## The shape

- **`Mode::Plain`** joins the picker's cycle, and `route` gains one arm:
  `(Aim::Map | Aim::Ticket { .. }, Mode::Plain) => Route::Plain`. It collapses
  the table for the opposite reason `Auto` does — `Auto` picks one skill for
  every node, `Plain` picks none — and the match stays exhaustive on all three
  axes, so a new mode, type or stage is still a compile error.
- **The prompt became an `Option`.** `claude_argv` no longer always has a third
  entry: with no skill to invoke and nothing typed there is no prompt argument
  at all, rather than an empty string (`claude ""` is a prompt, and an empty
  one). `skill_invocation` answers `None` for `Route::Plain` and the argument
  shapes stay in one exhaustive `(route, aim)` match with no unreachable arm.
- **Steering text is handled differently here, on purpose.** With a skill in
  front of it, typed text rides as ` steer: <text>`. With no skill, that suffix
  would be addressed to nobody — so the text is simply the session's opening
  prompt, verbatim.
- **Nothing else about a launch changes**: same checkout resolution, same
  `owner/repo@wayfinder/<repo>-<n>` workspace, same devcontainer detection, same
  shell-quoting through `dl`.

## Tests

Written test-first at the three seams #112 pre-agreed:

- routing — every type, every stage and both aims under `plain`, plus the exec'd
  argv being `claude` and nothing else
- the exec'd command — typed text as the whole prompt, and no prompt argument
  when nothing was typed, for a ticket aim and a map aim
- the mode list and the picker — `Mode::all` carries it, `↑`/`↓` reach it and
  wrap through it in both directions, and the rendered picker shows the row with
  what it execs

The two tests that pass over paths already covered (the container form, the
picker rendering) were **mutation-verified** rather than trusted: breaking the
prompt to `Some("")` and dropping the mode from the cycle each make them fail.

`cargo fmt --check`, `cargo clippy --all-targets --all-features` under
`-D warnings`, and 239 tests: green.

## Summary by Sourcery

Add a new plain launch mode that opens a bare claude session in the node’s workspace without invoking any skill, while keeping all existing workspace resolution behavior unchanged.

New Features:
- Introduce Mode::Plain and Route::Plain to support launching a skill-free claude session on a node’s workspace from the launch picker.

Enhancements:
- Refactor launch argument construction so claude’s prompt becomes optional and is derived from either the chosen skill route or directly from user steering text in plain mode.
- Extend the launch picker and mode cycling to include the new plain mode, ensuring consistent navigation and display across UI and app layers.

Documentation:
- Update README to document the plain launch mode, its behavior, and how it appears and is used in the launch picker.

Tests:
- Add routing, argv-shape, isolation, and UI tests to cover plain mode behavior, its inclusion in mode lists, and navigation in the launch picker.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,10 +71,13 @@ Three consequences for anyone editing here:
   it. `skills/wf-one/SKILL.md` and `wf-auto` link
   `../wf/GITHUB_TRACKER.md` and `../wf/LIFECYCLE.md` by relative
   path, so the five move as a set and the layout under `skills/` is load-bearing.
-- **Adding a `Route` means adding a skill.** `skills::BUNDLED` lists what ships
-  and a unit test asserts every `Route::label()` is in it, so a route pointing
+- **Adding a `Route` means saying which skill it invokes — or that it invokes
+  none.** `skills::BUNDLED` lists what ships and `Route::bundled_skill()` is the
+  exhaustive answer to which of them a route names; a unit test sweeps
+  `Route::all()` and asserts every `Some` is in the bundle, so a route pointing
   at a prompt the package does not carry fails the build rather than an agent
-  launch.
+  launch. `Route::Plain` (#112) is the one `None`: it execs the agent with no
+  slash command, so there is no prompt for the package to be missing.
 - **Renaming one leaves residue on every machine that had the old name.**
   `status` and `install` both iterate `BUNDLED`, so a link named for a skill
   that left the list is invisible to them while staying on disk, dangling.
--- a/README.md
+++ b/README.md
@@ -315,15 +315,16 @@ opens the **launch picker** over the list, showing what is about to happen and
 the one thing still undecided — who resolves the node:
 
 ```
-┌ launch #65 Author the /wf-tdd skill ────────────────────────────┐
-│                                                                 │
-│  ▶ interactive /wf-tdd   you are in the loop; it grills you     │
-│    auto        /wf-auto  the agent decides alone and drives it  │
-│                                                                 │
-│    steer  █                                                     │
-│                                                                 │
-│  enter launch · ↑/↓ mode · type to steer · esc cancel           │
-└─────────────────────────────────────────────────────────────────┘
+┌ launch #65 Author the /wf-tdd skill ──────────────────────────────────┐
+│                                                                       │
+│  ▶ interactive /wf-tdd   you are in the loop; it grills you           │
+│    auto        /wf-auto  the agent decides alone and drives it to done│
+│    plain       claude    no skill; a bare session on the node's branch│
+│                                                                       │
+│    steer  █                                                           │
+│                                                                       │
+│  enter launch · ↑/↓ mode · type to steer · esc cancel                 │
+└───────────────────────────────────────────────────────────────────────┘
 ```
 
 `↑`/`↓` (or `tab`) move between the modes and `enter` runs the one you are on,
@@ -361,6 +362,7 @@ headers are one `↑` away.
 | `wayfinder:build` | in review | interactive | `claude "/wf-review <n>"` |
 | research · prototype · grilling · task | any unfinished stage | interactive | `claude "/wf <map> <n>"` |
 | anything | any unfinished stage | auto | `claude "/wf-auto <map> [<n>]"` |
+| anything | any unfinished stage | plain | `claude` — no skill, no arguments |
 | a ticket | done | — | nothing — not launchable |
 
 The auto mode collapses the ticket rows on purpose: the launched session is a
@@ -369,6 +371,17 @@ the gate, then a fresh-context `/wf-review` — so it is the manager skill that 
 not the one skill that stage would have called. Steering text rides whichever
 route you got, as ` steer: <text>` on the end of the prompt; the mode itself
 never does, because it has already chosen the skill.
+
+**The plain mode collapses them for the opposite reason: it runs no skill at
+all.** Everything else about the launch is unchanged — same checkout, same
+per-node branch, same clone and container — so it is the way to get `wf`'s
+workspace without a skill's opinion about what to do in it: reading the code a
+ticket is about, a quick fix too small to be a node's lifecycle, or picking up
+after an agent that stopped somewhere awkward. There being no skill in front of
+it is also why steering text is handled differently here: a ` steer: <text>`
+suffix would be addressed to nobody, so whatever you type is simply the
+session's opening prompt, and typing nothing passes no prompt at all rather than
+an empty one.
 
 | key | what it does |
 | --- | --- |
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ headers are one `↑` away.
 | `wayfinder:build` | in review | interactive | `claude "/wf-review <n>"` |
 | research · prototype · grilling · task | any unfinished stage | interactive | `claude "/wf <map> <n>"` |
 | anything | any unfinished stage | auto | `claude "/wf-auto <map> [<n>]"` |
-| anything | any unfinished stage | plain | `claude` — no skill, no arguments |
+| anything | any unfinished stage | plain | `claude` — no skill; anything typed is the whole prompt |
 | a ticket | done | — | nothing — not launchable |
 
 The auto mode collapses the ticket rows on purpose: the launched session is a
--- a/src/app.rs
+++ b/src/app.rs
@@ -72,10 +72,9 @@ pub enum Overlay {
 
 /// The mode after `mode` in the launch picker, wrapping.
 ///
-/// Derived from [`Mode::all`] rather than written as a toggle: with two modes a
-/// toggle is the same thing, but it stops being the same thing the moment a
-/// third mode exists, and this is the code that would silently keep working
-/// while skipping it.
+/// Derived from [`Mode::all`] rather than written as a toggle. With two modes a
+/// toggle was the same thing; #112's third mode is exactly the change that
+/// would have left a toggle silently working and skipping the new one.
 fn next_mode(mode: Mode) -> Mode {
     stepped(mode, 1)
 }
@@ -720,9 +719,10 @@ impl App {
             KeyCode::Enter => {
                 return self.resolve_launch(&staged, &LaunchMode::picked(mode, &steer))
             }
-            // With two modes every step is the same step; `tab` joins the
-            // arrows because it is what the rest of the screen uses to move
-            // between two of something.
+            // The two directions are genuinely different steps now that there
+            // is a third mode (#112); `tab` joins the arrows because it is what
+            // the rest of the screen uses to move through a list, and it steps
+            // the way `down` does rather than toggling.
             KeyCode::Up => mode = previous_mode(mode),
             KeyCode::Down | KeyCode::Tab => mode = next_mode(mode),
             KeyCode::Backspace => {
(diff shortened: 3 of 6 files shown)


blooop (commented):

Two axes, run as independent subagents against `92655087...HEAD` with checks green, and not merged or reranked against each other.

## Standards

**Blocking**

**1. `src/app.rs` — the production comment now says the opposite of the code.** `// With two modes every step is the same step; tab joins the arrows because it is what the rest of the screen uses to move between two of something.` sits directly on the `Up`/`Down`/`Tab` arms. With three modes neither clause is true, and `tab` no longer round-trips in one press. The PR updated the mirroring *test* comment to say exactly this ("the two directions are no longer the same step") and left the real one stale. Same class, lower stakes: `next_mode`'s doc and `draw_launch_picker`'s ("so both are on screen").

**2. `AGENTS.md` documents an invariant this PR breaks, unamended.** "**Adding a `Route` means adding a skill.** `skills::BUNDLED` lists what ships and a unit test asserts every `Route::label()` is in it, so a route pointing at a prompt the package does not carry fails the build." `Route::Plain` is a route whose label ships no skill. The test (`every_route_wf_can_exec_is_in_the_bundle`) iterates a hand-written four-element array, so it never noticed — and a *future* route that genuinely is a skill won't be noticed either. That is precisely the second-place-to-remember `Mode::after`/`Mode::all` was built to avoid. Fix by making the exception typed: `Route::bundled_skill(self) -> Option<&'static str>`, `None` for `Plain`, iterated exhaustively — then AGENTS.md's claim becomes true again.

**3. `Route::label()`'s doc is now false.** "Every label is prefixed `wf`, because these names are claimed in a namespace `wf` does not own." `Plain`'s label is `claude` — unprefixed, and not a name in `~/.claude/skills` at all. One `&'static str` now carries two different kinds of thing: a slash command to exec, and the agent binary's own name, which `claude_argv` separately hardcodes.

**Non-blocking**

**4. `claude_argv`'s second paragraph describes a body that moved.** "The map aim is matched first because … `route` hands a cluster header nothing but a wayfinder skill." In `skill_invocation` the map arm is *second*, after `(Route::Plain, _)`, and `route(&Aim::Map, Mode::Plain) == Route::Plain`, so the stated premise is false. It belongs on `skill_invocation`, rewritten.

**5. `Launch::prompt` reaches into `self.mode.steer`.** `LaunchMode` documents itself as the owner of the steering axis and exposes `suffix()` for exactly this question; now two types decide what typed text becomes. A sibling `LaunchMode::opening_prompt()` keeps it in one place.

**6. `Launch.route` is derived state that can now disagree meaningfully.** It is recomputable from `aim` + `mode`, and since `prompt()` branches on `route` while the steer lives on `mode`, a mismatched pair silently changes what the typed text *means* — not just a printed label as before. `isolated(route, mode)` sets them independently. Cf. AGENTS.md house style: "where a sum type already says something, do not add a bool that can disagree with it."

**7. Duplicated test helpers.** `ticket_argv`/`ticket_prompt` and `map_argv`/`map_prompt` differ only by `.last()`; `map_prompt`'s doc ("The same, for a launch aimed at the whole map") now dangles after the inserted `map_argv`, and `ticket_argv` was parked 158 lines from its twin.

No unreachable arms: `skill_invocation` is exhaustive with `Plain` absorbed first, and no `Route`/`Mode` switch elsewhere needed a new arm. Tests are not tautological — `plain_launches_a_session_with_no_skill_in_it` mirrors the existing `auto_…` sweep and the argv assertions go through `plan`.

## Spec

Verified every acceptance criterion against the code and tests, and mutation-tested the seams rather than trusting the PR body.

**All nine acceptance criteria are met.** `route` gains `(Aim::Map | Aim::Ticket { .. }, Mode::Plain) => Route::Plain` with no wildcard on any axis; `Mode::all` is derived from the private `after` cycle so the picker picks the row up automatically; `prompt()` returns `None` when nothing was typed and the trimmed steer verbatim otherwise; the devlaunch test pins the unchanged `blooop/wayfinder@wayfinder/wayfinder-80` spec and the single-quoted command.

Mutation checks run independently: replacing the plain prompt with the ` steer: …` suffix form fails three tests, so "*Typed steering text is the prompt verbatim*" and "*With no steering text, no prompt argument is passed at all*" are genuinely pinned, at seam 2 (`Launch::agent_argv`) as the ticket named it. Seam 1 is walked over all six types × four launchable stages plus the map aim. Seam 3 is covered in both `src/app.rs` (up and down each walked the whole way round) and `src/ui.rs`.

**Non-blocking — one half of an acceptance criterion is unpinned.** "*The picker lists a third mode, naming what it execs and what picking it means*" has two halves; only the second is tested. The UI test asserts `screen.contains("plain")` and `screen.contains("no skill")` — the label and the blurb — but never asserts the route column reads `claude`. Mutating `Route::Plain => "claude"` to `"ZZZZ"` leaves the full suite green: `Route::Plain::label()` has no other caller, so nothing anywhere pins what the plain row says it execs. One `assert!(screen.contains("claude"))` beside the existing two closes it.

**Non-blocking — README table row slightly overstates.** "`claude` — no skill, no arguments" reads as unconditional, while the ticket says typed text "*becomes that session's opening prompt verbatim*". The prose immediately below does say this, so the criterion "*The README's launch section documents the mode*" is met.

**Nothing out of scope was done.** `interactive`/`auto` are untouched, no skill files are in the diff, and isolation/host-vs-container detection is unmodified.

Verified only by reading: "*`route` stays exhaustive*" is a compile-time property with no test; no wildcard arm exists in `route` or its inner matches. Note `skill_invocation` does use `(_, Aim::Map)`, so a hypothetical new `Route` would compile-error only on its ticket arm — the criterion as written concerns `route`, and holds.

## Verdict

**Request changes** — three blocking findings, all on the Standards axis, none of them in the behavior:

1. Stale production comments asserting a two-mode picker (`src/app.rs`, and the same claim in `next_mode` and `draw_launch_picker`).
2. `AGENTS.md`'s "adding a `Route` means adding a skill" invariant is now false and its test cannot catch the next violation — needs `Route::bundled_skill` and an exhaustive iteration.
3. `Route::label()`'s "every label is prefixed `wf`" doc is false for `Plain`.

The Spec axis is clean and would approve on its own; its one gap (nothing pins the `claude` label the picker draws) is worth closing in the same pass.

blooop:

**Blocking findings addressed** in `e960ea7`; CI green.

| finding | what changed |
| --- | --- |
| 1 — stale two-mode comments | The key handler, `next_mode` and `draw_launch_picker` all say what is now true: three modes, two genuinely different directions, `tab` stepping rather than toggling. |
| 2 — `AGENTS.md` invariant false, test blind | `Route::bundled_skill() -> Option<&'static str>` is the typed exception (`None` only for `Plain`), and the test sweeps `Route::all()` — derived from a private `after` cycle, the same device `Mode::all` already used for the same reason. `AGENTS.md` now states the invariant including its exception. |
| 3 — `label()`’s "prefixed `wf`" doc | Rewritten: the namespace argument is about *skill* labels, and `Plain` is outside it because it claims no name. |

The reviewer’s suggested fix for 2 was taken as given, and **mutation-verified rather than assumed**: dropping a route from the cycle now fails `every_route_wf_can_exec_is_in_the_bundle` (`left: 4, right: 5`), where before the hand-written array would have sailed past it.

Non-blocking, also taken: **5** — `LaunchMode::opening_prompt` owns both halves of the steering axis, since the reviewer is right that typed text means a different thing with no skill in front of it and two types should not both decide that; **4** — `claude_argv`’s doc no longer describes a body that moved into `skill_invocation`; **7** — the helper twins collapsed, `ticket_prompt`/`map_prompt` now reduce their `_argv` siblings.

Spec’s gap closed: `Route::Plain::label()` had no caller pinning it, so mutating `"claude"` to `"ZZZZ"` left the suite green. The picker test now asserts the route column, and that mutation fails it. README’s table row no longer reads as though typed text were dropped.

**Not taken — 6 (`Launch.route` as derived state).** Fair, and it predates this PR: `route` has been stored on `Launch` since the two-step launch, and `plan` — the only constructor outside tests — always derives it from the same `(aim, mode)` pair it stores. What this PR changes is the *stakes* of a mismatch, which is a real observation, but recomputing it (or making the pair unrepresentable apart) is a refactor of the launch struct rather than this ticket’s slice. Worth its own node.

PR #115, opened 2026-08-08
Delete the deferred-mode residue and pin the glyph display order

### LINKED ISSUE #86: The glyph display order is load-bearing and unpinned
## Question

The [Build 13 review](https://github.com/blooop/wayfinder/pull/85#pullrequestreview-4880754142) found that **`RowGlyph`'s declaration order is load-bearing and unguarded**.

`RowGlyph`'s derived `Ord` *is* the glyph display order, and three render sites now depend on it: the cluster header counts, the shut-group rollups, and the branch-root rollups. Swapping `Stage::Building` and `Stage::InReview` in the declaration leaves **158 of 159 tests green** — the one failure is an unrelated PR-semantics test that happens to notice. So the ordering that decides how every tally reads on screen can be changed by accident, in a type whose `Ord` is derived precisely so it can be reordered by editing the declaration.

This is the cost side of the [#78](https://github.com/blooop/wayfinder/issues/78) decision to make one `Ord` serve as both the stage lattice's severity order and the glyph display order. That decision was right — it is what let `Map::counts` and `Status::group` be deleted rather than wrapped — but it means one declaration now answers two questions, and only one of them is tested.

What pins it? Options worth weighing: a single test asserting the full six-glyph sequence; separating display order from severity order (paying back the collapse #78 bought); or accepting it with the coupling documented on the type.

Non-blocking, so this is not a defect to fix under time pressure — it is a question about how much a derived `Ord` should be allowed to carry.

### LINKED ISSUE #96: Three skills, one launch word: does `defer` retire and routing grow a mode axis?

blooop:
Closes #96. Closes #86.

Two small residues from this map, landed together.

## #96 — the deferred-mode residue

PR #99 retired the `defer` launch word in favour of `auto` → `/wf-auto`, leaving `/wf`'s deferred-mode section unreachable from `wf` but still in the skill. #96's breadcrumb ruled it **deleted outright**; the edit lands here rather than in dotfiles because the skill suite moved into this repo in PR #101.

- `skills/wf/SKILL.md`: the **Deferred mode** section is gone; the invocation grammar loses its two `defer` lines; the one-ticket-per-session rule loses its deferred exception; the HITL rule's exception is replaced by a pointer at `/wf-auto`, which is where handing judgement to the agent lives now.
- `skills/wf/LIFECYCLE.md`: the opener names `/wf-auto` as its referrer instead of "the `wf` skill's deferred mode".
- `skills/wf-auto/SKILL.md`: gains the one norm the deleted section carried that it lacked — *a human later re-opening an agent-decided ticket to re-decide it is normal, not a conflict*. Everything else in the deleted section (standing default prompt, as-you-go claiming, stop conditions, steering composition) was already superseded by `/wf-auto`'s own text.
- **Check**: `grep -rni defer skills/` now returns nothing — no skill offers `defer` as a word a user can type.
- Left alone as accurate history: `src/skills.rs:7`, the `Mode` doc comments in `src/launch.rs`, and the two README passages describing the era `defer` belonged to.

## #86 — pinning the glyph display order

`RowGlyph`'s derived `Ord` **is** the glyph display order, trusted by three render sites (cluster header counts, shut-group rollups, branch-root rollups), and no test failed when `Stage::Building` and `Stage::InReview` swapped places in the declaration.

Of the ticket's three candidates, this takes **the single test asserting the full six-glyph sequence** (`the_full_glyph_vocabulary_sorts_into_display_order` in `src/model.rs`): separating display order from severity order would re-add the duplication #78 deliberately collapsed — and #86 itself records that collapse as the right call — while documentation alone cannot go red on the swap. The test sorts all six `RowGlyph` values and asserts the rendered sequence `○◐◍!●⊘`, so any reorder of the declaration fails it.

**Mutation evidence** — with `Building` and `InReview` swapped in the declaration, the pin goes red:

```
---- model::tests::the_full_glyph_vocabulary_sorts_into_display_order stdout ----
thread 'model::tests::the_full_glyph_vocabulary_sorts_into_display_order' (3379) panicked at src/model.rs:1093:9:
assertion `left == right` failed
  left: "○◍◐!●⊘"
 right: "○◐◍!●⊘"
```

Swapped back, the suite is fully green: 227 lib + 10 bin unit tests, all live tests, `cargo clippy --all-targets -- -D warnings` clean, rustfmt clean.

## Summary by Sourcery

Remove the deprecated deferred mode from the wf skill and pin the glyph display order with a dedicated test.

Bug Fixes:
- Add a test to assert RowGlyphs always sort into the intended display order, preventing accidental changes when Stage variants are reordered.

Enhancements:
- Clean up wf skill documentation by removing the unreachable deferred mode and clarifying that handing judgement to the agent is owned by /wf-auto.
- Update lifecycle documentation to reference /wf-auto instead of wf deferred mode and to describe subtree handling more generally.
- Carry forward the norm that re-opening agent-decided tickets is normal into the /wf-auto skill documentation.

Tests:
- Introduce a glyph ordering test that sorts the full RowGlyph vocabulary and asserts the expected display sequence.

### DIFF
--- a/skills/wf-auto/SKILL.md
+++ b/skills/wf-auto/SKILL.md
@@ -87,7 +87,7 @@ Per ticket:
 1. Load the map (low-res; zoom into closed tickets on demand). Claim the ticket. If it was already claimed, read its trail first — the last `### handoff`, then the breadcrumbs after it — and open with a breadcrumb noting resumption.
 2. Hand it to a fresh subagent with the ticket body, the map's Decisions-so-far, the principles, and the skill its type calls for. Build tickets run the [LIFECYCLE.md](../wf/LIFECYCLE.md) stages with gates between; everything else resolves in one subagent.
 3. Breadcrumb each decision-grade moment on the ticket — a sub-decision settling, a direction changing, a stage transition, a gate result.
-4. **Record**: a resolution comment opening `**agent-decided (<principle>):**` and carrying the reasoning that would have been the grilling; close the ticket; append its one-line gist to Decisions-so-far with an *(agent)* suffix.
+4. **Record**: a resolution comment opening `**agent-decided (<principle>):**` and carrying the reasoning that would have been the grilling; close the ticket; append its one-line gist to Decisions-so-far with an *(agent)* suffix. A human later re-opening an agent-decided ticket to re-decide it is normal, not a conflict.
 5. **Advance the map**: graduate whatever fog the answer made specifiable into fresh tickets — including slicing new `wayfinder:build` tickets — clear those fog patches, and retire tickets the answer invalidated.
 
 Then take the next unblocked ticket. The run ends when the frontier is empty, everything left is parked, or the budget runs out — and it ends with a summary: what closed, what parked, what is still open.
--- a/skills/wf/LIFECYCLE.md
+++ b/skills/wf/LIFECYCLE.md
@@ -1,6 +1,6 @@
 # The manager protocol — a node's lifecycle, one fresh subagent per stage
 
-How a launched session drives a node (or a deferred subtree of nodes) through its stages without anyone watching. This is referenced by the `wf` skill's deferred mode and honored by `/wf-tdd` and `/wf-review`; nothing launches "the manager" directly — **the launched session is the manager**. wf itself never is: it exec'd and exited.
+How a launched session drives a node (or a whole subtree of nodes) through its stages without anyone watching. This is referenced by `/wf-auto` and honored by `/wf-tdd` and `/wf-review`; nothing launches "the manager" directly — **the launched session is the manager**. wf itself never is: it exec'd and exited.
 
 ## The stage lattice (decided on blooop/wayfinder#61)
 
--- a/skills/wf/SKILL.md
+++ b/skills/wf/SKILL.md
@@ -80,7 +80,7 @@ The answer isn't part of the body — it's recorded on resolution (see [Work thr
 
 ## Ticket Types
 
-Every ticket is either **HITL** — human in the loop, worked *with* a human who speaks for themselves — or **AFK**, driven by the agent alone. A HITL ticket only resolves through that live exchange; the agent never stands in for the human's side of it (a grilling agent that answers its own questions has broken this). The single exception is a **deferred launch** (see [Deferred mode](#deferred-mode)), where the human has explicitly handed judgement to the agent for a subtree.
+Every ticket is either **HITL** — human in the loop, worked *with* a human who speaks for themselves — or **AFK**, driven by the agent alone. A HITL ticket only resolves through that live exchange; the agent never stands in for the human's side of it (a grilling agent that answers its own questions has broken this). Handing judgement to the agent wholesale is `/wf-auto`'s job, never a variant of this skill.
 
 - **Research** (AFK): Reading documentation, third-party APIs, or local resources like knowledge bases to surface a fact a decision waits on. Resolved by a `/research` **subagent**. Use when knowledge outside the current working directory is required.
 - **Prototype** (HITL): Raise the fidelity of the discussion by making a cheap, rough, concrete artifact to react to — an outline, a rough take, a stub, or UI/logic code via the /prototype skill. Links the prototype as an asset. Use when "how should it look" or "how should it behave" is the key question.
@@ -121,14 +121,12 @@ Resolution comments are unchanged — the answer still lands once, at close. And
 
 ## Invocation
 
-Two modes, plus a deferred variant of the second. **Never resolve more than one ticket per session** — with two exceptions: research tickets, and a **deferred launch**, which may work its named ticket's whole subtree.
+Two modes. **Never resolve more than one ticket per session** — with one exception: research tickets.
 
 The invocation grammar (what `wf`'s launch line produces):
 
-- `/wf <map> [<ticket>]` — interactive (everything below as written)
-- `/wf <map> <ticket> defer` — deferred subtree (see [Deferred mode](#deferred-mode))
-- `/wf <map> <ticket> defer: <steering>` — deferred, with a steering prompt
-- `/wf <map> <ticket> steer: <steering>` — interactive, with a steering prompt
+- `/wf <map> [<ticket>]` — everything below as written
+- `/wf <map> <ticket> steer: <steering>` — with a steering prompt
 
 ### Chart the map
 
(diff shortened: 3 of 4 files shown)


blooop (commented):
Fresh-context self-review (subagent that did not write this code), on two independent axes. AI-generated review, posted by the driving session.

## Standards

**No blocking findings.**

Non-blocking:
- `skills/wf/LIFECYCLE.md:3` now says "referenced by `/wf-auto` and honored by `/wf-tdd` and `/wf-review`", but `skills/wf-one/SKILL.md:32` also follows LIFECYCLE.md unchanged. That omission predates this PR (the old text named only the deferred mode too), and the edit's job was just to repair the dangling referrer, so this is an observation, not a defect in the diff.
- Nothing else. Specifically checked:
  - **Dangling references**: no `#deferred-mode` anchor survives anywhere; every intra-file anchor remaining in `skills/wf/SKILL.md` (`#ticket-types`, `#work-through-the-map`, `#breadcrumbs-handoffs-re-entry`, `#keep-the-map-inside-its-repo`) resolves to a real heading; relative links between skill files (`LIFECYCLE.md`, `../wf/LIFECYCLE.md`, `../wf/GITHUB_TRACKER.md`) all point at files that exist. The "Two modes, plus a deferred variant" line was correctly reduced to "Two modes" with the exception list trimmed to research tickets only.
  - **Comment quality**: the new test's comment in `src/model.rs` explains *why* the pin exists (derived `Ord` is the display order, three render sites trust it, other tallies stay green on a swap — #86 cited), not what the code does.
  - **Test would fail on regression**: measured, not assumed — see Spec.
  - **Gates (all run by the reviewer)**: `cargo fmt --all --check` clean; `cargo clippy --all-targets -- -D warnings` clean (exact flag used); `cargo test` fully green — 227 lib + 10 bin + all integration/live suites pass.

## Spec

**No blocking findings.**

#96 (ruling in the breadcrumb comment), point by point:
- **Deleted outright**: the entire "Deferred mode" section is gone from `skills/wf/SKILL.md`, plus the two `defer` grammar lines and the deferred exception to the one-ticket rule. The skill reads as though the mode never existed; the HITL rule's old exception is replaced with a pointer at `/wf-auto`, which is a repaired cross-reference, not a survival of the mode.
- **No `defer` a user can type under skills/**: `grep -rni defer skills/` run by the reviewer — zero matches (exit 1).
- **Carried content**: `skills/wf-auto/SKILL.md` read in full against the deleted section. Standing default prompt → principles + step 4 + the silence tiebreak; as-you-go claiming → line 83; steering-composes-after and cannot-lift-stops → line 110; stop conditions (a)(c)(d) → "Where it stops"; (b) out-of-scope is deliberately handled differently by /wf-auto's own pre-existing text, which supersedes. The one genuinely missing norm — *re-opening an agent-decided ticket is normal, not a conflict* — is exactly the sentence added at line 90. Nothing else moved, matching "content /wf-auto does not already say".
- **Historical mentions untouched**: the diff touches only 4 files; `src/skills.rs:7`, the `Mode` doc comments in `src/launch.rs` (89–90, 161), and both README passages (40, 337) still contain their `defer` history, unmodified.
- **Nothing more**: the diff contains exactly the two tickets' work — three markdown files and one test. No scope creep.

#86:
- The PR body's mutation claim is **measured independently by the reviewer**: swapped `Stage::Building` and `Stage::InReview` in the declaration, ran `cargo test --lib the_full_glyph_vocabulary_sorts_into_display_order`, got the red run — `assertion left == right failed: left: "○◍◐!●⊘", right: "○◐◍!●⊘"` at `src/model.rs:1093` — matching the PR body's quoted output. Reverted, confirmed `git status` clean, re-ran the test green.
- The pin is sound by construction as well: `RowGlyph::char()` matches by variant (order-independent) while the sort depends on the derived `Ord`, so the test isolates declaration order; the input array is deliberately scrambled so the sort does real work; it also pins `Blocked` sorting after all stages.
- Choice among the ticket's three candidates (single test vs. splitting the orders vs. documentation) is the one #86's own framing favors — the ticket records the #78 collapse as correct, and docs can't go red.

## What was measured vs only read

Measured: `grep -rni defer skills/` (empty); `cargo fmt --all --check`; `cargo clippy --all-targets -- -D warnings`; full `cargo test`; the Building/InReview swap mutation (red), revert, and green re-run; final `git status` (clean).
Read only: both tickets and their comments, the PR body, the full diff, the post-edit skill files, the `RowGlyph`/`Stage` declarations, and the preserved historical mentions.

## Verdict

APPROVE

Issue #122, opened 2026-08-08
What shape should the launch context handoff take?


blooop:
## Question

What shape should the context `wf` hands a launched agent take, and what is the agent allowed to trust about it?

`wf` parses every ticket's title, type, stage, blockers and PR rollup from one GraphQL query and then execs `claude` with `/wf-tdd 16`, throwing the rest away. The agent rediscovers it with serial `gh` calls before it does anything useful. The fix is obvious in outline and genuinely undecided in detail:

- **The carrier.** A JSON file under the session dir with its path in the prompt? An env var pointing at it? `--append-system-prompt`? The carrier has to survive the `dl` seam — an isolated launch runs the agent inside a container where the host's paths do not exist, and today only `~/.claude` is mounted. That constraint probably decides this on its own.
- **The contract.** Which fields, and what does the agent do when they disagree with the tracker? The snapshot is stale the moment it is written; a ticket can be claimed or closed between the two enters. "Trust it for orientation, re-read before mutating" is one answer; "trust it for everything except the claim" is another.
- **Who consumes it.** All five skills read `GITHUB_TRACKER.md`, and they are also invoked by hand (`/wf-auto` typed into a session that never went through the picker) — so the skills cannot *require* the handoff. It has to be an accelerator with a working fallback, not a new precondition.
- **What it costs when wrong.** A stale or malformed blob that an agent trusts is worse than no blob: it would act on a ticket someone else already took. Whatever the shape, the failure has to be detectable and cheap.

Resolving this unblocks the implementation slice; it does not include writing it.

blooop:
**breadcrumb:** claimed by autonomous run (wf-auto); settling the handoff shape against the map's principles and the repo.

blooop:
**agent-decided (simplicity):** the launch context rides the opening prompt itself, as a `ctx:` block between the skill arguments and any `steer:` suffix — no file, no env var, no new channel. Principle 2 decides it because the prompt is the one carrier that already provably crosses every seam `wf` has (host exec, the `dl` shell-command seam, and the Claude/Codex axis the ticket didn't mention), so every alternative adds a concept to get across a seam the prompt crosses for free. Principle 1 concurs: a year from now the whole handoff is visible in the launch line the human watched `wf` exec.

## The carrier: `ctx:` in the prompt

[...]

Issue #124, opened 2026-08-08
Hand the launched agent its ticket context


blooop:
## Question

Implement the launch context handoff in the shape its decision ticket settles on.

`wf` writes the context it already parsed; the skills read it when it is there and fall back to today's `gh` rediscovery when it is not. The agent's first `gh` call becomes the **claim** rather than a rediscovery of the map, the ticket and the blockers.

Scope notes for whoever takes this:

- The fallback is not optional. A skill invoked by hand never went through the picker and will find no context; that path has to stay exactly as good as it is now.
- The seam matters more than the saving. Isolated launches run in a container, so whatever the decision picked has to be verified *inside* one, not just on the host.
- Measure it. The agent-side rediscovery cost is currently an estimate; the before number has to exist before the after number means anything.

Blocked by the decision ticket — the carrier and the staleness contract are exactly what this needs handed to it.

blooop:
**breadcrumb:** unblocked by #122's resolution (ctx: prompt block, versioned claim-free schema, orient/verify-live/argv-wins contract); claimed by the autonomous run, entering build stage via wf-tdd.

blooop:
**breadcrumb:** building #122's resolution as specced: a `ctx:` one-line-JSON block in the launch prompt between the skill args and any `steer:` suffix, schema v=1 (claim-free, stage as Launchable, aim as the map/ticket sum), plus the shared consumer section in GITHUB_TRACKER.md and one grammar line per skill. Seams under test, as the resolution pre-agreed: the launch module's own prompt-building seam (agent_argv) and the shell-quoting seam the dl form rides (shell_quote's inverse, byte-for-byte) — red-first, with Plain/creation launches asserted ctx-free and no test requiring ctx for a skill to proceed.

blooop:
**breadcrumb:** resumption — previous run killed mid-build (no handoff). Verified against origin: no branch, no PR, nothing pushed; the in-flight worktree was discarded. Stage resets to build. Restarting wf-tdd from the plan its last breadcrumb recorded.

blooop:
[...]

Issue #123, opened 2026-08-08
Frontier in one query, not one round trip per open child


blooop:
## Question

Replace the frontier query's per-child `blocked_by` loop with the `blockedBy` the map query already returns.

`skills/wf/GITHUB_TRACKER.md` derives the frontier by fetching a map's sub-issues and then running one `gh api .../dependencies/blocked_by` **per open child, in a serial `while read` loop** — O(open children) round trips, paid by every session that works a map, and again by every subagent under `wf-auto`. `wf`'s own GraphQL query in `src/fetch.rs` already selects `blockedBy` alongside sub-issues and labels in a single hop, so the tracker doc is doing by hand, slowly, what the binary next to it does in one request.

This is a known slice, not a decision: the replacement query is the one `fetch.rs` already issues. Independent of the launch handoff — it makes the agent's own rediscovery cheaper whether or not the handoff ever lands, and it is the fallback path the handoff will need to keep working anyway.

Worth pinning as part of the slice: the current cost is **estimated, not measured**. Time the loop against a real map with a realistic number of open children before and after, and put the numbers in the PR — this map's Notes ask for that.

blooop:
**breadcrumb:** claimed by autonomous run (wf-auto); entering build stage via wf-tdd. Before-measurement of the serial blocked_by loop is part of the slice.

blooop:
**breadcrumb:** build plan restated: replace the tracker doc's frontier derivation (serial per-child blocked_by REST loop) with one `gh api graphql` call whose selection mirrors the map query the wf binary already issues (subIssues + assignees + blockedBy in one hop). Seam under test: the fenced frontier snippet in the GitHub tracker skill doc — a Rust test extracts it and asserts one round trip, no per-child while-read loop, and blockedBy coming from the map query. Before/after wall-clock numbers measured against the live map (#121) go in the PR body.

blooop:
**breadcrumb:** build slice done — PR #125 replaces the doc's frontier derivation with the single map-query GraphQL call the binary already issues, shape-pinned by a new offline test. Measured on the live map: ~0.91 s (loop as written) / ~1.42 s (every open child paying a round trip, ~0.26 s marginal per child) before, ~0.48 s after, flat in map size. Handing off to review.

blooop:
[...]

Issue #121, opened 2026-08-08
Launch latency: the agent starts from what wf already knows


blooop:
## Destination

Pressing enter on a ticket in `wf` and having an agent *productively working it* costs only the work that genuinely has to happen — no re-deriving state `wf` already holds, no serial round trips where one would do. The container half of that gap is `dl`'s and is out of scope here; this map is everything on `wf`'s side of the seam.

## Notes

**Domain.** The `wf` binary (`src/`) and the five skills it execs (`skills/`), which ship in this repo and are part of `wf`'s interface — a change to a `SKILL.md` is a change to `wf`'s behaviour, and belongs in the same PR as whatever motivated it. See `AGENTS.md`.

**Execution is in scope.** This map carries the execution override: build tickets are welcome, worked via `wf-tdd` and reviewed via `wf-review`. The remaining wins are mostly known rather than undecided — the map exists to sequence them and to settle the one real design question, not to plan at length.

**Measure, don't assume.** Every claim below is marked measured or estimated, and the estimated ones are estimates until someone times them. The earlier launch-latency effort's numbers are the baseline to beat.

**Baseline (measured 2026-08-08, blooop/devlaunch#138 and blooop/wayfinder#119).** Container-side cost has already come down a long way: one `devpod status` instead of a `list`+`status` pair took ~0.4–0.5s off every launch, lending the host's own `claude`/`gh` into a cold container moved 342MB in 5.1s where the old path curl-installed from the network, and `WF_PREWARM=1` hides the remaining cold start behind the launch picker. What is left is almost entirely **agent-side**.

**Disk is in scope (user steer, 2026-08-09).** Each launched instance costs a full clone + devcontainer + image layers (~1 GB per branch workspace, measured on #39; ~0.5–1.5 GB per abandoned prewarm stage, estimated from #39's measured components), and the user notes disk pressure may itself hurt startup latency, not only capacity. Prior art: #72 (teardown stays human-triggered, wf grows no sweep), PR #108 (`wf reap`, keyed on ticket-closed only), #39 (measured sizes). Reap is not to be redone — the open questions are what wf's lifecycle knowledge adds to it.

[...]

PR #125, opened 2026-08-09
Frontier in one query, not one round trip per open child

blooop:
Part of #123

## What changed

`skills/wf/GITHUB_TRACKER.md` derived a map's frontier by listing sub-issues and then running one `gh api .../dependencies/blocked_by` REST call **per open unclaimed child, in a serial `while read` loop** — O(open children) round trips per frontier read. The doc's Frontier query section now teaches the same single-hop GraphQL query the `wf` binary itself issues (`src/fetch.rs`): `subIssues` with `assignees` and `blockedBy` edges in one request, filtered to open + unclaimed + unblocked in `jq`. One round trip, however many children the map has, still pinned to the explicit `$REPO`.

A new offline test (`tests/skill_docs.rs`) extracts the fenced frontier snippet from the doc and pins its shape: exactly one `gh` call, GraphQL, no per-child `dependencies/blocked_by` loop, `subIssues`/`blockedBy`/`assignees` selected, explicit repo. Written red-first: both tests failed against the old snippet, pass against the new one.

## Measured, not assumed

Timed read-only against the live map **#121** on this repo (3 open children: #122, #123, #124; 1 unclaimed). Three runs each, wall clock:

| Derivation | Round trips | Runs (s) | Median |
|---|---|---|---|
| Before — doc snippet as written (assignee filter first, so 1 child pays the per-child call) | 2 | 1.107 / 0.897 / 0.905 | **0.91 s** |
| Before — same loop with every open child paying a round trip (N=3) | 4 | 1.403 / 1.431 / 1.418 | **1.42 s** |
| After — single GraphQL query | 1 | 0.414 / 0.484 / 0.488 | **0.48 s** |

Marginal cost of the old loop is ~0.26 s per open child on top of the base call; the new query stays flat as the map grows. The map's Notes asked for more open children than exist today — per the ticket, no throwaway issues were created on this public repo, so N=3 is what was measured, with the all-open-children variant included to show the per-child scaling. Both derivations return the same frontier on #121 (empty: #124 is open and unclaimed but blocked by open #122), and the raw GraphQL nodes were inspected to confirm the empty result is real data, not a silent query error.

## CI

All four green locally: `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test` (289 tests), `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps`.

## Summary by Sourcery

Align the GitHub tracker frontier query documentation with the binary’s single-hop GraphQL map query and add tests to pin the documented snippet’s shape.

Documentation:
- Update the GitHub tracker frontier query to use a single GraphQL request that retrieves sub-issues with assignee and blocked-by data instead of a per-child REST loop.

Tests:
- Add offline tests that extract the frontier snippet from the tracker doc and assert it is a single GraphQL map query using the expected selections and explicit repository targeting.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,13 +73,21 @@ jobs:
       - name: Clippy
         run: cargo clippy --all-targets --all-features --locked
 
-      # The unit tests, the ones that need nothing but a compiler. Everything
-      # under `tests/` is a `live_*` integration test — a real `gh`, real
-      # network, and assertions against this project's own tracker as it stands
-      # today — so running them here would make CI fail whenever the map moves
-      # on. They are compiled below instead of run.
+      # The unit tests, the ones that need nothing but a compiler. The `live_*`
+      # integration tests under `tests/` — a real `gh`, real network, and
+      # assertions against this project's own tracker as it stands today — are
+      # excluded, because running them here would make CI fail whenever the map
+      # moves on. They are compiled below instead of run.
       - name: Unit tests
         run: cargo test --locked --lib --bins --examples
+
+      # The other offline resident of `tests/`: shape checks on the doc
+      # snippets agents copy-paste, no network and no `gh` — the seam is the
+      # doc text itself. It has to be named here because the step above stops
+      # at `--lib --bins --examples`, and an assertion no workflow runs is not
+      # an assertion.
+      - name: Skill doc shape checks
+        run: cargo test --locked --test skill_docs
 
       # The one part of `tests/` that needs neither network nor `gh`: the
       # fixture's own diagnostic, which decides what a failing live test tells
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -112,15 +112,18 @@ RUSTFLAGS=-D\ warnings RUSTDOCFLAGS=-D\ warnings sh -c '
   cargo fmt --all --check &&
   cargo clippy --all-targets --all-features --locked &&
   cargo test --locked --lib --bins --examples &&
+  cargo test --locked --test skill_docs &&
   cargo doc --no-deps --all-features --locked'
 ```
 
-Run all four before pushing, `cargo fmt --all` first — a formatting diff fails
-the build before any of the interesting checks get a chance to run.
+Run the whole chain before pushing, `cargo fmt --all` first — a formatting diff
+fails the build before any of the interesting checks get a chance to run.
 
 The `tests/live_*.rs` files are excluded from that test command on purpose: they
 talk to real GitHub and drive a real pty. Run them by name when the thing you
-changed is what they cover.
+changed is what they cover. `tests/skill_docs.rs` is the exception under
+`tests/` — offline shape checks on the skill docs' snippets, so it runs in the
+chain (and in CI) like any unit test.
 
 ## House style
 
--- a/skills/wf/GITHUB_TRACKER.md
+++ b/skills/wf/GITHUB_TRACKER.md
@@ -89,16 +89,26 @@ An open, unassigned ticket is unclaimed. Claim **before any work**.
 
 ## Frontier query
 
-Open, unblocked, unclaimed children of the map:
+Open, unblocked, unclaimed children of the map — one round trip, however many children the map has. This is the same single-hop GraphQL query the `wf` binary issues to render a map: every child arrives with its `blockedBy` edges attached, so there is nothing left to fetch per child. (GraphQL spells enum values in caps: `OPEN`, not `open`.)
 
 ```bash
-gh api "repos/$REPO/issues/MAP_NUMBER/sub_issues" --paginate \
-  --jq '.[] | select(.state == "open" and (.assignees | length == 0)) | .number' |
-while read -r n; do
-  open_blockers=$(gh api "repos/$REPO/issues/$n/dependencies/blocked_by" \
-    --jq '[.[] | select(.state == "open")] | length')
-  [ "$open_blockers" -eq 0 ] && echo "$n"
-done
+gh api graphql -F owner="${REPO%/*}" -F name="${REPO#*/}" -F number=MAP_NUMBER -f query='
+query($owner: String!, $name: String!, $number: Int!) {
+  repository(owner: $owner, name: $name) {
+    issue(number: $number) {
+      subIssues(first: 100) {
+        nodes {
+          number state
+          assignees(first: 5) { nodes { login } }
+          blockedBy(first: 50) { nodes { number state } }
+        }
+      }
+    }
+  }
+}' --jq '.data.repository.issue.subIssues.nodes[]
+  | select(.state == "OPEN" and (.assignees.nodes | length == 0)
+      and ([.blockedBy.nodes[] | select(.state == "OPEN")] | length == 0))
+  | .number'
 ```
 
 ## Journal on a ticket (breadcrumbs and handoff)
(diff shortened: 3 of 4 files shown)


blooop (commented):

Two-axis review of #125 at head 2fa3839 against merge-base 24f1be1 (diff: `git diff 24f1be1...2fa3839`). Spec is ticket #123. Axes are independent and not merged.

## Standards

Repo standards consulted: AGENTS.md, the CI workflow's normative comments, README. No CHANGELOG exists or is expected — no finding there.

1. **`tests/skill_docs.rs` is never run in CI** — **blocking**. CI's "Unit tests" step is `cargo test --locked --lib --bins --examples` (`.github/workflows/ci.yml:82`), which excludes `tests/`; the only `tests/` target CI executes is `--test live_fetch -- common::` (line 94). The new guard test is therefore compiled (clippy `--all-targets`, the `--no-run` step) but never executed by any workflow. The repo's own CI comment states the standard this violates: *"an assertion no workflow runs is not an assertion."* A shape-pinning test added specifically as enforcement, that no workflow runs, is dead weight — the doc could regress to the per-child loop and CI would stay green. Fix: add a run step (e.g. `cargo test --locked --test skill_docs`) and consider whether AGENTS.md's local check command should mention it.
2. **CI comment now false** — should-fix, resolves with finding 1. `.github/workflows/ci.yml:76-80` justifies the exclusion with "Everything under `tests/` is a `live_*` integration test — a real `gh`, real network." `skill_docs.rs` is offline by design (its own module doc says so). In a repo whose house style treats comments as load-bearing, this comment now lies. `AGENTS.md`'s "The `tests/live_*.rs` files are excluded..." similarly now reads as exhaustive when it isn't.
3. **Prose overpromises pagination** — nit. `skills/wf/GITHUB_TRACKER.md:92` says "one round trip, however many children the map has," but `subIssues(first: 100)` (and `blockedBy(first: 50)`) truncate silently past those limits. The limits mirror `src/fetch.rs` exactly, so the "same query the binary issues" claim holds; the prose just claims slightly more than the query delivers.

Otherwise clean: the doc edit fits the file's why-first teaching voice, stays pinned to explicit `$REPO`, and remains copy-pasteable and self-contained (required of this fallback path by the #122 decision). The test matches repo test style (design-decision doc comments, message-bearing asserts). Fowler baseline: no genuine hits — the repeated `split/nth/expect` extraction in `frontier_snippet()` is borderline Duplicated Code, below the threshold worth a helper. Hygiene: no hardcoded `/home/<user>` paths (uses `CARGO_MANIFEST_DIR`), no secrets, no visibility leaks.

## Spec

All checks were run hands-on, not read off the diff.

1. **New snippet works as written — verified live.** Spec: *"Replace the frontier query's per-child `blocked_by` loop with the `blockedBy` the map query already returns."* The head snippet was extracted and run read-only against `REPO=blooop/wayfinder`, map #121: exit 0, valid data, empty frontier. The raw nodes confirm the empty result is real data, not a silent query error (#122 CLOSED; #123 OPEN/assigned; #124 OPEN blocked by closed #122 but assigned; #126 OPEN/unassigned blocked by open #124). Cross-checked against raw REST `dependencies/blocked_by` for #123/#124/#126 — GraphQL and REST imply the identical frontier. (Live state has drifted since the PR body was written — #122 closed, #126 added — both states yield the same empty frontier consistently.)
2. **Same selection as the binary.** Spec: *"the replacement query is the one `fetch.rs` already issues."* The doc query is a recognizable subset of `src/fetch.rs`'s map query — identical `repository → issue → subIssues(first: 100)` shape with `number state`, `assignees(first: 5)`, `blockedBy(first: 50)` — omitting only fields the frontier doesn't need (`title`, `labels`, `closedByPullRequestsReferences`). No material divergence.
3. **The test is real — red/green flip reproduced.** With the base version of `GITHUB_TRACKER.md` written into a head worktree, `cargo test --test skill_docs` fails both tests; restored, both pass. Not vacuous (a regression to a per-child loop cannot sneak past) and not over-fitted (assertions target semantics — one `gh` call, GraphQL, the three selections, explicit `$REPO` — not incidental prose; the only structural coupling is the section heading and the bash fence, a fair contract for a copy-pasteable snippet).
4. **Measurements present and coherent.** Spec: *"Time the loop against a real map ... before and after, and put the numbers in the PR."* The PR body carries a 3-row table (medians 0.91 s / 1.42 s / 0.48 s, 3 runs each); internally consistent (~0.45 s per round trip; marginal ≈ 0.26 s/child matches the stated figure). N=3 is fewer open children than the map Notes hoped for, but the PR states why (no throwaway issues on a public repo) and includes the all-open-children variant to show scaling. Acceptable.
5. **Old loop gone.** No `while read` remains in the doc; remaining `dependencies/blocked_by` mentions are edge create/inspect/delete commands, not frontier derivation.

No blocking Spec findings — the diff delivers the ticket's stated behavior in full.

## Verdict

**Request changes.** Blocking finding: Standards #1 — `tests/skill_docs.rs` must actually be executed by a CI step (e.g. `cargo test --locked --test skill_docs`); fix the now-false CI comment at `.github/workflows/ci.yml:76-80` (Standards #2) in the same change. Standards #3 is a take-it-or-leave-it nit. The Spec axis passes outright — once the guard test runs in CI, this is an approve.

blooop:
Review fixes pushed in a67332b:

- **Standards #1 (blocking):** CI now actually runs the guard test — a new `Skill doc shape checks` step (`cargo test --locked --test skill_docs`) sits after the unit-test step, with a why-comment in the workflow's own words: an assertion no workflow runs is not an assertion.
- **Standards #2:** the unit-test step's comment no longer claims everything under `tests/` is a live network test; AGENTS.md's local check chain now includes the skill_docs command and names `tests/skill_docs.rs` as the offline exception.
- **Standards #3 (nit):** the doc prose (and the test's doc comment) now say one round trip for up to 100 children — the snippet's `first: 100`, same cap as the binary — instead of "however many children".

All four local checks pass at head, including `cargo test --locked --test skill_docs` (2 passed).

> *This comment was generated by AI in response to review.*

blooop (commented):

Follow-up to the review at [pullrequestreview-4890720910](https://github.com/blooop/wayfinder/pull/125#pullrequestreview-4890720910), re-verified at new head a67332b (delta `git diff 2fa3839..a67332b`).

All three findings are resolved, checked against the live artifacts, not just the diff:

1. **Blocking — guard test now runs in CI.** `.github/workflows/ci.yml` gains a "Skill doc shape checks" step running `cargo test --locked --test skill_docs`. Verified in the live CI run on this exact SHA (run 31300063237): the step executed `tests/skill_docs.rs` — `test result: ok. 2 passed; 0 failed`. AGENTS.md's local check chain gains the same command.
2. **Should-fix — the false comment is gone.** The ci.yml comment no longer claims everything under `tests/` is a live network test; the new step's comment even cites the house rule it exists to satisfy. AGENTS.md now names `skill_docs.rs` as the offline exception.
3. **Nit — prose no longer overpromises.** "however many children the map has" is now "up to 100 children (the `first: 100` below, the same cap the binary uses)", and the test's doc comment matches.

The delta touches nothing else: the frontier snippet and the test assertions are byte-identical to what the original review verified hands-on (snippet run live against map #121, red/green flip reproduced), so the Spec axis's clean pass carries over unchanged.

## Verdict

**Approve.** No open findings on either axis. (GitHub refuses an approving review state on a same-account PR, so this written verdict is the gate's signal.)

Issue #128, opened 2026-08-09
Reap the prewarm nobody entered


blooop:
## Question

A prewarmed workspace nobody ever entered: whose fact is "never attached", and what should `wf reap` do with it?

`WF_PREWARM` warms a staged node's container while the human is still typing; backing out of the launch leaves the workspace, container, and clone behind, and `wf reap` will not collect them while the ticket is open. This is real in `wf`'s own words — `src/launch.rs` carries both warnings ("a launch you back out of leaves all three behind — and `wf reap` will not collect them while the ticket is open"; "an abandoned stage of an open ticket waits for a hand to remove it"). Cost per abandoned stage: roughly 0.5–1.5 GB, estimated from [#39's measured components](https://github.com/blooop/wayfinder/issues/39) (clone + container + 342 MB lent tools + up to ~1 GB target once anything builds) — and the container may be Running, which reap never waives even after the ticket closes.

`wf` is the only party that knows "staged then backed out". The open design choice is where the never-attached fact lives:

- a wf-side record of the prewarm it fired (written at prewarm, cleared on a real launch), or
- asking `dl`/devpod for a last-attach time and inferring it.

And what reap does with the fact: a new claimable verdict ("prewarmed for #n, never entered, ticket untouched since") versus a keep-with-warning line in the listing. [#72's no-sweep decision](https://github.com/blooop/wayfinder/issues/72) stands: reap stays human-triggered; the prewarm's own silent-failure design stays (a prewarm must never make a launch louder or slower).

Resolving this unblocks an implementation slice; it does not include writing it. The resolution must say how the outcome will be tested. Resolve consistently with the sibling ticket "When is a workspace dead weight before its ticket closes?" — same verdict vocabulary, same guard order.

Prior art to read first: #72, PR #108, #39, and the prewarm caveat comments in `src/launch.rs`.

blooop:
**breadcrumb:** unblocked by #127's resolution (NodeFact sum, Verdict::Warn arm, stage-lattice-derived predicate); claimed by the autonomous run, resolving consistently with that vocabulary.

blooop:
**breadcrumb:** resumption — the previous autonomous run was killed here with no handoff; its trail ends at the claim, and nothing was recorded. Restarting the resolution from scratch against #127's settled vocabulary.

blooop:
[...]

Issue #127, opened 2026-08-09
When is a workspace dead weight before its ticket closes?


blooop:
## Question

When is a workspace dead weight *before* its ticket closes — and which of the signals `wf` already holds should let `wf reap` say so?

Today reap's only liveness signal is ticket state: open means never collected, regardless of disk cost or staleness (`src/reap.rs` — the ticket-open guard). But `wf` knows more than open/closed: it holds every node's stage, PR rollup, and blockers from one GraphQL query (`src/fetch.rs`). A workspace whose node is in-review with its PR merged, or whose branch was superseded by a re-stage, is ~1 GB of dead weight ([measured on #39](https://github.com/blooop/wayfinder/issues/39): ~1 GB per branch workspace) that reap will not even name.

The decision is only which signals count as "finished enough" — the predicate, the data source, and the guard order all already exist ([PR #108](https://github.com/blooop/wayfinder/pull/108)). Constraints that stand:

- [#72's settled posture](https://github.com/blooop/wayfinder/issues/72) is untouched: no sweep, no timer — reap stays human-triggered; this ticket only sharpens what a triggered reap may claim (or list as a warning).
- The existing guard order stays: unsaved/unpushed work waivable with `-f`, a Running container never waivable.
- Whatever is added must come from data the map query already returns — no new per-workspace round trips (this map's whole point).

Resolving this unblocks an implementation slice; it does not include writing it. The resolution must say how the outcome will be tested (reap has existing verdict-level tests to extend).

Prior art to read first: #72 (teardown decision), PR #108 (reap's contract), #39 (measured sizes). Sibling ticket "Reap the prewarm nobody entered" covers the staged-then-abandoned case specifically; this ticket is the general lifecycle-signal question — resolve consistently.

blooop:
**breadcrumb:** claimed by the autonomous run (wf-auto, disk steer). Working this before its sibling #128 so the two resolutions share one verdict vocabulary.

blooop:
**agent-decided (1 long-term maintainability):** reap claims exactly what wf's own stage lattice already calls **Done**, and warns on what the model already calls **no evidence** — no new judgment vocabulary, one derivation shared with the screen.

## The predicate

Reap's tracker question widens from "is the ticket closed?" to the reap-relevant reading of the *same fields the map query already returns* (`state` + `closedByPullRequestsReferences` with per-PR `state`/`isDraft` — src/fetch.rs's `MAP_QUERY`). Per node, three facts beyond plain open:

[...]

Issue #72, opened 2026-08-06
Container teardown and cleanup


blooop:
## Question

Who ends a container's life, and when? The engine research ([Devcontainer CLI as wf's provisioning engine](https://github.com/blooop/wayfinder/issues/68)) found the devcontainer CLI has **no `stop`/`down` subcommand** — wf owns the teardown end of the lifecycle itself, via `docker stop`/`docker rm` against the `devcontainer.local_folder` label. Decide:

- When does wf stop a container — on agent exit, on idle, never (leave running for reuse)?
- Does wf honor the config's `shutdownAction`?
- Garbage collection of stale containers and images from checkouts that no longer exist — a `wf`-driven sweep, or leave it to the user/docker prune?
- What surface (if any) does the TUI expose for "stop/remove this repo's env"?

Grilling ticket — best resolved after the prototype has produced real lifecycle experience.

blooop:
**breadcrumb:** [Credential posture and config-injection mechanics](https://github.com/blooop/wayfinder/issues/73) simplifies this ticket's scope: the settled posture uses a live rw bind-mount of host `~/.claude`, not a per-container config volume, so **teardown has no credential-bearing volume to scrub or garbage-collect** — only the container itself. The agent binary is installed in-container via `--additional-features`, so it lives in image layers, which is a separate (image-cache) cleanup question from container teardown.

blooop:
**breadcrumb:** Concrete teardown facts from [blooop/devlaunch#50](https://github.com/blooop/devlaunch/pull/50), found by driving `dl` against a real compose/GPU/host-network project — carry these into this ticket rather than rediscovering them:

- **`devpod delete` re-parses the workspace's `devcontainer.json`** to tear the container down, so **deletion fails if that file has moved or been renamed**. devpod's lifecycle subcommands are real (unlike the devcontainer CLI, whose lack of `down` [#68](https://github.com/blooop/wayfinder/issues/68) recorded as a gap) but they are **not unconditional**.
- The matching failure mode: devlaunch used to remove the local clone regardless of whether `devpod delete` succeeded, leaving devpod with no config to ever retry from — a **stranded workspace with no escape hatch**. The fix keeps the clone unless devpod succeeded. Whatever wf does here needs the same ordering: don't destroy the thing the teardown depends on until teardown has actually succeeded.
- Compose-based projects bring up **sidecar containers**; teardown has to account for more than one container per workspace.

[...]

PR #130, opened 2026-08-09
Reap what the lattice calls Done, warn at its other end

### LINKED ISSUE #129: Reap the dead weight the tracker already names
## Question

Implement the dead-weight verdicts its decision tickets settled on: `wf reap` claims what the stage lattice already calls Done, and warns at the lattice's other end — on Superseded, and on a node nothing has come of.

The full design is in **two** resolutions, and both are implemented here. Implement them; do not re-design.

- [When is a workspace dead weight before its ticket closes?](https://github.com/blooop/wayfinder/issues/127) — the predicate table, the `NodeFact`/`Verdict::Warn` type sketch, the guard interaction, the batched per-repo GraphQL data source, and its eight red-first test cases.
- [Reap the prewarm nobody entered](https://github.com/blooop/wayfinder/issues/128) — **folded in as of its resolution** (see the scope note below).

Scope notes:

- Reuse the existing per-PR interpretation (`fetch.rs`'s `pr_status` / the `model::stage()` table) rather than restating it — one vocabulary for "done" is the point of the decision.
- The signature change (`plan()` taking the fact map) breaks existing tests' compilation first; that is the migration's own red. The named cases in the two resolutions are the test list (#127's eight plus #128's ten, which overlap on the regression pins — roughly 15 distinct tests).
- **`NodeFact` carries six arms, not four.** #128 splits #127's `Open` catch-all on the claim bit into `InFlight` / `Claimed` / `Unstarted`, with `Unstarted` (open, zero linked PRs, unassigned) the new **Warn**. Take the derivation order verbatim from #128's resolution; it reproduces #127's four arms exactly. `Unstarted` must be a positive observation, never reachable from a failed lookup — a node the batch did not answer for is an error, not `Unstarted`.
- **Select `assignees(first: 5)`** in the batched per-repo query. It is already in the `MAP_QUERY` field subset and `model::classify()` already interprets it as the claim, so the round-trip claim below is unaffected: zero additional round trips.
- **No change to `src/launch.rs` or `src/app.rs`.** #128's resolution turns on the prewarm path being untouched — nothing written at prewarm, nothing cleared at launch, no new syscall on the path that `exec`s. A diff that touches the prewarm path has misread the decision. The existing `claim_prewarm` and `prewarm` tests stay green untouched; they are the pin.
- **Zero new deletion paths.** `Warn` is display-only. A warned workspace must stay out of the doomed set even under `-y -f`, and that invariant gets its own test.
- `finished_nodes`' per-issue REST loop becomes one `gh api graphql` call per repo — strictly fewer round trips, zero per-workspace calls; fix the doc comment that already claims this.
- Nothing else moves: no sweep, no size/age heuristics, `-f` semantics unchanged, Running never waivable.
- Measure, don't assume (map Notes): time `wf reap`'s tracker phase before and after against a real multi-workspace state if available, or at minimum count the round trips per repo before/after in the PR body, marking each number measured or derived.

*Scope amended 2026-08-09, on [#128](https://github.com/blooop/wayfinder/issues/128) resolving.* This ticket originally said #128's outcome "must not be implemented here", which was the right guard while #128 was open. Its resolution turned out to be the same enum, the same match sites and the same migration this ticket already performs — no signature change beyond #127's. Splitting it would migrate `plan()`'s callers and the verdict tests twice, which is the more expensive and less reviewable path (principle 2, simplicity). So the two land together.

blooop:
Reap now claims what `wf`'s own stage lattice already calls **Done**, and warns
at the lattice's other end. Implements the resolutions of #127 and #128; the
predicate, the verdicts and the type sketch are theirs, not this PR's.

Closes #129. Also lands #128's slice, folded in per #129's amended scope note —
same enum, same match sites, same migration, so splitting it would have migrated
`plan()`'s callers and the verdict tests twice.

## What changed

| tracker fact | signals | verdict |
| --- | --- | --- |
| `Closed` | issue CLOSED | **Reap** — unchanged, same words |
| `DoneByMerge` | OPEN, ≥1 merged PR, zero open/draft | **Reap** |
| `Superseded` | OPEN, ≥1 PR, all closed-unmerged | **Warn** |
| `InFlight` | OPEN, ≥1 open or draft PR | **Keep** |
| `Claimed` | OPEN, zero PRs, assigned | **Keep** |
| `Unstarted` | OPEN, zero PRs, unassigned | **Warn** |

`plan()` takes a `BTreeMap<Node, NodeFact>` where it took a `BTreeSet<Node>` of
closed nodes, and `Verdict` grows a third arm. Guard order is untouched:
unsaved → Keep (waivable by `-f`), Running → Keep (never waivable), then the
fact. Worse news still wins the row.

## The one place this departs from #128's sketch

#128 sketched `node_fact(is_open, is_assigned, prs: &[PrLink])`. **That
signature cannot satisfy #128's own totality test.** A linked PR whose `state`
this binary does not recognise (`fetch.rs`'s `"SOMETHING_NEW"` fixture) yields
no badge *on purpose* — so it is not representable as a `PrLink`, it would be
dropped before the derivation saw it, and the node would read as having no PR
at all and slide into `Unstarted`. That is exactly the sentinel #128 exists to
forbid: a warning invented out of a parse failure.

The fix is a type, not a re-design. The derivation takes `PrOutcome` — a total
three-arm projection of the *existing* per-PR interpretation (in flight /
merged / closed-unmerged), each arm carrying its number, with an **unreadable
PR landing on in-flight**: the only claim that stays true whatever a newer
GitHub state means, and the arm that keeps the workspace and prints nothing. A
state added after this binary ships costs a workspace that stays, never one
that goes. It is a projection of `fetch::parse_pr`'s reading rather than a
second reading of the tracker's strings, so #127's one-vocabulary constraint
holds — `parse_pr` is now `pub(crate)` and the batch reads PR nodes through it.

Everything else in #128 stands as resolved: predicate, verdicts, and the
Warn-never-deletes posture.

Also settled while building: **the Warn rows do not borrow Reap's "discarding"
cadence** — nothing is discarded on a row that never deletes. Under `-f` they
read `… reap by hand if so, discarding <unsaved>`, where the discard belongs to
the by-hand reap being advised. Same wording for both warn facts.

## Safety

**Zero new deletion paths.** `reap::doomed()` is now the single definition of
what gets removed, in the reap module rather than as a `matches!` at the call
site — which is what makes *"a warned workspace never reaches the doomed set
even under `-y -f`"* a test with an owner instead of a coincidence between two
partitions. `-f` semantics unchanged, Running never waivable, no new flags.

`Unstarted` is a positive observation, never an absence: a node the batch did
not answer for **fails the whole call**, and `plan` refuses to act on a node it
has no fact for even if one ever reached it.

## Round trips, per repo, per `wf reap`

| | before | after |
| --- | --- | --- |
| `dl --ls --json` | 1 | 1 |
| tracker calls | **1 REST call per node** | **1 GraphQL call per repo** |
| per-workspace calls | 0 | 0 |

Both rows *derived* from the code: the old `finished_nodes` looped
`issue_is_closed` once per node; the new one aliases every wanted number into
one query per repo. The doc comment that already claimed "one search per repo"
is finally true.

**Measured**, on this repo, 8 real nodes, 3 runs each, warm `gh` auth:

```
one batched graphql call:      0.44 / 0.41 / 0.44 s
eight per-issue REST calls:    2.52 / 2.58 / 2.52 s
```

~6x on the tracker phase at 8 nodes, and the gap widens with node count since
the new shape is flat in it. This is `gh` invocation time against the live API,
not `wf` itself — `wf reap`'s tracker phase is one such call per repo either
way, so it is the phase's cost. No multi-workspace `dl` state was available
here, so the end-to-end `wf reap` timing is **not** measured.

## Scope held

- **`src/launch.rs` and `src/app.rs` have an empty diff.** #128's resolution
  turns on the prewarm path being untouched: nothing written at prewarm,
  nothing cleared at launch, no syscall added to the path that `exec`s. The
  existing `claim_prewarm` and `prewarm` tests are the pin and are green
  untouched.
- No sweep (#72 stands), no size or age heuristics, no `headRefName`.
- `fetch.rs` changes are the minimum to share one vocabulary: the GraphQL
  envelope is now generic over its selection, and the PR node type, the badge
  parse and `is_open` are `pub(crate)`.

## Tests

Red-first, one slice at a time: the derivation, then the `plan()` migration,
then the batched parse, then the printer. 21 new tests in `src/reap.rs` —
the six arms table-driven for totality over
{closed, open} × {none, draft, open, merged, closed-unmerged, merged+open,
**unreadable**} × {assigned, unassigned}; the two Warn facts and their
under-fires (Running hides it; a dirty lockfile hides it until `-f`); the
doomed-set invariant; and parse-boundary fixtures for the batch including the
never-partial posture and `assignees` reaching the fact.

Local suite: **310 passed, 0 failed** (`cargo test`), with `cargo fmt --check`,
`cargo clippy --all-targets --all-features -D warnings` and `cargo doc` clean.

## Summary by Sourcery

Align `wf reap` with the tracker’s stage lattice, treating tickets as finished when closed or done-by-merge, and introduce non-deleting warnings for suspected but uncertain dead work.

New Features:
- Add a `Warn` verdict alongside `Reap` and `Keep` so `wf reap` can surface suspected dead work without deleting it.
- Derive per-node `NodeFact` (closed, done-by-merge, superseded, in-flight, claimed, unstarted) from tracker data to drive reap decisions.
- Introduce `PrOutcome` as a projection of per-PR status so unreadable or new GitHub states are treated as PRs in flight rather than misclassified.
- Batch tracker lookups with a per-repo GraphQL query that fetches issue state, assignees and linked PRs in one call, shared with the map fetch envelope.

Enhancements:
- Refactor deletion selection into a dedicated `doomed` helper so only reap verdicts can lead to actual workspace removal.
- Rework `plan` to consume a map of node facts instead of a set of closed nodes, making keep/reap/warn decisions explicit and testable.
- Group CLI output for `wf reap` by keep/warn/reap and clarify the prompt to only count truly doomed workspaces.
- Expose selected `fetch` types and helpers (`GraphQlResponse`, `Nodes`, `Assignee`, `PrNode`, `parse_pr`, `is_open`) as `pub(crate)` to share one vocabulary between the map view and reap.

Documentation:
- Update `wf reap` help text and README to describe done-by-merge semantics, non-deleting warn rows, and the use of batched GraphQL calls per repo.

Tests:
- Add an extensive test suite for `PrOutcome`, `NodeFact`, `plan`, `doomed`, and GraphQL batch parsing, covering unreadable PR states, guard precedence, warn invariants, and batch error handling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -604,16 +604,43 @@ an agent exiting. `dl <ws> stop`, `dl <ws> rm` and `dl --ls` are yours to run.
 ### `wf reap` — clearing away finished tickets
 
 A workspace per ticket means workspaces accumulate as fast as tickets are
-worked. `wf reap` removes the ones whose tickets are **closed**:
+worked. `wf reap` removes the ones whose **work is over**:
 
 ```
 $ wf reap
   keep  devlaunch-wayfinder-devlaunch-131-a  (devlaunch#131 is still open)
   keep  wayfinder-wayfinder-wayfinder-42-b   (still running — stop it first)
+  warn  devlaunch-wayfinder-devlaunch-118-d  (devlaunch#118's PR #120 closed unmerged — superseded? reap by hand if so)
+  warn  wayfinder-wayfinder-wayfinder-96-e   (wayfinder#96 unclaimed and no PR — an abandoned stage? reap by hand if so)
   reap  devlaunch-wayfinder-devlaunch-127-c  (devlaunch#127 is closed)
+  reap  devlaunch-wayfinder-devlaunch-80-f   (devlaunch#80 open but its PR #97 merged)
 
-delete 1 workspace(s)? [y/N]
+delete 2 workspace(s)? [y/N]
 ```
+
+"Over" is not a judgement `wf` invents for the occasion — it is exactly what
+the stage lattice already calls Done, read off the same fields
+as the `⇄` badge on the screen: a **closed** ticket, or an **open** one whose
+PR merged with nothing still in flight. The second case is the one that earns
+this its keep, since a ticket often outlives the branch that finished it. And
+because the guards run first, an un-`-f`'d reap only ever deletes a checkout
+whose every byte exists on the remote: being wrong costs a re-clone, never
+work.
+
+`warn` rows are the other end of that lattice, and they are **never deleted** —
+no flag makes them so. `wf` prints them because it suspects dead weight on
+evidence too weak to act on:
+
+- every linked PR **closed unmerged** — a human's "not this way", which is not
+  the same as "this branch is disposable";
+- **nobody claimed it and nothing came of it** — no PR, no assignee. An open,
+  unassigned ticket is unclaimed by `wf`'s own convention, and that one bit is
+  what keeps this from firing on every ticket someone is mid-way through.
+
+A stale claim therefore keeps a workspace: an agent that died leaves its ticket
+assigned, and reap does not overrule a person's stated intent. `wf` says only
+what it observed — it does not know whether anyone ever entered a container,
+and does not claim to.
 
 This is the same division of labour the launch draws: **`dl` owns the
 containers, `wf` owns the tickets.** `dl` deliberately does not decide what is
@@ -625,8 +652,12 @@ decides. Needs `dl` **0.0.21 or newer** for the JSON listing.
 
 Kept, always: workspaces `dl` did not create (they are not `wf`'s, whatever
 their branch looks like), branches that are not `wayfinder/<repo>-<n>` for that
-repo, tickets still open, and anything **running** — a ticket closing is no
+repo, tickets with an open or draft PR (in review is where review fixes happen),
+tickets someone has claimed, and anything **running** — a ticket closing is no
 evidence that the session in the container ended.
+
+One `gh api graphql` call per repo answers all of this, however many workspaces
+that repo has.
 
 Kept by default but waivable: a workspace whose clone holds uncommitted or
 unpushed work. `-f` reaps those too, and the plan says what it is discarding:
@@ -640,6 +671,7 @@ unpushed work. `-f` reaps those too, and the plan says what it is discarding:
 *every* workspace it builds, so without it those are unreapable forever. It
 waives the unsaved-work guard only — a running container is still kept, because
 that is a session in progress rather than bytes on disk. `-y` skips the prompt.
+Neither flag can turn a `warn` row into a deletion.
 
 Your host `~/.claude` is bind-mounted in, which is how the Claude agent arrives
 already logged in — and is the reason `wf skills install` keeps the prompts
--- a/src/fetch.rs
+++ b/src/fetch.rs
@@ -66,16 +66,19 @@ query($owner: String!, $name: String!, $number: Int!) {
   }
 }";
 
+/// The envelope every `gh api graphql` answer arrives in. Generic over the
+/// selection because reap batches its own query through the same shape (#129),
+/// and "errors instead of data" must mean the same thing to both readers.
 #[derive(Deserialize)]
-struct GraphQlResponse {
-    data: Option<ResponseData>,
+pub(crate) struct GraphQlResponse<T> {
+    pub(crate) data: Option<T>,
     #[serde(default)]
-    errors: Vec<GraphQlError>,
+    pub(crate) errors: Vec<GraphQlError>,
 }
 
 #[derive(Deserialize)]
-struct GraphQlError {
-    message: String,
+pub(crate) struct GraphQlError {
+    pub(crate) message: String,
 }
 
 #[derive(Deserialize)]
@@ -103,8 +106,8 @@ struct MapIssue {
 }
 
 #[derive(Deserialize)]
-struct Nodes<T> {
-    nodes: Vec<T>,
+pub(crate) struct Nodes<T> {
+    pub(crate) nodes: Vec<T>,
 }
 
 impl<T> Default for Nodes<T> {
@@ -135,7 +138,7 @@ struct Label {
 }
 
 #[derive(Deserialize)]
-struct Assignee {
+pub(crate) struct Assignee {
     #[allow(dead_code)]
     login: String,
 }
@@ -147,8 +150,8 @@ struct Blocker {
 }
 
 #[derive(Deserialize)]
-struct PrNode {
-    number: u64,
+pub(crate) struct PrNode {
+    pub(crate) number: u64,
     state: String,
     #[serde(rename = "isDraft")]
     is_draft: bool,
@@ -177,7 +180,7 @@ struct RepoRef {
 /// wrong one. The inner strings are open GraphQL enums too: an unknown check
 /// rollup or review decision reads as "in flight", the only claim that stays
 /// true whatever the new value means.
-fn parse_pr(pr: &PrNode) -> Option<PrLink> {
+pub(crate) fn parse_pr(pr: &PrNode) -> Option<PrLink> {
     let status = match (pr.state.as_str(), pr.is_draft) {
         ("MERGED", _) => PrStatus::Merged,
         ("CLOSED", _) => PrStatus::Closed,
@@ -205,7 +208,7 @@ fn parse_pr(pr: &PrNode) -> Option<PrLink> {
     })
 }
 
-fn is_open(state: &str) -> bool {
+pub(crate) fn is_open(state: &str) -> bool {
     state == "OPEN"
 }
 
@@ -258,7 +261,7 @@ pub async fn fetch_map(id: &MapId) -> Result<Map> {
 /// network. Every raw tracker string a ticket is derived from (state,
 /// assignees, blocker states, labels) is interpreted exactly here.
 fn parse_map(body: &[u8], id: &MapId) -> Result<Map> {
-    let resp: GraphQlResponse =
+    let resp: GraphQlResponse<ResponseData> =
         serde_json::from_slice(body).context("unparseable GraphQL response from gh")?;
     if let Some(err) = resp.errors.first() {
         bail!("GraphQL error: {}", err.message);
--- a/src/main.rs
+++ b/src/main.rs
@@ -78,10 +78,14 @@ ctrl-g narrow and widen the scope, ctrl-r refetches, esc quits.
 
 wf skills          report which prompt each route would actually run
 wf skills install  link this build's skills into ~/.claude/skills and ~/.codex/skills
-wf reap            remove the workspaces whose tickets are closed (-y to skip
-                   the prompt). Keeps anything running or holding work that is
-                   not pushed anywhere; -f reaps the unpushed ones too, naming
-                   what it discards. Needs dl 0.0.21 or newer.
+wf reap            remove the workspaces whose work is over — a closed ticket,
+                   or an open one whose PR merged with nothing still in flight
+                   (-y to skip the prompt). Warns about the ones it suspects
+                   but will not delete: a ticket whose PRs all closed unmerged,
+                   or one nobody claimed that nothing came of. Keeps anything
+                   running or holding work that is not pushed anywhere; -f
+                   reaps the unpushed ones too, naming what it discards.
+                   Needs dl 0.0.21 or newer.
 
 The skills wf execs ship in this package, so they update with it. `install`
 links both agents' skills directories at copies kept beside them; every launch
@@ -224,11 +228,11 @@ fn run_skills(install: bool) -> Result<()> {
     Ok(())
 }
 
-/// `wf reap`: remove the workspaces whose tickets are closed.
+/// `wf reap`: remove the workspaces whose nodes the tracker calls finished.
 ///
 /// The division of labour is the one the launch already draws — `dl` owns the
 /// containers, `wf` owns the tickets — so this asks `dl` what exists, asks the
-/// tracker which of those nodes are closed, prints the plan, and hands the
+/// tracker what has become of those nodes, prints the plan, and hands the
 /// finished ones back to `dl`. No terminal is taken: this is a stream command
 /// like `wf skills`, not a second TUI.
 ///
@@ -236,6 +240,12 @@ fn run_skills(install: bool) -> Result<()> {
 /// because a workspace someone expected to go and that stayed is the thing they
 /// most need told about, and a reason they disagree with ("still running" when
 /// they thought they had stopped it) is only actionable while no is an answer.
+///
+/// `warn` rows are the same argument pointed the other way: workspaces `wf`
+/// suspects are dead weight on evidence too weak to act on — a superseded
+/// ticket, or a node nothing has come of. They are printed and never counted
+/// into the prompt, because the only safe thing to do with a suspicion is say
+/// it out loud.
 async fn run_reap(yes: bool, insist: bool) -> Result<()> {
     use std::collections::BTreeSet;
     use std::fmt::Write;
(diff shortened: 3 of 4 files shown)


blooop (commented):

Reviewed at head `417624dd244ac36851963bff5f779a8afa79b59f`, merge-base with `main` `8ff6b6ebfc6fcd74b2d78da9b4f3fd44a72813f6`, three-dot diff. Two axes were run as independent subagents in fresh context and are reported **unmerged and unranked against each other** — a change can pass one and fail the other.

Spec for this node is the resolutions on #127 and #128 plus #129's amended scope note (the amendment lives in #129's breadcrumb trail).

### Preflight — verified in this session

| check | command | result |
| --- | --- | --- |
| merge-base resolves | `git merge-base origin/main HEAD` | `8ff6b6e` |
| diff non-empty | `git diff 8ff6b6e...HEAD --stat` | 4 files, +986/−121 |
| real CI runs on head SHA | `gh api "repos/blooop/wayfinder/actions/runs?head_sha=417624d" --jq .total_count` | `2` (CI, Package — both `completed`/`success`, event `pull_request`) |
| all checks green | `gh pr checks 130` | 4 rows, none failing |
| formatting | `cargo fmt --all --check` | clean |
| lints | `RUSTFLAGS=-D warnings cargo clippy --all-targets --all-features --locked` | clean |
| CI test selection | `cargo test --locked --lib --bins --examples` | **291 passed, 0 failed** (lib 281, bins 10) |
| doc guard | `cargo test --locked --test skill_docs` | **2 passed, 0 failed** |

The PR body's "310 passed" is consistent with what ran: 291 above + 19 integration tests (`--list` counts: live_discovery 4, live_fetch 4, live_launch_exec 1, live_streaming_startup 8, skill_docs 2) = 310. **The 17 `live_*` tests were listed, not run** — they hit the real tracker, and one green run of them would not be proof anyway. `cargo doc` was **not run** here; CI's Docs step covers it.

"21 new tests in `src/reap.rs`" checks out by count: `#[test]` count goes 11 → 32 between `8ff6b6e` and head.

---

## Standards

The seam judgement first: exporting **`parse_pr`** is right — `PrOutcome::project` matches `Option<&PrStatus>` exhaustively, so a new `PrStatus` variant is a compile error in both readers, and that is a real invariant the widening buys. Exporting `PrNode`/`Nodes`/`GraphQlResponse` is defensible for the envelope. What leaks is that the *type* moved to `pub(crate)` and the **query text that must satisfy it did not**.

- **should fix** — The PR selection is duplicated verbatim: `src/fetch.rs:56-62` ≙ `src/reap.rs:479-485` (plus `assignees(first: 5) { nodes { login } }`). `PrNode`'s `state`/`is_draft`/`repository` are non-`default` serde fields now demanded by two hand-written strings; adding a field to `PrNode` breaks `reap` at runtime only. Shotgun Surgery. Fix: a `pub(crate) const` selection in `fetch.rs` beside `PrNode`, interpolated by both queries.
- **should fix** — Nothing tests either query against the types it feeds. I gutted `NODE_FACT_SELECTION` (`src/reap.rs:477`) — dropped issue `state` and every PR field but `number` — and all 281 `--lib` tests stayed green. `the_batch_asks_one_question_per_node…` (`src/reap.rs:1301`) asserts only two substrings; `BATCH_RESPONSE` (`src/reap.rs:1240`) is hand-written and drifts independently of the query.
- **should fix** — `node_fact_is_total_over_the_rollup_and_the_claim` (`src/reap.rs:663`) is not total over what its own doc claims. `node_fact`'s doc (`src/reap.rs:186-190`) asserts "a merge outranks the closed PRs beside it", but no case pairs `Merged` with `ClosedUnmerged`. I swapped the two blocks in `node_fact` (`src/reap.rs:198-220`) so `Superseded` wins — a real regression (a merged ticket would `warn` instead of `reap`) — and 281/281 stayed green.
- **should fix** — `README.md:656` lists "tickets someone has claimed" under **Kept, always**. `NodeFact::Claimed` requires *no PRs*; a claimed ticket whose PR merged is `DoneByMerge` and is deleted. The rest of the README (row wording, keep/warn/reap order, one-call-per-repo) matches the code exactly. No `skills/*` mentions reap, so the docs/skills sync rule is not engaged.
- **nit** — `src/main.rs:271-281` re-switches over `Verdict` to a `&str` and string-compares against a literal array. A `Verdict::label()` beside `id()`/`reason()` collapses it; today a fourth arm compiles the inner match but silently prints nothing if the array is missed.
- **nit** — 20 full-`reason`-string equality assertions in `src/reap.rs:601-1358`. Most are testing *which arm*, not the wording; a wording tweak touches 20 sites.

**Axis judgement:** structurally sound and unusually well-argued in its doc comments, but the visibility widening stopped one step short — the schema-coupled query string stayed behind, untested, and the totality test does not cover the precedence it advertises.

**Executed (Standards):** all in a private worktree at `417624d`, removed after. `cargo test --locked --lib` → `281 passed; 0 failed` (baseline); same after swapping `Superseded`/`DoneByMerge` precedence in `node_fact` → `281 passed; 0 failed`, and `cargo test --locked --lib reap::` → `32 passed; 0 failed; 249 filtered out` (**same verdict under both selections** — not an isolation artefact); same after gutting `NODE_FACT_SELECTION` → `281 passed; 0 failed`. Both mutations reverted via `git checkout -- src/reap.rs`.

**Not verified (Standards):** `cargo fmt`/clippy/`--bins --examples`/`skill_docs` (cited from the preflight table above, not re-run); `cargo doc`; any live `gh`/`dl` behaviour; the README's rendered example against a real `wf reap` run.

---

## Spec

Conformance against #127/#128 resolutions + #129's amended scope note.

- **Empty prewarm diff — holds.** `git diff --stat 8ff6b6e...417624d` = README.md, src/fetch.rs, src/main.rs, src/reap.rs only. Spec: *"The diff to `src/launch.rs` and `src/app.rs` is **empty**."* — satisfied. (ok)
- **`doomed` is the single deletion definition and genuinely constrains.** `src/reap.rs:110`; `main.rs:284` asks for it. Mutation-tested: adding `| Verdict::Warn { .. }` to the filter turned **CI-effective red** under the full `--lib --bins --examples` selection — 279 passed / 2 failed (`a_warned_workspace_never_reaches_the_doomed_set_even_with_y_and_f` `src/reap.rs:954` and `all_prs_closed_unmerged_warns_...` `src/reap.rs:894`). Not an isolation-only red. Spec: *"the headline safety property: #128 adds zero new deletion paths."* (ok)
- **18/18 test cases covered; 0 missing.** #127 1–8 → `src/reap.rs:858, 691(table), 894, 1039, 1069, 1088, 830, 1278/1316/1340`. #128 1–10 → `916, 937, 954, 1020, 993, 877, 664, 1316, 1304, 830/858/894`. Caveat: #127-2/#128-6 (*"beside a merged sibling"*) is pinned only in the totality table (`src/reap.rs:691`); the plan-level `an_open_or_draft_pr_still_keeps_and_never_warns` (`src/reap.rs:877`) loops PR *numbers* 40/42, not PR *shapes*, so its name overstates it. (nit)
- **Unknown *issue* state deletes — the amendment's asymmetry.** `src/reap.rs:197` `if !is_open { return NodeFact::Closed }` over `is_open(s) = s == "OPEN"` (`src/fetch.rs:211`). Probe fixture `"state": "SOMETHING_NEW"` through `parse_node_facts`+`plan`: `fact=Closed verdict=[Reap{reason:"devlaunch#80 is closed"}] doomed=[Reap{...}]`. `main`'s `issue_is_closed` was `== "closed"` — a positive test that *kept*. Against `src/reap.rs:141`: *"a state added after this binary shipped costs a workspace that stays, never one that goes."* Fix is local: a three-arm issue projection mirroring `PrOutcome::project`. Note the spec's letter is on the PR's side (#128: *"1. `!is_open` → `Closed`"*). (should fix)
- **`PrOutcome` total; `Unstarted` unreachable from failure.** `src/reap.rs:147` maps `None`→`InFlight`; `src/reap.rs:533` errors on missing alias, null alias, null repository, and `errors[]`; `src/reap.rs:330` refuses an unknown node. (ok)
- **Warn wording + forbidden words — exact.** `src/reap.rs:357/362` end `reap by hand if so{discard}`; no `discarding` when quiet. #128: *"Three things the reason line must **not** say."* (ok)
- **Derivation order verbatim**, incl. `InFlight{min}` / `DoneByMerge{min merged}` / `Superseded{max}` — `src/reap.rs:196-224`, pinned at `src/reap.rs:732`. **Guard order** unsaved→Running→fact intact, `src/reap.rs:317-327`. **1 GraphQL call/repo, 0 per workspace** — `src/reap.rs:429-443` loops repos, not nodes. (ok)
- **Timing reproduced** (8 real wayfinder issues, 3 runs, warm auth): batched **0.46/0.44/0.47 s** vs REST loop **2.63/2.54/2.70 s** ≈ 5.7x — corroborates the PR's 0.44/0.41/0.44 vs 2.52/2.58/2.52 within network variance, and the PR correctly scopes it to `gh` invocation time, not `wf reap`. (ok)

**Axis judgement:** conformant to both resolutions and the amendment; one should-fix (unknown issue state → Reap) that the resolutions themselves carry.

**Executed (Spec):** `git diff --stat 8ff6b6e...417624d`; `git show 8ff6b6e:src/reap.rs` (old `== "closed"`); `cargo test --locked --lib reap::` → 32 passed/0 failed; mutated `doomed` + `cargo test --locked --lib --bins --examples` → **279 passed/2 failed** (both doomed-set tests); reverted + same selection → 281 + 10 + 0 passed, 0 failed; probe test on `"state":"SOMETHING_NEW"` → `fact=Closed … doomed=[Reap]`; `/usr/bin/time gh api graphql …` ×3 → 0.46/0.44/0.47 s; `/usr/bin/time bash -c 'for n in 129 126 124 121 117 90 59 47; do gh api repos/blooop/wayfinder/issues/$n --jq .state; done'` ×3 → 2.63/2.54/2.70 s. All mutations were in a private worktree, now removed.

**Not verified (Spec):** no `live_*` test was run; `cargo fmt`/clippy taken from the preflight table, not re-run; end-to-end `wf reap` timing not measured (the PR says so too); the printer's grouped output was read, not executed; single-run timings carry network variance and were not statistically controlled.

---

## Verdict — Approve

**Zero blocking findings.** GitHub refuses `--approve` on a same-account PR, so this is posted with `--comment`; per `wf-review`, the written verdict above is the gate, not the GitHub review state.

What the approval rests on, all executed in this session or by an axis subagent with output pasted above: the prewarm-path diff really is empty; `reap::doomed` really is the single deletion definition and its guard test goes **red under the CI-effective selection** when mutated (279/2), not merely in isolation; all 18 named cases from #127 and #128 map to real tests; `PrOutcome` is total and `Unstarted` is unreachable from a parse failure, a null alias, a null repository, or a GraphQL error; the warn wording matches #129's amendment exactly and avoids #128's three forbidden phrases; and the batched-call timing reproduces independently at ~5.7x, correctly scoped by the PR body to `gh` invocation time rather than end-to-end `wf reap`.

Six should-fixes and three nits are recorded above and none of them block. The two worth doing soonest, as follow-ups rather than on this branch:

1. **`src/reap.rs:197` — unknown issue state lands on the deleting arm.** `main` kept in that case; this PR reaps. It is what #128's derivation order literally says, and GitHub's `IssueState` is `OPEN|CLOSED` today, so this is a hazard rather than a live bug — but it is the mirror image of the defect #129's amendment exists to kill, pointed at the one arm that removes ~1 GB. Cheap to close with a three-arm issue projection beside `PrOutcome::project`.
2. **Two real regressions passed the suite under mutation** — swapping `DoneByMerge`/`Superseded` precedence, and gutting `NODE_FACT_SELECTION` — both 281/281 green. The behaviour in the diff is correct; it is the pinning that is thin, and the second one means no test binds either query string to the types it must satisfy.

Neither is a spec violation and neither changes what this branch does today, which is why the verdict is Approve and not Request changes.

PR #131, opened 2026-08-09
Hand the launched agent its ticket context

### LINKED ISSUE #124: Hand the launched agent its ticket context

blooop:
Closes #124. Implements the shape #122's resolution settled: the launch context rides the opening prompt as a `ctx:` block between the skill's own arguments and any `steer:` suffix.

A launched skill's first tracker call can now be the **claim** rather than a rediscovery of the map, the ticket and its PRs.

```
/wf-review 124 ctx: {"v":1,"repo":"blooop/wayfinder","map":{…},"aim":{"ticket":{…,"prs":[…]}}} steer: <text>
```

## The numbers

**Before** — re-measured for this revision (mean of 5, this machine, this session; the earlier revision quoted figures inherited from a discarded attempt and framed them as measured, which they were not). The discovery prelude a build launch runs today is 3 serial `gh` calls:

| call | command timed | mean of 5 |
| --- | --- | --- |
| parent map + labels | `gh api repos/blooop/wayfinder/issues/124` | **402 ms** |
| linked PRs | `gh issue view 124 --json closedByPullRequestsReferences` | **423 ms** |
| map title | `gh issue view 121 --json title` | **347 ms** |

**~1.17 s** for a build launch. A review launch handed a bare ticket number pays **612 ms** more to find its PR (`gh pr list --search 124 --json number,headRefName,statusCheckRollup,reviewDecision`), so **~1.78 s**. These are network round trips from one machine on one day — the point is the order of magnitude and the fact that they are serial, not the third digit.

**After**: those calls are gone, not made faster — every fact they fetched is in the prompt, because `wf` had already fetched it to draw the row you picked. The launch itself makes no extra call: the block is serialized from data already in memory. Cost is prompt bytes — **310 bytes** for this repo's test fixture with one linked PR, **381** for this very ticket's node (arithmetic on the same shape with its real titles), against #122's estimated ~1.5 KB worst case and three orders of magnitude under any argv limit.

What is *not* eliminated: ticket bodies, comments and map bodies. `wf` never fetches them (there is no `body` in the GraphQL selection), so the schema does not pretend to carry them.

## Shape

The schema is the launch's **own** types serialized directly, not a mirror pair beside them — the aim sum, the launchable stage and the PR link each carry their own snake-case wire names, and golden literals in the tests pin the format so a rename fails. Departures from #122's literal Rust sketch, all deliberate:

- **No `CtxAim`/`CtxPr`.** They would be the same sums written twice and free to drift; one type serialized once cannot disagree with itself.
- **The map reference carries its repo**, not just its number: a ticket can sit on a map in another repo, and a bare number would point a cross-repo launch at whatever issue holds that number in the ticket's own repo. #122 gave the PR link a repo for exactly this reason.
- **The snapshot instant is dropped.** No consumer in the contract reads it, the staleness guard is the live claim rather than a timestamp comparison, and a wall-clock read would make prompt building non-deterministic.
- **The staged stop's separate title is deleted rather than duplicated** — the title the picker draws and the title the agent is handed are one fact, so it is derived from the aim.
- **§4's "serializer round-trip … property-tested" is an enumeration, not a round trip.** There is no round trip to test: the context types are `Serialize` only, deliberately — `wf` writes the block and never reads one, and adding `Deserialize` would ship a parser with no caller. And the vocabularies are small and closed (6 types × 4 stages; 3 PR states plus 4×4 open ones), so the whole matrix is enumerated exhaustively against golden literals, which pins *which* word the wire uses where a property run only pins that some round trip holds. This departure was missing from this list in the first revision; the matrix itself was missing too, and is now `every_stage_type_and_pr_state_spells_itself_on_the_wire`.

Two things are unrepresentable rather than checked for: **the claim** (no assignee, no ticket status — so a stale block cannot make an agent act on someone else's ticket; the worst it can do is misdescribe a PR for a few orientation tokens) and **a done stage** (`Launchable`, not `Stage`).

## The seams, and what was red

Both were pre-agreed by #122's resolution; no test was written at an unconfirmed seam.

1. **The prompt-building seam** — the argv a launch execs, asserted as whole literal strings so the block's *position* is pinned, not merely its presence. **Genuinely red first**: the assertion failed with the prompt as the bare skill invocation (`/wf-tdd 16`), not a compile error.
2. **The container seam** — the single shell command the `dl` form hands `devpod ssh --command`, parsed back by a **real POSIX shell** rather than a hand-written inverse of `shell_quote`, and compared against what the *same node launched on the host* hands its agent, so it cannot pass by both sides being wrong together. The fixture title carries a single quote, a command substitution and a double quote. **This one was green on its first run** — the quoting already held; what was missing was any proof from the far side of the seam. Checked non-vacuous rather than assumed: with the quoting removed the recovered command comes back as nine arguments with the JSON's quotes eaten and the substitution executed.

**The fallback is asserted negatively**: plain sessions and all three creation launches carry no block, and no test anywhere makes ctx a precondition for a skill to proceed. A hand-invoked skill finds nothing and discovers exactly as before.

## Consumers

One shared section in `skills/wf/GITHUB_TRACKER.md` — the binding wording from #122 verbatim, the schema with its worked example, the per-field vocabularies and the omissions — and one grammar line in each of the five skills, each pointing at that section rather than restating it. `wf-one` says a creation carries no block and may not synthesize one for its subagents. The README's two verbatim launch lines are updated.

`tests/skill_docs.rs` (from #125) is extended deliberately rather than worked around: it builds real launches through the binary's own path and asserts the docs describe the block the code writes — **as JSON, field for field and word for word**, not merely as a shape.

## What changed after the first review

- **`pwned` is gone from the branch, ignored, and the canary is scoped.** The empty file at the repo root was residue of the container-seam test's own `$(touch pwned)` fixture firing during a `shell_quote` mutation. The canary is worth keeping — it is the only evidence that the far side of the seam did *not* execute anything — so the recovering shell now runs in a scratch directory of its own and that directory being empty afterwards is an assertion in its own right, where before nothing looked at the canary at all. `.gitignore` carries `/pwned` for anything run from an older tree.
- **The docs guard now binds the schema, not just its shape.** It pinned the block's prefix and version; the doc's field names and vocabularies could be rewritten to nonsense with the suite green. Two checks replace that: the doc's worked example is built as a real launch of the node it describes and compared as parsed JSON, and every enumerated field's whole vocabulary is compared against what the types serialize. The doc comment that claimed the two "cannot drift" is rewritten to say what is actually enforced.
- **#122 §4's matrix exists** — see the departures list above for what it is and is not.
- **The claim guard no longer collides with a legal value.** It forbade the substring `"needs"`, which `"stage":"needs_attention"` contains; it is now the block's exact set of *field names*, so a value cannot trip it and any new field fails it.
- **`/wf-one` on a node emits no block**, matching `wf-one`'s own doc. Unreachable from the picker, but `plan` takes any route.
- **The before-numbers were re-measured**, with the commands stated.

## Checks

Full chain green locally on this head, with `RUSTFLAGS=-D warnings RUSTDOCFLAGS=-D warnings`: `fmt --all --check`, `clippy --all-targets --all-features --locked`, `test --locked --lib --bins --examples` (**290 / 10 / 0** passed — `main`'s #130 merge brought the lib count up from 269), `test --locked --test skill_docs` (**7** passed), `doc --no-deps --all-features --locked`.

`tests/live_launch_exec.rs` learned the block (it splits the context off before its numeric checks and asserts the version, the repo anchor and the absence of any assignee) but was **not run** — it needs a real pty, a live tracker and an authenticated `gh`, and is excluded from the CI chain by design. It is updated, not verified. Partial mitigation: `ci.yml` builds the live tests (`cargo test --locked --all-targets --no-run`), so the edit is at least type-checked on every push.

One acceptance check from #122 is left for a human: launch a ticket through `wf` and confirm the session's first tracker call is the claim.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,12 @@
 # build output. Only the worktree directory is ignored: `.claude/` also holds
 # settings and skills, which are worth committing.
 /.claude/worktrees/
+
+# The container-seam test's canary (#124). Its fixture title carries
+# `$(touch pwned)` so that a *broken* `shell_quote` is caught by the file
+# appearing rather than only by an argv mismatch — the substitution firing is
+# the whole evidence. The test now runs its shell in a scratch directory so the
+# canary lands there, but a mutation experiment run from an older tree (or from
+# a shell that inherits the crate root) still drops one here, and this file was
+# once committed to a public repo that way.
+/pwned
--- a/README.md
+++ b/README.md
@@ -435,6 +435,22 @@ through `codex --dangerously-bypass-approvals-and-sandbox`. Both receive one
 prompt argument, so the route, numbers, and steering text cannot be split apart
 by the shell.
 
+**Every launch of a node also hands the agent what `wf` already knew**, as a
+`ctx: <json>` block between the skill's arguments and any steering suffix:
+
+```
+/wf-review 124 ctx: {"v":1,"repo":"blooop/wayfinder","map":{…},"aim":{"ticket":{…,"prs":[…]}}} steer: <text>
+```
+
+That is the parent map, the ticket's type and stage, and its linked PRs — the
+three serial `gh` calls a launched skill used to open with, answered before it
+starts. It is an accelerator and never a precondition: a skill invoked by hand
+never went through the picker, finds no block, and discovers exactly as before.
+The one thing it deliberately cannot say is whether the ticket is still yours to
+take — there is no assignee and no ticket status in the schema, so claiming stays
+a live call and a stale block cannot make an agent act on someone else's work.
+The creation rows carry no block at all: they name nothing that exists yet.
+
 The auto mode collapses the ticket rows on purpose: the launched session is a
 *manager*, and what it manages is the node's whole remaining lifecycle —
 `wf-tdd`, the gate, then a fresh-context `wf-review` — so it is the manager
@@ -494,7 +510,7 @@ A checkout that carries a **`.devcontainer/devcontainer.json`** (or a top-level
 [`dl`](https://github.com/blooop/devlaunch):
 
 ```
-dl owner/repo@wayfinder/repo-80 -- 'claude' '--dangerously-skip-permissions' '/wf 67 80'
+dl owner/repo@wayfinder/repo-80 -- 'claude' '--dangerously-skip-permissions' '/wf 67 80 ctx: {"v":1,…}'
 ```
 
 Codex deliberately stays on the host even for such a checkout. `dl` mounts
--- a/skills/wf-auto/SKILL.md
+++ b/skills/wf-auto/SKILL.md
@@ -65,6 +65,8 @@ Blocking is the tracker's native dependency edge, so the frontier renders in the
 
 **Not yet specified** is in-scope fog you can't yet phrase sharply; **Out of scope** is work past the destination. Ticket it when the question is already sharp, even if blocked and unactionable; leave it as fog when it isn't.
 
+A launch from `wf` reads `<sigil>wf-auto <map> [<ticket>] ctx: <json> [steer: <text>]`. The block is the snapshot `wf` already held — the map's title, and for a ticket its type, stage and linked PRs — so a run can open on the node it was handed without rediscovering it. It is an **accelerator, never a precondition**, it never survives into a subagent's own invocation unexamined, and every gate below is still checked from live tracker state. See **The launch context** in [GITHUB_TRACKER.md](../wf/GITHUB_TRACKER.md).
+
 Tracker mechanics — pinning the repo, creating and parenting issues, blocking edges, the frontier query, the local-markdown fallback — live in [GITHUB_TRACKER.md](../wf/GITHUB_TRACKER.md) in the sibling `wf` skill. Pin the repo from the working repo's own remote, pass it explicitly on every call, and push branches by name: a fork's map belongs to the fork, never its parent. State the repo and its visibility in your first line of output — an autonomous run on a public repo publishes the whole fog sketch, and the invocation is the only consent there is for that.
 
 ## A run is a manager
--- a/skills/wf-one/SKILL.md
+++ b/skills/wf-one/SKILL.md
@@ -15,6 +15,8 @@ Two issues, not one: wf's cluster header is a map and headers are not cursor sto
 
 ## Set it up
 
+A `wf-one` launch is a **creation**: it names work that does not exist on the tracker yet, so its line is `<sigil>wf-one <task>` and never carries a `ctx: <json>` block. Nor may you synthesize one when handing the ticket to `wf-tdd` or `wf-review` below — the block is `wf`'s record of what it fetched, and a hand-written one would be a claim about tracker state nobody read. Those subagents discover from the ticket, exactly as a hand-typed invocation does. See **The launch context** in [GITHUB_TRACKER.md](../wf/GITHUB_TRACKER.md).
+
 Tracker mechanics — pinning the repo, creating and parenting issues, labels, claiming — are in [GITHUB_TRACKER.md](../wf/GITHUB_TRACKER.md) in the sibling `wf` skill. Pin the repo from the working repo's own remote and pass it explicitly on every `gh` call; push branches by name.
 
 1. **The map** (`wayfinder:map`), titled as the work. Body is short, because a one-ticket map has no route to index:
--- a/skills/wf-review/SKILL.md
+++ b/skills/wf-review/SKILL.md
@@ -15,7 +15,9 @@ Adapted from aihero.dev's `/code-review` (see `research/aihero-spine.md` in bloo
 
 ## Invocation
 
-`wf-review <pr>` — a PR number or URL (or a ticket number whose linked open PR is unambiguous). Resolve the repo explicitly and pass `--repo` on every `gh` call.
+`wf-review <pr> [ctx: <json>] [steer: <text>]` — a PR number or URL (or a ticket number whose linked open PR is unambiguous). Resolve the repo explicitly and pass `--repo` on every `gh` call.
+
+A launch from `wf` adds a `ctx: <json>` block after the number, and it carries the ticket's **linked PRs** — which is precisely the lookup this skill otherwise opens with when it is handed a bare ticket number. It is an **accelerator, never a precondition**: with no block, or one you cannot read, resolve the PR the way a hand-typed invocation does. See **The launch context** in [GITHUB_TRACKER.md](../wf/GITHUB_TRACKER.md).
 
 ## Inputs — deterministic, never asked
 
--- a/skills/wf-tdd/SKILL.md
+++ b/skills/wf-tdd/SKILL.md
@@ -15,7 +15,9 @@ Adapted from aihero.dev's `/tdd` + `/implement` (see `research/aihero-spine.md` 
 
 ## Invocation
 
-`wf-tdd <ticket>` — an issue number or URL. Resolve the repo explicitly from the working checkout's remote and pass `--repo` on every `gh` call.
+`wf-tdd <ticket> [ctx: <json>] [steer: <text>]` — an issue number or URL. Resolve the repo explicitly from the working checkout's remote and pass `--repo` on every `gh` call.
+
+A launch from `wf` adds a `ctx: <json>` block after the ticket number: the parent map, the ticket's type and stage, and its linked PRs, all already fetched — so the map lookup this stage opens with is a read you can skip. It is an **accelerator, never a precondition**; a hand-typed invocation carries none and discovers exactly as it always has, and the claim below is a live call either way. See **The launch context** in [GITHUB_TRACKER.md](../wf/GITHUB_TRACKER.md).
 
 ## The contract is the ticket
 
--- a/skills/wf/GITHUB_TRACKER.md
+++ b/skills/wf/GITHUB_TRACKER.md
@@ -41,6 +41,47 @@ If `origin` is the **parent** — you cloned upstream and added your fork as a s
 gh api "repos/$REPO/issues/NUMBER" --jq .id
 ```
 
+## The launch context (`ctx:`)
+
+A launch line may carry `ctx: <json>` — a snapshot of what `wf` knew at exec time. It is an accelerator, never a precondition: use it to skip discovery reads (which map, which PR, what type and stage); never let it substitute for a live read before any write — claiming, commenting, closing, gating. If it is absent, does not parse, has a `v` you don't recognise, or names a repo or ticket other than the one your arguments and pinned `$REPO` name, ignore it entirely and discover via the commands below, as a hand-invoked session always does.
+
+It sits between the skill's own arguments and any `steer:` suffix, on one line:
+
+```
+<sigil><skill> <args> ctx: {"v":1,…} steer: <text>
+```
+
+Everything after `steer:` is still the human's text, so a steer that happens to contain the letters `ctx:` is steering text and nothing else.
+
+**What it carries** (schema `v: 1`; a ticket aim shown — a whole-map launch reads `"aim": "map"` and names no ticket):
+
+```json
+{
+  "v": 1,
+  "repo": "owner/name",
+  "map": { "repo": "owner/name", "number": 121, "title": "the map's title" },
+  "aim": {
+    "ticket": {
+      "number": 124,
+      "title": "the ticket's title",
+      "ticket_type": "build",
+      "stage": "in_review",
+      "prs": [{ "repo": "owner/name", "number": 130, "status": { "open": { "checks": "passing", "review": "required" } } }]
+    }
+  }
+}
+```
+
+The map carries its own repo because a ticket can sit on a map in another repo. `ticket_type` is one of `build` · `research` · `task` · `grilling` · `prototype` · `untyped`; `stage` one of `ready` · `building` · `in_review` · `needs_attention` — never `done`, because a finished node is not launchable. A PR's `status` is `draft`, `merged`, `closed`, or `open` with its `checks` (`absent` · `pending` · `passing` · `failing`) and `review` (`not_required` · `required` · `approved` · `changes_requested`).
+
+**What it deliberately does not carry**, and why the omissions are the contract rather than an oversight:
+
+- **No assignee and no ticket status.** The one fact whose staleness is dangerous — is this ticket still mine to take — is not in the schema at all, so everything present *is* safe to orient from. Claim-first is unchanged: `gh issue edit <n> --repo "$REPO" --add-assignee @me` is still the first thing a session does, and a ticket taken or closed between the pick and the launch fails there, before any work.
+- **No blockers.** The picker refuses a blocked node before staging, so a launched ticket has none open.
+- **No bodies and no comments** — of the ticket, of the map, of anything. `wf` never fetches them, so the reads that need them are unchanged.
+
+**Precedence: your arguments beat the block, and the tracker beats both.** The ticket number in your invocation is the assignment; a block naming a different number or repo is discarded whole, never merged field by field.
+
 ## Labels (once per repo, idempotent)
 
 ```bash
--- a/skills/wf/GITHUB_TRACKER.md
+++ b/skills/wf/GITHUB_TRACKER.md
@@ -72,7 +72,18 @@ Everything after `steer:` is still the human's text, so a steer that happens to 
 }
 ```
 
-The map carries its own repo because a ticket can sit on a map in another repo. `ticket_type` is one of `build` · `research` · `task` · `grilling` · `prototype` · `untyped`; `stage` one of `ready` · `building` · `in_review` · `needs_attention` — never `done`, because a finished node is not launchable. A PR's `status` is `draft`, `merged`, `closed`, or `open` with its `checks` (`absent` · `pending` · `passing` · `failing`) and `review` (`not_required` · `required` · `approved` · `changes_requested`).
+The map carries its own repo because a ticket can sit on a map in another repo.
+
+Every enumerated field, with its whole vocabulary — one line each, and each line is the complete list:
+
+- `aim` — `map` · `ticket`
+- `ticket_type` — `build` · `research` · `task` · `grilling` · `prototype` · `untyped`
+- `stage` — `ready` · `building` · `in_review` · `needs_attention`
+- `status` — `draft` · `open` · `merged` · `closed`
+- `checks` — `absent` · `pending` · `passing` · `failing`
+- `review` — `not_required` · `required` · `approved` · `changes_requested`
+
+A `stage` is never `done`: a finished node is not launchable, so a block cannot claim one. `checks` and `review` exist only inside an `open` status — they are questions about something still in flight.
 
 **What it deliberately does not carry**, and why the omissions are the contract rather than an oversight:
 
--- a/skills/wf/SKILL.md
+++ b/skills/wf/SKILL.md
@@ -134,6 +134,7 @@ Claude and `$` for Codex):
 
 - `<sigil>wf <map> [<ticket>]` — everything below as written
 - `<sigil>wf <map> <ticket> steer: <steering>` — with a steering prompt
+- `<sigil>wf <map> [<ticket>] ctx: <json> [steer: <steering>]` — the same, with the snapshot `wf` already held: the map's title, and for a ticket its type, stage and linked PRs. An **accelerator, never a precondition** — orient from it, verify live before any write, and ignore it entirely if it is absent or disagrees with your arguments. See **The launch context** in [GITHUB_TRACKER.md](GITHUB_TRACKER.md).
 
 ### Chart the map
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -12,7 +12,7 @@ use std::collections::{BTreeMap, BTreeSet};
 
 use ratatui::crossterm::event::{KeyCode, KeyEvent, KeyModifiers};
 
-use crate::launch::{self, Agent, Candidate, Launch, LaunchMode, Route, Staged, Targets};
+use crate::launch::{self, Agent, Candidate, Launch, LaunchMode, MapRef, Route, Staged, Targets};
 use crate::model::{stage, Activity, Map, MapId, MapSet, Status, Ticket};
 use crate::projects::Checkout;
 use crate::refresh::Startup;
@@ -36,7 +36,12 @@ pub enum Outcome {
     Refresh,
     /// Run this ticket's agent. The last thing `wf` ever does: the loop
     /// returns, the terminal is restored, and the process becomes the agent.
-    Launch(Launch),
+    ///
+    /// Boxed because it is by far the largest thing an outcome can carry — a
+    /// launch holds the whole snapshot it hands its agent (#124) — and every
+    /// other outcome is a keystroke's worth of nothing. One allocation on the
+    /// last keypress of the session buys a small `Outcome` for all the rest.
+    Launch(Box<Launch>),
 }
 
 /// A modal layer over the main screen: the staged second step of a launch
@@ -680,8 +685,7 @@ impl App {
                 Outcome::Continue
             }
             Some(Stop::Map(id)) => {
-                let title = self.clusters[&id].title.clone();
-                let staged = Staged::map(&id, &title);
+                let staged = Staged::map(&MapRef::new(&id, &self.clusters[&id].title));
                 self.prewarm(&staged);
                 self.overlay = Overlay::PickLaunch {
                     candidate: staged.default_candidate(),
@@ -724,7 +728,11 @@ impl App {
             ));
             return Outcome::Continue;
         }
-        match Staged::ticket(ticket, row.map.number, stage(&ticket.prs, &ticket.status)) {
+        // The map the row was picked in, carried whole: a ticket can sit on a
+        // map in another repo, and a bare number would name the wrong issue
+        // there (#124).
+        let map = MapRef::new(&row.map, &self.clusters[&row.map].title);
+        match Staged::ticket(ticket, &map, stage(&ticket.prs, &ticket.status)) {
             None => {
                 self.notice = Some(format!("#{} is done — nothing to launch", ticket.number));
                 Outcome::Continue
@@ -824,7 +832,7 @@ impl App {
             }
             Targets::One(launch) => {
                 self.notice = Some(format!("→ {}", launch.describe()));
-                Outcome::Launch(launch)
+                Outcome::Launch(Box::new(launch))
             }
             Targets::Many(launches) => {
                 self.notice = Some(format!("{repo}{key}: which checkout?"));
@@ -935,7 +943,7 @@ impl App {
                     .nth(cursor)
                     .expect("picker cursor stays in range");
                 self.notice = Some(format!("→ {}", launch.describe()));
-                Outcome::Launch(launch)
+                Outcome::Launch(Box::new(launch))
             }
             KeyCode::Down | KeyCode::Char('j') => {
                 self.overlay = Overlay::PickCheckout {
@@ -1895,7 +1903,11 @@ mod tests {
                     "a task is a decision node"
                 );
                 assert_eq!(staged.key(), "#6");
-                assert_eq!(staged.title, "Re-entry breadcrumbs", "the picker names it");
+                assert_eq!(
+                    staged.title(),
+                    "Re-entry breadcrumbs",
+                    "the picker names it"
+                );
                 assert_eq!(
                     *candidate,
                     Candidate::Launch {
@@ -1916,7 +1928,10 @@ mod tests {
         assert_eq!(launch.key(), "wayfinder#6");
         assert_eq!(launch.cwd(), std::path::Path::new("/data/proj/wayfinder"));
         // The map issue is the cluster's — not a per-repo lookup.
-        assert_eq!(launch.agent_argv().last().unwrap(), "/wf 1 6");
+        assert_eq!(
+            launch::eliding_ctx(launch.agent_argv().last().unwrap()),
+            "/wf 1 6 ctx: …"
+        );
         assert_eq!(app.overlay, Overlay::None, "one candidate must not prompt");
         assert!(app
             .notice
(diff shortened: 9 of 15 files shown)


blooop (commented):

Two-axis review per `wf-review`, run in a fresh context that never saw this code written. The axes ran as parallel subagents in separate working copies and their findings are **not merged or reranked** — each section is its axis's own.

**Fixed point**: `git diff 8ff6b6ebfc6fcd74b2d78da9b4f3fd44a72813f6...HEAD` (three-dot merge-base with `main`). **Spec**: the resolution comment on #122, plus #124's own trail for the four shaping decisions settled before this build.

**Preflight**: merge-base resolves, diff non-empty (15 files, +861/−139), and **2 real workflow runs exist on `a04c324`** (`CI` and `Package`, both `success`) — confirmed via `gh api .../actions/runs?head_sha=…`, not inferred from a quiet checks list.

---

## Standards

**Verified by execution** (full CI chain, `RUSTFLAGS=-D warnings RUSTDOCFLAGS=-D warnings`): `fmt --check` exit 0; `clippy --all-targets --all-features --locked` exit 0; `cargo test --lib --bins --examples` → **268 / 10 / 0 passed**; `cargo test --test skill_docs` → **5 passed**; `cargo doc --no-deps` exit 0. Builder's counts confirmed exactly. Mutation experiments were run in a private copy, not the shared worktree.

1. **blocking** — `pwned` (repo root, from b3b839f) is committed to a public repo, and `.gitignore` has no entry. Verified: deleted it, ran `launch::tests::the_containers_own_shell_hands_the_agent_what_the_host_would` (single-test selection) → passed, file absent. Then mutated `shell_quote` (src/launch.rs:997) to weak double-quoting → same test failed **and recreated `pwned`**. So the file is residue of the builder's own red run, and the canary writes into the crate root on every future red run.
2. **blocking** — the drift guard does not cover the schema. `tests/skill_docs.rs:139-161` asserts only the prefix spelling and that the section contains a backticked `v`. Verified: rewrote `GITHUB_TRACKER.md`'s example to `"v": 7` / `"repository"` / `"type":"BUILD"` / `"stage":"IN-REVIEW"` plus the stage enumeration line → `cargo test --test skill_docs` **5 passed**. Nothing else reads the docs, so the schema the five skills parse can drift silently — while `tests/skill_docs.rs:105` claims "the docs and the code cannot drift". (Prefix drift *is* caught, both directions: code-side `ctx:`→`context:` → 1 failed; doc-side `ctx: <json>`→`ctx: <JSON>` in wf-tdd → 1 failed. Schema is accurate *today* — inspected serde attrs vs golden literals.)
3. **non-blocking** — `src/launch.rs:1190` `(Route::One, _) => skill` then unconditionally appends `ctx:` at :1198, so a `Job::Node` routed `Route::One` emits `/wf-one … ctx: {…}`, contradicting `skills/wf-one/SKILL.md`. Unreachable from the picker but representable: `plan()` (src/launch.rs:1535) accepts any `Route`.
4. **non-blocking** — boxing is lint-threshold-driven, not design-driven. Measured `size_of::<Launch>() = 216`, `size_of::<Targets>() = 216`: `Outcome::Launch`/`Ending::Handover` are boxed, `Targets::One(Launch)` (src/launch.rs:1523) is not, and `src/app.rs:835` boxes what `Targets::One` just handed over unboxed. Boxing is the right response; the split representation and the decision-flavoured comment at `src/app.rs:39` are the wart.
5. **non-blocking** — Divergent Change: `src/launch.rs` is 3011 lines and now also owns the wire schema alongside routing, checkout resolution, shell quoting and the `dl` seam.
6. **non-blocking** — `src/launch.rs:2130` `elided()` is a Middle Man over `eliding_ctx`. Eliding does *not* weaken the argv tests (inspected): block bytes stay pinned by golden literals at :2212/:2270/:2302/:2942/:2972.

**Inspected only** (not executed): docs-vs-code schema equivalence, README wording, cohesion judgements.

**Axis verdict:** structurally sound and unusually well-modelled (borrowed `LaunchContext`, no parallel wire types), but not mergeable until the committed `pwned` artifact is removed/ignored and the drift guard either covers the schema or its doc comment stops claiming it does.

---

## Spec

All mutation experiments were run in a private copy; the shared worktree was left untouched.

**Execution-verified (commands run this session):**
- Baseline: `cargo test --lib` → `268 passed; 0 failed`; `cargo test --test skill_docs` → `5 passed`. Matches the PR body.
- **Slice-1 red-first claim holds.** Replacing `Some(format!("{invocation} ctx: {ctx}"))` (`src/launch.rs:1205`) with `Some(invocation)` and running the single test gave an *assertion* failure, `left: "/wf-tdd 16"` — exactly the builder's claim. Selection: single test, then **full `--lib` suite → 246 passed / 22 failed**, so the guard is CI-effective, not red-only-in-isolation.
- **Slice-2 mutation proof reproduces.** `shell_quote` → identity (`src/launch.rs:997`) made the container-seam test fail with the recovered command as nine arguments after `claude`, JSON quotes eaten (`{v:1,repo:blooop/wayfinder,…}`), and `$(touch pwned)` executed. A verification slice proved non-vacuous is a legitimate substitute here.
- **Binding fallback wording is verbatim**: `diff` of #122's quoted paragraph vs `skills/wf/GITHUB_TRACKER.md:46` → identical.
- Cost claim **310 bytes** for the one-linked-PR fixture: recomputed, exact.

**Findings**
1. **Blocking** — a stray empty file `pwned` is committed at repo root (added in b3b839f, in the diff as `pwned | 0`): the artifact of the mutation proof's `$(touch pwned)` fixture firing. Not spec'd, public repo.
2. **Blocking** — #122 §4: *"the serializer round-trip is property-tested across the `CtxAim` × `PrStatus` matrix"*. No round-trip exists (all ctx types are `Serialize`-only) and no test touches `PrStatus::Draft/Merged/Closed`, `Checks::Absent/Pending/Failing`, `Review::NotRequired/ChangesRequested` — so the vocabularies `GITHUB_TRACKER.md:75` documents are unpinned. This departure is absent from the PR body's "deliberate departures" list.
3. **Non-blocking** — #122 §5: *"`tests/live_launch_exec.rs` … pass unchanged with no ctx present"*. It now hard-requires ctx (`tests/live_launch_exec.rs:406`), as does `tests/skill_docs.rs:151`. Both assert wf's *emission*, not skill behaviour, so the fallback path is not gated — but the named file did not stay unchanged, and the assertion is dead (never run).
4. **Non-blocking** — the claim-free guard forbids the substring `"needs"` (`src/launch.rs:2332`), which matches the spec-mandated `"stage":"needs_attention"` (#122: *"`stage` is `Launchable`, not `Stage`"*). Verified by execution: swapping the fixture to `Stage::NeedsAttention` fails the guard. Fixture-bound blacklist, not a structural check.
5. **Non-blocking** — #124: *"Measure it… the before number has to exist before the after number means anything."* The 444/343/351/599 ms figures are quoted verbatim from a **discarded earlier attempt's** breadcrumb, whose own handoff said "re-measure … if anything about the launch path has moved" — it moved. The PR body's "(mean of 5, measured against this tracker)" does not disclose the inheritance. The *after* is honestly framed as structural, and the 310/381-byte cost is correctly labelled arithmetic.
6. **Non-blocking** — `skills/wf-auto/SKILL.md:68` and `GITHUB_TRACKER.md:51` write `ctx: <json>` unbracketed where #122's grammar is `[ctx: <one-line JSON>]`; the other three skills bracket it.

**Inspection-only (not executed):** `tests/live_launch_exec.rs` — the edit reads correctly (splits ctx before the numeric checks, asserts `v`, repo anchor, no assignee), but nothing has ever run it; a test file changed with zero execution behind it. Disclosure is clear in both the PR body and the #124 breadcrumb.

**Also by inspection + one execution:** the schema (`src/launch.rs:566-644`) carries no assignee/status/blocked_by/body, `aim` is the existing sum serialized directly, `Aim::Map` → `"aim":"map"`, `Launchable` not `Stage`, map ref carries its repo. The emitted JSON obtained from a live test run matches `GITHUB_TRACKER.md:57-75` field-for-field and vocabulary-for-vocabulary — but `skill_docs.rs` only checks the doc mentions `` `ctx: <json>` ``/`` `v` `` and that the block starts `{"v":1,`, so that agreement is unguarded against future drift.

**Axis verdict:** implements #122's shape faithfully — grammar, position, claim-freeness and the verbatim fallback all verified by execution — but ships one committed junk file and one silently-dropped spec test requirement; **request changes**.

---

## Reviewer's own execution notes

Run by the review coordinator independently of both axes, in a third private copy with its own target dir, so the headline claims are not taken on any subagent's word:

- `cargo fmt --all --check` → clean; `cargo clippy --all-targets --all-features --locked` → clean; `cargo doc --no-deps --all-features --locked` → clean, all under `RUSTFLAGS=-D warnings RUSTDOCFLAGS=-D warnings`.
- `cargo test --locked --lib --bins --examples` → `268 passed` / `10 passed` / `0 passed`. `cargo test --locked --test skill_docs` → `5 passed`. **The builder's stated counts are exact.**
- The `shell_quote`→identity mutation reproduces verbatim: the container-seam test fails, the recovered argv is the command plus nine arguments, the JSON's quotes are stripped (`{v:1,repo:blooop/wayfinder,…}`), the fixture title comes back as `don't  "x"` with the substitution consumed, and `pwned` appears in the crate root. Reverting `src/launch.rs` restores green and the file is **not** recreated. This independently confirms both that the container-seam test genuinely bites and that `pwned` is residue of that experiment.
- `git status --porcelain` is clean after a full passing suite run, so the committed `pwned` is not produced by any green path.
- Mitigation worth recording against the "edited but never run" concern: `ci.yml` compiles the live tests (`cargo test --locked --all-targets --no-run`), so the `tests/live_launch_exec.rs` edit is at least type-checked by CI even though it is never executed.

Note on the docs guard, offered as context rather than as a finding of either axis: `every_bundled_skill_documents_the_context_in_its_grammar` is satisfied for `wf-one` by a sentence that says it *never* carries a `ctx: <json>` block — the substring check passes on a negation, which is weaker than the test's own name.

---

## Verdict

**Request changes.** (Posted as a comment: GitHub refuses `--request-changes` on a same-account PR. The written verdict is the gate.)

Both axes independently reached the same verdict. This is good work — the shape is faithful to #122, the modelling is genuinely constructive (one set of types serialized once, the claim absent from the type rather than checked for), the seams are real, and the builder's honesty about slice 2 being green-first held up under reproduction. Three blocking items stand between it and merge:

1. **`pwned` is committed at the repo root** (Standards 1, Spec 1) — an empty file, residue of the mutation experiment, in a public repo with no `.gitignore` entry. Delete it; and give the canary a target outside the crate root (or a `TempDir`) so a future red run does not litter the tree again.
2. **The drift guard does not guard the schema** (Standards 2) — `GITHUB_TRACKER.md`'s field names and value vocabularies can be rewritten to nonsense with `skill_docs` still green, while `tests/skill_docs.rs:105` states the opposite. Either assert the documented example against the emitted JSON, or correct the doc comment's claim. Docs/code drift is this change's most likely long-term failure and it is currently unguarded.
3. **#122 §4's `CtxAim` × `PrStatus` matrix test is absent and undisclosed** (Spec 2) — every `PrStatus`, `Checks` and `Review` spelling documented in `GITHUB_TRACKER.md:75` is unpinned by any test; only the empty-`prs` path is covered. Either add the matrix coverage or record it in the PR body as a fifth deliberate departure with its reason.

Worth fixing while in there, none of them blocking: the `"needs"` blacklist collides with the legal `"stage":"needs_attention"` (Spec 4); `Route::One` on a `Job::Node` emits a block `wf-one`'s doc forbids (Standards 3); and the PR body should say plainly that the before-numbers are **inherited from a discarded earlier attempt**, not measured in this build (Spec 5) — everything else in that section is scrupulously labelled, which is why this one omission stands out.

blooop (commented):

**Re-verdict after the retry.** Two-axis review per `wf-review`, run in a fresh context that never saw this code written and never saw the previous review written. The axes ran as parallel subagents in separate working copies and their findings are **not merged or reranked** — each section is its axis's own. The coordinator ran its own experiments in a third copy.

**Fixed point**: `git diff 8f7f5810bab4492230e3392dd5c1036898d0bc8b...HEAD` (three-dot merge-base with `main`; the merge-base moved because the branch merged `main`/#130). **Spec**: the resolution comment on #122, plus #124's own body.

**Preflight**: merge-base resolves; diff non-empty (15 files, +1377/−139); **2 real workflow runs exist on `8736b9e`** (`gh api .../actions/runs?head_sha=…` → `total_count 2`), all checks `SUCCESS`, `mergeStateStatus CLEAN`. `git ls-tree -r HEAD` contains no `pwned`.

### Status of the three findings from [the previous review](https://github.com/blooop/wayfinder/pull/131#pullrequestreview-4891194404)

| # | finding | status |
| --- | --- | --- |
| 1 | `pwned` committed at repo root | **closed** |
| 2 | the docs guard did not bind the schema | **closed** |
| 3 | #122 §4's matrix test absent and undisclosed | **closed** |

Each was re-verified by mutation in three independent working copies (both axes and the coordinator), not accepted on the PR body's evidence.

---

## Standards

**Verified by execution** (`RUSTFLAGS=-D warnings RUSTDOCFLAGS=-D warnings`): `cargo fmt --all --check` exit 0; `cargo clippy --all-targets --all-features --locked` exit 0; `cargo test --locked --lib --bins --examples` → **290 / 10 / 0**; `cargo test --locked --test skill_docs` → **7 passed**; `cargo doc --no-deps --all-features --locked` exit 0; `cargo test --locked --all-targets --no-run` exit 0. Builder's counts match exactly.

- **Mutation A** (`shell_quote` → `arg.to_string()`), selection = single test `launch::tests::the_containers_own_shell_hands_the_agent_what_the_host_would`: **RED** — `panicked at src/launch.rs:3189: the shell executed something the quoting should have made inert, leaving ["pwned"] behind`. Afterwards `git status --porcelain` showed only ` M src/launch.rs`; no `pwned` at repo root; no stray `/tmp/wf-seam-*`. **The scoping holds.**
- **Mutation B** (`Launchable` `rename_all` → `SCREAMING-KEBAB-CASE`): selection = file `--test skill_docs` → **5 passed / 2 failed**; selection = single test `every_stage_type_and_pr_state_spells_itself_on_the_wire` → **RED**.
- **Mutation C** (this axis's own, not the builder's): added `Launchable::Spiking`, fixed only the two compile errors (`stage_word`, `route`) — selection = that single matrix test → **green**; selection = file `--test skill_docs` → **7 passed**.

**Findings**

1. **non-blocking** — `src/launch.rs:2422`. The `*_word` doc comment claims a new variant "stops this module compiling until the new word has been decided **and pinned here**". Mutation C disproves the second half: iteration comes from hand-written arrays (`src/launch.rs:2493`, `:2501`), so a new variant is *named* but never launched; `tests/skill_docs.rs:285-323` uses arrays too, so its published vocabulary would be silently incomplete. The house pattern `Route::all()` / `Mode::all()` / `CreationKind::all()` already exists and would close it. The mirror otherwise earns its size (verified biting in B).
2. **non-blocking** — `README.md:438`. "**Every** launch of a node also hands the agent what `wf` already knew" is false for plain mode, which `src/launch.rs:1174` and `nothing_that_has_no_skill_to_address_is_handed_context` both exclude; `README.md:461` then says plain leaves "everything else unchanged". Otherwise **the merged README reads as one document**: the `ctx:` paragraph sits correctly after the prompt-argument paragraph, the `dl` example at `:513` matches it, and no `reap` text is contradicted. (The orphaned `~/.claude` paragraph at `:692` is pre-existing on `8f7f581`, not this merge.)
3. **non-blocking** — `tests/live_launch_exec.rs:418` substring-scans the whole block for `"assignee"`/`"claim"` — the exact anti-pattern retry item 4 removed from the unit test. A live ticket titled e.g. "claim the frontier" fails it.
4. **non-blocking** — prior items re-derived. **Middle Man** `src/launch.rs:2138` `elided()` is still pure delegation to `eliding_ctx` — *unchanged*. **Boxing** `src/app.rs:44` now carries a design rationale and the box travels unbroken to `Ending::Handover` — *better*, but sibling `Targets::One(Launch)` (`src/launch.rs:1528`) has the same shape unboxed and unexplained. **Divergent Change**: `src/launch.rs` 2545→3294 lines, but production is only +150 (context schema, `MapRef`); tests are +598 / 9 tests, well-organized behind a coherent helper vocabulary — *unchanged, not unmaintainable*.

**Judgements requested.** The no-wildcard `*_word` mirror is a good shape, **not** a Repeated Switch — exhaustive matches are the documented house pattern (`src/model.rs:213`) and hand literals are right for a golden; only the iteration half is weak (finding 1). The exact-key-set claim guard is **right, not over-constraining**: the block is a *published* schema (`GITHUB_TRACKER.md`), so a new field should require an explicit edit, and under Principle 3 the invariant is already type-enforced — this is legitimate wire defence-in-depth.

**Re-derived vs accepted**: re-derived **by execution** — canary scoping (incl. reasoning about `cd`, absolute-path, panic and concurrency escapes), the docs guard code side, the matrix test. Re-derived **by inspection** — the `Route::One` behaviour, the claim-guard shape, the whole merged README read end to end, and all three prior non-blocking findings. **Correction to the retry's own list**: the `--no-run` mitigation is *not this PR's* — the `.github/` diff against the merge-base is empty and `cargo test --locked --all-targets --no-run` landed in `8ff6b6e` (#125), already on `main`. The mitigation is real; it is inherited, not added here.

**Inspected only (not executed)**: `skills/*/SKILL.md` and `GITHUB_TRACKER.md` prose (Spec axis owns it); `tests/live_*.rs` runtime behaviour (compiled, never run).

**Axis verdict: Approve.**

---

## Spec

**Execution-verified**, all in a private worktree at `8736b9e`, `git status --porcelain` empty before and after every mutation.

| selection | result |
| --- | --- |
| `cargo test --locked --lib` | `ok. 290 passed; 0 failed` |
| `cargo test --locked --test skill_docs` | `ok. 7 passed; 0 failed` |
| doc `"v":1`→`7` → `--test skill_docs` | `FAILED. 6 passed; 1 failed` |
| doc **one** field `repo`→`repository` | `FAILED. 6/1` |
| doc vocab `build`→`BUILD` | `FAILED. 6/1` |
| doc vocab `in_review`→`IN-REVIEW` | `FAILED. 6/1` |
| doc: **delete the whole `checks` vocabulary line** | `FAILED. 6/1` |
| doc: extra `assignee` field in the example | `FAILED. 6/1` |
| doc value `in_review`→`ready` | `FAILED. 6/1` |
| code `Launchable` → `SCREAMING-KEBAB-CASE` → `--test skill_docs` | `FAILED. 5 passed; 2 failed` |
| …same, single test `every_stage_type_and_pr_state_…` | `FAILED. 0 passed; 1 failed` |
| code: extra `assignee` in the **emitted** block → single test `the_handed_context_cannot_speak_about_the_claim` | `FAILED. 0/1` |
| code: golden `stage_word` `in_review`→`in-review` → single matrix test | `FAILED` |
| code: golden `status_words` `draft`→`DRAFT` → single matrix test | `FAILED. 0/1` |
| `shell_quote` → identity, selection `--lib container` | `FAILED. 6 passed; 2 failed`, msg `…leaving ["pwned"] behind`; afterwards only ` M src/launch.rs`, no `pwned`, no `/tmp/wf-seam-*` |
| `gh issue view 121 --json title` ×5 | 0.33 / 0.31 / 0.35 / 0.32 / 0.39 s |
| restore → `--lib`, `--test skill_docs` | `290 passed`, `7 passed`, tree clean |

Counted in code: the matrix walks 6 types × 4 stages = **24**, and 3 + (4×4) = **19** PR states, plus the map arm (`src/launch.rs:2492-2545`). The fallback wording at `GITHUB_TRACKER.md:44` is **byte-verbatim** from #122.

**Prior blocking findings** — 1 **closed** (`git ls-tree -r HEAD` has no `pwned`; `.gitignore:15` carries `/pwned`; the canary was *strengthened*: a broken `shell_quote` now fails on the canary's presence inside the scratch dir, satisfying #122 §3's *"container-seam survival test"*, and writes nothing outside it). 2 **closed** (seven independent mutations bite in both directions, three of them never demoed by the builder). 3 **closed** as to existence and disclosure.

**Ruling on the §4 round-trip departure.** #122 §4: *"`v == 1` is pinned literally, and the serializer round-trip is property-tested across the `CtxAim` × `PrStatus` matrix."* **Accept.** The types are `Serialize`-only, so a round trip would require shipping a `Deserialize` with no caller, and a property run would prove serde agrees with itself rather than *which word crosses the wire*. Exhaustive enumeration over a closed 24 + 19 vocabulary against golden literals is strictly stronger here — and a single-cell literal change is caught (rows C3/C4 above). The disclosure is honest: departure #5 states plainly that *"the matrix itself was missing too"* in the first revision.

**Findings**

1. **blocking (disclosure only)** — #122 §5: *"the existing skill-invocation tests (`tests/live_launch_exec.rs` and the SKILL.md grammars) **pass unchanged with no ctx present**."* `tests/live_launch_exec.rs:404-406` now hard-requires it (`.split_once(" ctx: ").unwrap_or_else(|| panic!(…))`). §5's control — a test that never mentions ctx — is gone, and this departure is absent from the PR body's *departures* list (it appears only as "learned the block" under Checks). Substantively the guarantee survives via `src/launch.rs:2571` and the doc contract. One-line fix: a sixth departures entry.
2. **non-blocking** — #122's *"`ctx` before `steer`"* is pinned on the code side (`src/launch.rs:2257`) but not on the doc side: reordering `GITHUB_TRACKER.md`'s grammar line to `steer: <text> ctx: {…}` left `--test skill_docs` at `7 passed`.

**Docs guard — catches vs misses.** *Catches*: the `v` value; any single field rename on either side; any value respelling; an extra field on either side; any vocabulary word respelled; a whole vocabulary line deleted; a `rename_all` change; a field-label rename. *Misses* (all confirmed `7 passed`): the grammar-line ordering; the prose *"A `stage` is never `done`"* inverted; the *"No assignee and no ticket status"* bullet inverted; **the precedence sentence inverted** (*"argv beats ctx"* → *"the block beats your arguments"*); the map-aim prose sentence respelled. The schema is bound; the *contract prose* around it is not.

**Other spec obligations.** §1 satisfied — whole-string equality (`src/launch.rs:2261`) pins exactly one block *and its position*; the claim whitelist bites. §2 satisfied — `"aim":"map"` at `:2618` / `:3253`; `Route::Plain`, all three creations and `(Route::One, _)` assert `None` at `:2571`, matching `skills/wf-one/SKILL.md:18`. §6 disclosure adequate. **Precedence** is expressed at `GITHUB_TRACKER.md:98`, which all five SKILL.mds link to and `skill_docs` requires them to link to — reachable where skills actually read. **#124 "Measure it" — provenance now honest**: the body states outright that the earlier figures were *"inherited from a discarded attempt and framed as measured, which they were not"*, and five independent timings of the map-title call (mean **0.34 s**) match its claimed **347 ms** within noise.

**Inspection-only (not executed)**: `tests/live_launch_exec.rs` (never run by anyone); `ci.yml` job behaviour; §6's transcript acceptance check; clippy/fmt/doc; the scratch-dir code-path analysis (Standards axis owns it).

**Axis verdict: Request changes** — 1/2/3 genuinely closed and the §4 argument accepted, but the §5 departure is undisclosed.

---

## Reviewer's own execution notes

Run by the coordinator in a third private worktree, independently of both axes, so no headline claim rests on a subagent's word.

- Full chain under `RUSTFLAGS=-D warnings RUSTDOCFLAGS=-D warnings`: `fmt --all --check` exit 0; `clippy --all-targets --all-features --locked` exit 0; `doc --no-deps --all-features --locked` exit 0; `test --locked --lib --bins --examples` → `290 passed` / `10 passed` / `0 passed`; `test --locked --test skill_docs` → `7 passed`. **The builder's stated counts are exact.**
- **Both docs-guard directions reproduced personally.** Doc side, selection = `cargo test --locked --test skill_docs` with `"v": 7` / `"repository"` / `"BUILD"` / `"IN-REVIEW"`: `test result: FAILED. 5 passed; 2 failed`, with `left: {"BUILDING", "IN-REVIEW", "NEEDS-ATTENTION", "READY"}` / `right: {"building", "in_review", "needs_attention", "ready"}`. Restored → `7 passed`. Code side, `Launchable` `rename_all` → `SCREAMING-KEBAB-CASE`, same selection: `FAILED. 5 passed; 2 failed`; single test `launch::tests::every_stage_type_and_pr_state_spells_itself_on_the_wire` → `FAILED. 0 passed; 1 failed; … 289 filtered out`.
- **Canary reproduced personally.** `shell_quote` → `arg.to_string()`, selection = the single container-seam test: `panicked at src/launch.rs:3189: the shell executed something the quoting should have made inert, leaving ["pwned"] behind`. Note the *new* assertion fires **before** the argv comparison — the canary is what catches it. `ls pwned` in the repo root → `No such file or directory`; `git status --porcelain` → only ` M src/launch.rs`; `ls -d /tmp/wf-seam-*` → none. **The temp-dir scoping holds, and it cleans up on the red path** (`remove_dir_all` runs before the emptiness assertion).
- **One residual on the canary, non-blocking.** The `assert!(out.status.success(), …)` at `src/launch.rs:3179` fires *before* `read_dir`/`remove_dir_all`, so a shell that refuses the command leaks its scratch dir. Verified by execution (script mutated to `touch escaped-on-failure; exit 1`): `/tmp/wf-seam-2158734-0/` survived, containing `escaped-on-failure`. It leaks into `$TMPDIR`, never the repo — which is exactly the property finding 1 demanded — so this is a tidiness note, not a regression.
- **The claim guard bites**, verified personally: adding `assignee: "someone-else"` to `LaunchContext` and re-emitting → single test `the_handed_context_cannot_speak_about_the_claim` `FAILED`, `left` carrying `"assignee"` against a `right` without it. Adding the field to `Aim::Ticket` instead is a **compile error at five sites** — the invariant is type-shaped first and wire-checked second.
- **The `wf-one` negation loophole is closed**, verified personally: flipping its line to "sometimes carries" → `every_bundled_skill_documents_the_context_in_its_grammar` `FAILED` (`wf-one is a creation's skill: it must say it carries no block`); stripping `[ctx: <json>]` from `wf-tdd` → same test `FAILED` (`must show the launch context in its invocation grammar, not merely mention it in prose`). Selection for both: `--test skill_docs`, `6 passed; 1 failed`.
- Independently found the same two unguarded doc regions the Spec axis reports (grammar-line ordering, contract prose) and the same `README.md:438` overclaim the Standards axis reports — reached in separate copies, so they are three independent derivations, not one repeated.
- `.github/` diff against the merge-base is **empty**; `ci.yml:108`'s `--no-run` is inherited from #125. The PR body states it as a fact about the repo rather than as a change, which is accurate.
- Minor: `src/launch.rs:1201`'s `(Route::Plain, _) => return None` is unreachable — `route.bundled_skill()?` at `:1187` already returns early for `Plain`. Required for match exhaustiveness, but two places now encode the same fact.
- **Not verified by anyone**: `tests/live_launch_exec.rs` has still never been run (real pty + live GitHub + authenticated `gh`; excluded from CI by design), and #122 §6's transcript-level acceptance check is correctly left for a human. Both are disclosed in the PR body.

---

## Verdict

**Approve.** (Posted as a comment: GitHub refuses `--approve` on a same-account PR. The written verdict is the gate.)

**All three original blocking findings are closed**, and none of them on the builder's say-so — each guard was re-mutated in three separate working copies, and each bit:

1. **`pwned` — closed.** Gone from the tree, ignored with an explanatory comment, and the canary was **scoped rather than weakened**: it now asserts on evidence nothing looked at before, fires ahead of the argv comparison, and cleans up its own scratch on the red path.
2. **The docs guard — closed.** It binds the schema now. Between the three copies it was mutated more than a dozen ways, including a single field rename, a deleted vocabulary line, an added field and a code-side `rename_all` — every one red, restore green. This was the guard that was hollow on this very PR; it is not hollow now.
3. **#122 §4's matrix — closed.** It exists, it covers 24 type×stage cells and 19 PR states against golden literals, a one-character change to any literal turns it red, and the departure is disclosed and argued.

**On the §4 departure I rule with the Spec axis: accept.** There is no round trip to test without shipping a `Deserialize` that nothing calls, and exhaustive enumeration against literals answers the question the spec actually cared about — *which word* the wire uses — where a property run answers only that some inverse exists. Over a closed 24 + 19 vocabulary, enumeration is the stronger test, not the cheaper one.

**One axis-level disagreement, resolved here.** The Spec axis grades its §5 item blocking; I grade it **non-blocking** and record the disagreement rather than hide it. Finding 3 was blocking because a required test *did not exist* and nothing in the body said so. §5 is not that: the PR body states in two places that `live_launch_exec.rs` "learned the block" and that "no test anywhere makes ctx a precondition for a skill to proceed" — which is §5's actual binding clause, and it is true. What is missing is only the *label* "departure", on a file that has never been executed and whose assertion concerns `wf`'s emission rather than a skill's behaviour. That is a body edit, not a code gate. Adding the sixth departures entry is still worth doing.

Nothing new is blocking. Worth picking up next, in rough order of value:

- **The `*_word` iteration is hand-written arrays** (`src/launch.rs:2493`, `tests/skill_docs.rs:285-323`), so a new variant compiles once its word is pinned but is never launched and never required of the docs — the comment at `:2422` reads stronger than that. The repo's own `Route::all()` pattern closes it in a line per type. This is the same *class* of gap as finding 2, caught early only because the `*_word` compile error forces a human into that module first.
- `README.md:438`'s "**Every** launch of a node" excludes plain mode, contradicting `:422` and `:461` and the PR's own test at `src/launch.rs:2578`.
- The docs guard binds the schema but not the surrounding contract prose — notably the precedence sentence, which can be inverted with the suite green.
- `tests/live_launch_exec.rs:418`'s substring scan for `"assignee"`/`"claim"` is the anti-pattern this retry removed from the unit test.
- The scratch dir leaks into `$TMPDIR` on the shell-refusal path (`src/launch.rs:3179`).
- The sixth departures entry for §5.

This is a good retry. It fixed the three findings by making the guards *stronger* rather than by making the assertions quieter — which, given that three tests in this run were caught passing while guarding nothing, is the distinction that mattered most.

PR #149, opened 2026-08-13
Drop the allow for a lint that does not exist

blooop:
`clippy::duration_suboptimal_units` is not a clippy lint. On the toolchain this crate pins, every `cargo clippy` run printed:

```
warning[E0602]: unknown lint: `clippy::duration_suboptimal_units`
  = note: requested on the command line with `-A clippy::duration_suboptimal_units`
```

once per target — eleven lines of noise on a clean tree.

`cargo clippy -- -W help` on clippy 0.1.93 lists no such lint. The nearest names are `duration_subsec` and `unchecked_time_subtraction`, and neither does what the comment described.

So the `allow` silenced nothing, and the comment justified something that never happens — the `from_secs(60)` timeouts in `tests/live_launch_exec.rs` it claimed to protect were never at risk. AGENTS.md requires every `allow` to carry a comment saying why; this one had a comment, about a lint nobody has ever run.

**Deleted rather than renamed**, because there is no name to rename it to. If a future clippy adds the lint it will warn on its own, and can be silenced then under a name that exists.

### Why it survived

`-D warnings` never promoted it to an error: an unknown lint named on the command line is reported before lint levels apply. CI stayed green while printing it on every run, which is precisely the shape of thing that stops getting read.

### Verification

`cargo clippy --all-targets --all-features --locked` under `-D warnings` is now completely silent. Full gate green: fmt, clippy, 390 lib + 18 bin tests, 7 skill-doc tests, docs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,13 +53,6 @@ cast_possible_truncation = "allow"
 cast_possible_wrap = "allow"
 cast_sign_loss = "allow"
 
-# Wants `Duration::from_mins(1)` where a test writes `from_secs(60)`. The
-# timeouts in tests/live_launch_exec.rs are meant to be compared with each other
-# at a glance — `from_secs(60)` beside `from_secs(30)` says more than one of each
-# unit does — and the suggested constructor is newer than the toolchain the
-# recipe asks for anyway.
-duration_suboptimal_units = "allow"
-
 # `dbg!` survives a review far too easily.
 dbg_macro = "warn"
 


blooop:
Withdrawn — I had this backwards.

`clippy::duration_suboptimal_units` **does** exist, as of clippy 1.97.0, which is what CI's `stable` toolchain resolves to. This PR removed the allow and CI immediately failed with the very lint it claimed was fictional:

```
error: constructing a `Duration` using a smaller unit when a larger unit would be more readable
795 |         Duration::from_secs(60),
    = note: `-D clippy::duration-suboptimal-units` implied by `-D warnings`
```

So the `allow` and its comment were right, and `tests/live_launch_exec.rs:795` is exactly the site it was written for.

My evidence was `cargo clippy -- -W help` listing no such lint — true on **1.93.0**, which is what this machine has, and which I generalised into a claim about the lint itself. The real finding is the inverse: a local toolchain four releases behind CI prints `warning[E0602]: unknown lint` on every run for an allow that is doing its job. That is fixed by updating the local toolchain, not by editing Cargo.toml.

Issue #43, opened 2026-08-06
How the devcontainer image gets prebuilt in CI


blooop:
## Question

How does the devcontainer image get prebuilt and published, in a repo that deliberately has exactly one workflow?

`.github/workflows/package.yml` is documented as "one workflow for the whole life of the package" — there is a stated design preference against a second one. Prebuilding the devcontainer image is nonetheless in scope, so decide:

- A **second workflow** (`devcontainer.yml`) or a **new job in `package.yml`**? Weigh against the existing rationale for one workflow; the concurrency group and the `v*`-tag-only release gate are both `package.yml` conventions that a new job would inherit and a new workflow would have to restate.
- Trigger: on every push to `main`, only when `.devcontainer/**` changes, or on a schedule? A stale image is a slow `devpod up`, not a broken one.
- Tagging: `latest` only (python_template's choice), or content-addressed by Dockerfile hash so a workspace can pin?
- Registry and permissions: GHCR needs `packages: write`, and `package.yml` currently declares `contents: read` with a deliberately narrow release job. Say exactly which permission goes where.
- Whether the prebuild should also warm cargo (bake compiled deps into the published image) — this is where this ticket meets the cargo-warmth decision, and it may be the whole answer to the "fast pole".
- How `devcontainer.json` refers to the result, and how a contributor builds locally instead.

blooop:
## Resolution

**A third workflow, the simplest honest version of it — because the build cost is low and the guiding steer is: no machinery whose only job is optimizing a cheap thing.** (Grilled 2026-08-14; facts verified against the working tree and a live devpod.)

1. **Placement: a third workflow, `devcontainer.yml`.** The ticket's premise — "a repo that deliberately has exactly one workflow" — is stale: `ci.yml` exists now, and its header states the live convention ("Kept apart on purpose… the cheapest possible failure" should not be "the slowest one to hear about"). The image is its own concern with its own cost profile; it gets its own workflow, restating the few lines of concurrency/permissions boilerplate that costs.
[...]

Issue #38, opened 2026-08-06
Keeping cargo warm across container-per-branch


blooop:
## Question

What is the right way to keep cargo **warm across containers** when every branch is its own container?

The map settles that "fast" means a warm cargo cache rather than a warm container. Establish which mechanism actually delivers that, and which are traps:

- **Shared `CARGO_HOME` volume** (registry + git checkouts, ~566 MB here): is a docker volume safe to mount into several containers building concurrently? Cargo takes file locks — does that serialise builds or corrupt anything? Saves download only, not compile time.
- **Baking compiled deps into the image** (`cargo chef`, or a plain dep-only prebuild): can `target/` be seeded in the Dockerfile at a path cargo will actually reuse? Note cargo fingerprints are **path-sensitive**, so the baked path must match the workspace path the container ends up using — say what that path must be.
- **`CARGO_TARGET_DIR` outside the workspace mount**: does this help or hurt when the workspace itself is a fresh clone per branch?
- **`sccache` with a shared local disk cache**: is it path-independent enough to hit across branches, and what is the real hit rate for a dep set like this one (91 crates, tokio + ratatui)?
- Which of these compose, and which conflict.

Beware: `recipe/recipe.yaml` records "reusing a target dir saves nothing", but that is a rattler-build artefact (freshly timestamped `$SRC_DIR` each build). Do not treat it as evidence about devcontainers, which have a stable workspace path.

Answer with a ranked recommendation and the concrete config each option needs.

blooop:
## Answer

**Do nothing. Ship no cargo cache machinery at all.**

This ticket assumed keeping cargo warm across containers was worth engineering. #39 measured the premise away: a cold `cargo build --locked` with an *empty* registry and an *empty* target is **9 seconds**, and that 9 s includes fetching all 91 crates. Incremental rebuild after touching one file is 1 s.

Every option this ticket was asked to rank is therefore rejected on cost-benefit, not on mechanism:

- **Shared `CARGO_HOME` volume** — would save part of a 9 s build. This crate's own registry footprint is only 73 M (the host's 566 M is other projects). Rejected: adds a shared mutable volume and a concurrent-lock question to save single-digit seconds.
- **Baking compiled deps into the image (cargo-chef / dep-only prebuild)** — would work in principle (registry-sourced dep artifacts are path-stable even though the local crate's are not). Rejected: it would require widening the build context to the repo root, which is exactly what tripped devpod's 5000-file context-hash limit in #37.
[...]

PR #152, opened 2026-08-14
Boot the devcontainer from a published image, not a build

### LINKED ISSUE #150: Prebuild the devcontainer image: devcontainer.yml + cacheFrom, probe-first
Graduated from #43's resolution — build the decided shape, simplest version, probe before trusting.

## Spec (decided on #43, not re-openable here)

- **Probe first, against a real devpod:** does `devpod up` honor `build.cacheFrom` in devcontainer.json (layers pulled from the registry instead of rebuilt)? This is load-bearing and unverified. If devpod ignores it, fall back to `"image":` + a `.devcontainer/local/devcontainer.json` variant (`dl --devcontainer local` selects variants) and say so in the PR.
- **`.github/workflows/devcontainer.yml`** (third workflow — the one-workflow premise is stale per #43): triggers are push to `main` filtered on `.devcontainer/**`, plus `workflow_dispatch`. No PR leg, no schedule.
- Plain `docker build .devcontainer` + push — not `devpod build` (no features to bake). Registry `ghcr.io/blooop/wayfinder-devcontainer`, tag `latest` only, `GITHUB_TOKEN`.
- Permissions: workflow-level `contents: read`; `packages: write` on the build-push job only (package.yml's narrow-job pattern).
- `.devcontainer/devcontainer.json`: keep the `build:` block, add `cacheFrom` — and update the file's header comment, which currently says #43 is still open.
- Optional while there: check if claude's `install.sh` takes a version argument; pin via ARG if so, else leave the documented drift alone.
- **No cargo warming** (#38 reaffirmed on #43).

## Verification

A workspace booted via `dl blooop/wayfinder@<branch>` on a layer-cold host pulls layers from GHCR rather than rebuilding (or, on the fallback path, boots from `image:`). Before/after timing of that cold `devpod up` goes in the PR, per the map's measure-don't-assume rule.

blooop:
Closes #150.

## Which path, and why

**The fallback path — `image:` plus a `local` variant. devpod does not honor `build.cacheFrom`.**

That was the load-bearing unknown, so it was probed against a real devpod on a host with docker before anything was wired to it. Method: a throwaway registry, and a *separate* builder instance so the layer under test was genuinely absent from the builder devpod uses. A marker layer costing a fixed 25 s was pre-built in the isolated builder and pushed with inline cache metadata; then a workspace was brought up from a config declaring `cacheFrom` pointing at it.

- The layer rebuilt in full — 25.1 s, against 25.2 s for a control run with no `cacheFrom` at all.
- The stronger evidence is upstream of the timing: devpod logs the build command it shells out to, and with `cacheFrom` declared that command is byte-identical to the control. No cache flag is passed at all.
- Tried both spellings the devcontainer spec allows, single string and array. Both ignored, both rebuilt — so this is not a wrong-form mistake.
- The binary does contain the config accessor, so it parses the key and drops it before the build. Reading the schema would have said yes and been wrong, which is the whole reason #150 asked for a probe.

Recorded and deliberately **not** acted on: devpod has a native prebuild mechanism of its own (a repository flag it consults for prebuilt images). Out of scope here — noted so a future effort doesn't re-derive it.

## Measured, not assumed

Per the map's measure-don't-assume rule. Same host, base image present in both columns.

| | build from source (before) | prebuilt image (after) |
|---|---|---|
| image build | **34.9 s** (`--no-cache`) | **none** — zero build lines in the log |
| `devpod up` wall clock, image/layers present | 10.4 s | 10.6 s |
| cold-host download | 963.3 MB base + the tool tarballs fetched inside the build | 1087.3 MB total |

The download rows are the honest comparison, and they come from layer digests rather than a stopwatch on loopback: of the prebuilt image's 17 layers, **15 (963.3 MB) are digest-identical to the base image** the build path pulls anyway. This repo's own contribution is **2 layers, 123.9 MB**. So the prebuild is close to free in bytes — roughly the same base pull either way — and buys back the 35 s build on every cold workspace. A host that already has the image boots in 10.6 s having built nothing.

The `local` variant was booted too, not just asserted: devpod resolves its `../` paths, builds from the same Dockerfile, and the container comes up as `vscode` with `gh` 2.97.0, `zellij` 0.44.3 and the agent CLI at the newly pinned 2.1.223.

## Two operational consequences, flagged rather than assumed away

1. **A new GHCR package is private by default.** Until this one is made public, an anonymous `dl blooop/wayfinder@<branch>` cannot pull it and every default boot fails on authentication. Making the package public after the first publish (or having every user `docker login ghcr.io`) is a step outside this diff.
2. **The tag does not exist until this merges.** The merge commit touches `.devcontainer/`, so the push trigger publishes it automatically — but between merge and that run finishing, a default boot has nothing to pull. `--devcontainer local` is the way through that window, and `workflow_dispatch` re-fires the publish if the webhook is dropped.

## Shape

Simplest version throughout, as scoped: one workflow, one mutable `latest` tag, no pull-request leg and no schedule, plain `docker build` rather than `devpod build` (no features to bake), workflow-level `contents: read` with `packages: write` on the publishing job only, and no cargo warming (#38, reaffirmed on #43).

Seven offline assertions in `tests/devcontainer_prebuild.rs`, wired into `ci.yml` — this repo's rule is that an assertion no workflow runs is not an assertion. The one that earns its place most is the cross-file check that the workflow publishes *exactly* the reference the config boots from: renaming the registry in one place and not the other publishes an image nothing boots and leaves every workspace pulling a tag nothing publishes, and both halves look correct on their own. The drift guard between the two configs was green on arrival, so it was mutation-checked — perturbed until it failed, then restored — because a guard that has never been seen red is worth nothing.

The stale header comment claiming #43 is still open is replaced, and it now records the `cacheFrom` measurement so the next reader doesn't re-run the probe.

## Summary by Sourcery

Publish a prebuilt devcontainer image and wire configs, CI, and tests to ensure all references and triggers stay in sync.

New Features:
- Introduce a dedicated workflow to build and push the devcontainer image to GHCR on main devcontainer changes or manual dispatch.
- Add a local devcontainer variant that builds the image from the repository Dockerfile instead of pulling the published image.

Enhancements:
- Pin the agent CLI version in the devcontainer Dockerfile alongside existing `gh` and `zellij` pins to stabilize workspace startup state.
- Restrict devcontainer image publishing permissions to a single job and serialize runs with concurrency settings to avoid racing tag updates.

Tests:
- Add devcontainer prebuild contract tests that assert the default and local devcontainer configs, publishing workflow triggers, permissions, and image references remain consistent.
- Wire the devcontainer prebuild contract test binary into the main CI workflow so its assertions run on every commit.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -55,9 +55,20 @@ RUN set -eux; \
 # user that will run it. `~/.claude` is a bind mount from the host (see
 # devcontainer.json) and is deliberately *not* what this writes to — the binary
 # lives in the image, the credentials and settings come from the host at runtime.
+#
+# The pin is an argument to a third-party installer, so it is compared, not
+# trusted: `gh` and `zellij` above carry their versions in the download URL and
+# a wrong version cannot resolve, but an installer that silently ignored its
+# argument would leave this pin buying nothing with the build still green. The
+# comparison also backstops the unpiped `curl | bash` — a truncated script that
+# installed something else fails here.
 USER vscode
 RUN set -eux; \
     curl -fsSL https://claude.ai/install.sh | bash -s -- "$CLAUDE_VERSION"; \
-    /home/vscode/.local/bin/claude --version
+    installed="$(/home/vscode/.local/bin/claude --version)"; \
+    case "$installed" in \
+      *"$CLAUDE_VERSION"*) ;; \
+      *) echo "pinned CLAUDE_VERSION=$CLAUDE_VERSION but installed: $installed" >&2; exit 1 ;; \
+    esac
 ENV PATH=/home/vscode/.local/bin:$PATH
 USER root
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -16,9 +16,18 @@
 FROM mcr.microsoft.com/devcontainers/rust:1-bookworm
 
 # Pinned rather than "latest" so an image rebuild is a deliberate version bump
-# and not a silent one. Both are plain static binaries with no build step.
+# and not a silent one. `gh` and `zellij` are plain static binaries with no build
+# step; the agent CLI arrives through its own installer, which turns out to
+# accept a version (`install.sh [stable|latest|VERSION]`) — so it is pinned here
+# too rather than left to drift, which matters more now that this image is
+# published and shared than it did when every workspace built its own.
+#
+# It is a floor and not a ceiling: the agent CLI updates itself in place, so a
+# long-lived container will run something newer than this. What the pin buys is
+# that two workspaces created from the same image *start* from the same version.
 ARG GH_VERSION=2.97.0
 ARG ZELLIJ_VERSION=0.44.3
+ARG CLAUDE_VERSION=2.1.223
 
 # One layer, and the tarballs are deleted in it: an unpacked tarball left behind
 # in an earlier layer stays in the image however thoroughly a later layer removes
@@ -48,7 +57,7 @@ RUN set -eux; \
 # lives in the image, the credentials and settings come from the host at runtime.
 USER vscode
 RUN set -eux; \
-    curl -fsSL https://claude.ai/install.sh | bash; \
+    curl -fsSL https://claude.ai/install.sh | bash -s -- "$CLAUDE_VERSION"; \
     /home/vscode/.local/bin/claude --version
 ENV PATH=/home/vscode/.local/bin:$PATH
 USER root
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -19,9 +19,12 @@
   // spellings the spec allows (string and array) were ignored. So `cacheFrom`
   // here would look like a cache and never be one.
   //
-  // To build from source instead — a `Dockerfile` change, or a host whose
-  // architecture the published tag does not cover — use the variant beside this
-  // file: `dl blooop/wayfinder@<branch> --devcontainer local`. It carries the
+  // To build from source instead — a `Dockerfile` change, a host whose
+  // architecture the published tag does not cover, or the pull was denied
+  // (GHCR packages are private by default until made public, and the tag does
+  // not exist at all before `devcontainer.yml`'s first run on `main`) — use
+  // the variant beside this file:
+  // `dl blooop/wayfinder@<branch> --devcontainer local`. It carries the
   // `build:` block, and a test holds the two configs to the same settings so
   // this one cannot drift away from it.
   "image": "ghcr.io/blooop/wayfinder-devcontainer:latest",
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,23 +1,30 @@
 {
   "name": "wayfinder",
 
-  // Built from this repo's own Dockerfile rather than a prebuilt image. A cold
-  // `cargo build` here is 9 seconds (see blooop/wayfinder#39), so the thing a
-  // prebuilt image would save is the *image* build, not the crate build —
-  // whether that is worth a registry and a publishing workflow is
-  // blooop/wayfinder#43, still open. Until then, local build.
-  // `context` is this directory, NOT the repo root. The Dockerfile copies nothing
-  // out of the context — it only downloads pinned binaries — so a repo-root
-  // context would buy nothing and costs a great deal: devpod hashes the build
-  // context and gives up past 5000 files, and `target/` alone is far beyond that
-  // ("failed to compute context hash: exceeded limit of 5000 files"). If a future
-  // change needs Cargo.toml/Cargo.lock in the image (e.g. baking compiled deps),
-  // widen this to ".." and add a .dockerignore excluding target/ in the same
-  // commit.
-  "build": {
-    "dockerfile": "Dockerfile",
-    "context": "."
-  },
+  // A prebuilt image, not a build from `Dockerfile` — blooop/wayfinder#43 is
+  // settled and this is its answer (blooop/wayfinder#150). What a prebuild saves
+  // was never the crate build: a cold `cargo build` here is 9 seconds
+  // (blooop/wayfinder#39). It is the *image* build — the pinned `gh`, `zellij`
+  // and agent-CLI downloads in `Dockerfile`, which every branch's first `dl`
+  // paid for again. `devcontainer.yml` publishes this tag on any push to `main`
+  // that touches this directory.
+  //
+  // `image:` rather than the `build:` block plus `cacheFrom`, which is what #43
+  // expected, and the difference is a probed fact rather than a preference:
+  // **devpod does not honor `build.cacheFrom`**. Measured on 2026-08-14 against
+  // devpod v0.26.1 with a marker layer costing a fixed 25 s, pushed to a
+  // registry with inline cache metadata and absent from the builder's local
+  // cache — the layer rebuilt in full, and devpod's own log of the
+  // `docker buildx build` it shells out to carries no cache flag at all. Both
+  // spellings the spec allows (string and array) were ignored. So `cacheFrom`
+  // here would look like a cache and never be one.
+  //
+  // To build from source instead — a `Dockerfile` change, or a host whose
+  // architecture the published tag does not cover — use the variant beside this
+  // file: `dl blooop/wayfinder@<branch> --devcontainer local`. It carries the
+  // `build:` block, and a test holds the two configs to the same settings so
+  // this one cannot drift away from it.
+  "image": "ghcr.io/blooop/wayfinder-devcontainer:latest",
 
   // Note what is NOT here: `"runArgs": ["--network=host"]`. python_template sets
   // it, and copying it would defeat one of the three reasons this container
--- a/.devcontainer/local/devcontainer.json
+++ b/.devcontainer/local/devcontainer.json
@@ -1,0 +1,59 @@
+{
+  "name": "wayfinder (local build)",
+
+  // The build-from-source variant, selected by name:
+  //
+  //     dl blooop/wayfinder@<branch> --devcontainer local
+  //
+  // Reach for it when the published image is not what you want to run: you are
+  // changing `Dockerfile` itself and want the change before it is merged and
+  // published, or you are on an architecture the published tag does not carry.
+  // Everything else should use the default config beside this one, which pulls
+  // the prebuilt image instead (blooop/wayfinder#150).
+  //
+  // A separate file rather than `cacheFrom` on the default, because devpod does
+  // not honor `build.cacheFrom` — probed, not assumed; the default config's
+  // header records the measurement.
+  //
+  // Paths are relative to *this* file, so both point one level up: one
+  // `Dockerfile`, shared with the image the workflow publishes, so a local build
+  // and a published image can never be built from different instructions. Why
+  // the context is the parent and not the repo root is not stylistic — devpod
+  // hashes the build context and gives up past 5000 files, which `target/` alone
+  // exceeds. If a future change needs Cargo.toml/Cargo.lock in the image (e.g.
+  // baking compiled deps), widen it and add a .dockerignore excluding `target/`
+  // in the same commit.
+  "build": {
+    "dockerfile": "../Dockerfile",
+    "context": ".."
+  },
+
+  // Everything below is deliberately identical to the default config, and a test
+  // holds it that way (tests/devcontainer_prebuild.rs): these two files may
+  // differ in *how the image is obtained* and in nothing else, because a variant
+  // that quietly stopped mounting the agent's credentials would be a confusing
+  // way to find out. The reasoning for each setting lives in the default config
+  // and is not duplicated here — read it there.
+  "containerEnv": {
+    "CLAUDE_CONFIG_DIR": "/home/vscode/.claude"
+  },
+
+  "mounts": [
+    "source=${localEnv:HOME}/.claude,target=/home/vscode/.claude,type=bind"
+  ],
+
+  "initializeCommand": "mkdir -p \"$HOME/.claude\"",
+
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "rust-lang.rust-analyzer",
+        "tamasfe.even-better-toml",
+        "mhutchie.git-graph",
+        "anthropic.claude-code"
+      ]
+    }
+  },
+
+  "remoteUser": "vscode"
+}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,14 @@ jobs:
       - name: Skill doc shape checks
         run: cargo test --locked --test skill_docs
 
+      # The other config-as-behavior binary, same reasoning: what a workspace
+      # boots from and what `devcontainer.yml` publishes are promises split
+      # across two devcontainer configs and a workflow, and nothing but this
+      # notices when one of them moves. Named here for the same reason as the
+      # step above — the unit-test step stops at `--lib --bins --examples`.
+      - name: Devcontainer prebuild contract
+        run: cargo test --locked --test devcontainer_prebuild
+
       # The one part of `tests/` that needs neither network nor `gh`: the
       # fixture's own diagnostic, which decides what a failing live test tells
       # you about a missing GH_TOKEN. It is split from the environment it reads
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -1,0 +1,82 @@
+name: Devcontainer image
+
+# Publishes the image every `dl blooop/wayfinder@<branch>` boots from, so that
+# the pinned `gh`, `zellij` and agent-CLI downloads in `.devcontainer/Dockerfile`
+# are paid for once here instead of once per branch on every developer's machine.
+# The crate build was never the cost — a cold `cargo build` is 9 seconds
+# (blooop/wayfinder#39) — which is why this publishes a thin image and warms no
+# cargo cache at all (blooop/wayfinder#38, reaffirmed on #43).
+#
+# The fourth workflow, and separate for the same reason the other three are:
+# `ci.yml` answers "is this commit sound" with nothing but a compiler,
+# `package.yml` proves the recipe installs, `devlaunch-contract.yml` runs against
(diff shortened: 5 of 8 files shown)


blooop (commented):

Fixed point: `git diff a3903fa...HEAD` (merge-base with `main`), head `b81c0a6`. Spec: #150. Preflight green — all checks SUCCESS including `wf against a real dl`. The two axes below were produced by independent agents in separate contexts and are **not** merged or reranked: code can satisfy every standard while implementing the wrong thing, and vice versa.

## Standards

- **Blocking — the cross-file guard is blind in exactly the direction the PR body advertises.** `published_references` filters workflow tokens with `token.starts_with(repository)` (`tests/devcontainer_prebuild.rs:311`), where the prefix comes from the `PUBLISHED_IMAGE` literal. A push target renamed *away* from that prefix is never collected, so the set still equals `{booted}`. Mutation-checked both ways, and both stay **7/7 green**:
  - `docker push ghcr.io/blooop/wayfinder-image:latest` with the build line untouched → green. The workflow now publishes an image nothing boots.
  - The Push step deleted outright (`.github/workflows/devcontainer.yml:81-82`) → green. Nothing in the suite mentions `docker push` at all.

  This is the "renaming the registry, the repository or the tag in one place and not the other" case the doc comment at `:316-320` claims to catch, and the test the PR body singles out as earning its place most. The cross-*file* direction (config renamed) genuinely is caught; the within-workflow direction is not. The companion assertion meant to cover it — `assert!(workflow.contains("packages: write"), "naming the reference is not publishing it")` at `:336-339` — checks a permission block, not a push, and re-asserts what `:293-299` already checks exactly-once, so it cannot go red independently. The drift guard between the two configs *was* mutation-checked (per the PR body); this one was not.
- **Blocking — `AGENTS.md` now misdescribes the check chain.** The documented local command set (`AGENTS.md:110-116`) does not include `--test devcontainer_prebuild`, while `ci.yml:103-104` does, and `AGENTS.md:124-126` still says `tests/skill_docs.rs` "is the exception under `tests/`" that runs in the chain. An agent following the documented chain goes green locally and red in CI — the same drift class this diff otherwise works hard to prevent. Two-line fix.
- **Non-blocking — the secret is interpolated into the shell body**, `.github/workflows/devcontainer.yml:63-65`. `package.yml:154-156` and `:183-186` establish the house pattern of passing tokens through step `env:`, with a comment stating the reason ("so the token never appears in a command line or a process listing"). This workflow renders `${{ secrets.GITHUB_TOKEN }}` and `${{ github.actor }}` into the script instead. Not an injection vector — neither value is attacker-controlled and Actions masks the token — but it diverges from a convention this repo documented on purpose.
- **Non-blocking — `workflow_dispatch` has no ref guard**, `:34`. Dispatched from any branch it builds that branch's `.devcontainer` and overwrites the single mutable `latest` every workspace boots from. `package.yml:137` shows the repo's habit of gating a writing job on the ref. Requires push access, so a foot-gun rather than a hole; the test name at `:244` ("...to main or by hand") overstates what is enforced.
- **Non-blocking — the workflow assertions are coupled to formatting, not meaning.** `triggers.contains("branches: [main]")` (`:248`) goes red on the semantically identical block-sequence spelling, and `top_level_block` (`:213-236`) is a hand-rolled YAML slicer over comment-stripped text. The file sets a better standard for itself at `:130-131` ("asserted by where the paths resolve rather than how they are spelled") and does not apply it to the workflow half.
- **Non-blocking — the `CLAUDE_VERSION` pin is not self-verifying.** `.devcontainer/Dockerfile:57-61` hands the version to a third-party installer, then runs `claude --version` without comparing the output. `gh`/`zellij` carry their versions in the download URL, so a wrong version cannot resolve; here a silently-ignored argument leaves the pin buying nothing with the build still green. No `pipefail` on the `curl | bash`, mitigated only by that following invocation. (The PR body reports the booted container at 2.1.223, so it does work today — the gap is that nothing keeps it working.)
- **Non-blocking — helper overlap.** `repo_file`, `repo_path` and `resolved_from` (`:16-107`) are three spellings of "path under `CARGO_MANIFEST_DIR`"; `concat!(..., "/")` at `:91` is redundant with `join`. `strip_jsonc_comments` accepts a stricter dialect than JSONC (no trailing commas), which its own panic message concedes. The ~25 duplicated lines across the two configs are forced by the format — devcontainer.json has no inheritance — and the drift test at `:174-197` is the right mitigation.
- **Clean:** `actions/checkout@v7` matches all three existing workflows exactly; workflow-level `contents: read` plus job-level `packages: write` follows `package.yml:15-17`; concurrency shape follows `ci.yml`; test naming and doc-comment density match `skill_docs.rs`. Nothing reportable from the Fowler baseline beyond the duplication noted above. Formatting and lint are tooling's job and were skipped.

## Spec

Measured against #150, taking the fallback branch as sanctioned: *"If devpod ignores it, fall back to `\"image\":` + a `.devcontainer/local/devcontainer.json` variant (`dl --devcontainer local` selects variants) and say so in the PR."* The probe was run and reported, so the bullet asking to *"keep the `build:` block, add `cacheFrom`"* is superseded and its absence is not a finding.

**Met, verified literally:**
- *"Registry `ghcr.io/blooop/wayfinder-devcontainer`, tag `latest` only, `GITHUB_TOKEN`"* — config boots that exact reference (`.devcontainer/devcontainer.json:27`), workflow builds and pushes byte-identical strings (`devcontainer.yml:79,82`), login uses `secrets.GITHUB_TOKEN` (`:65`). No `build:` remains in the default config, which is correct — `build` would win over `image`.
- *"`dl --devcontainer local` selects variants"* — confirmed against the real tool, not the spelling: `dl --help` states a bare name means `.devcontainer/<name>/devcontainer.json`, so the new file is precisely what `--devcontainer local` selects. Its `../Dockerfile` and `..` are asserted by canonicalized resolution (`tests/devcontainer_prebuild.rs:143-186`).
- *"triggers are push to `main` filtered on `.devcontainer/**`, plus `workflow_dispatch`. No PR leg, no schedule"* — literally so (`devcontainer.yml:20-38`), with negative assertions at `tests/devcontainer_prebuild.rs:261-271`.
- *"workflow-level `contents: read`; `packages: write` on the build-push job only"* — `:41` and `:56`, matching `package.yml`.
- *"Plain `docker build .devcontainer` + push — not `devpod build`"* and *"No cargo warming"* — both honored; no cargo step anywhere in the workflow.
- *"update the file's header comment, which currently says #43 is still open"* — replaced (`.devcontainer/devcontainer.json:4-26`) and accurate: it records the probe date, devpod version, and both `cacheFrom` spellings tried.
- *"check if claude's `install.sh` takes a version argument; pin via ARG if so"* — it does (`$1`, validated `stable|latest|VERSION`); `bash -s -- "$CLAUDE_VERSION"` passes it correctly. The pinned binary lands under `~/.local/share/claude/versions/`, not the bind-mounted `~/.claude`, so the pin is not shadowed at runtime.
- *"Before/after timing of that cold `devpod up` goes in the PR"* — present (34.9 s build → zero build lines; 10.4 s vs 10.6 s wall).

**Gaps:**

- **Non-blocking — the private-package failure is not discoverable in-repo.** Spec: *"A workspace booted via `dl blooop/wayfinder@<branch>` on a layer-cold host pulls layers from GHCR."* Until the package is made public that boot fails on `unauthorized`. Judged on legibility, not on being solved: the default config's escape-hatch note (`.devcontainer/devcontainer.json:22-24`) lists two reasons to reach for `local` — a `Dockerfile` change, or an uncovered architecture — and *not* "the pull was denied". `README.md` and `AGENTS.md` mention neither GHCR nor the prebuilt image (grep for `ghcr` / `prebuil`: zero hits). So both flagged conditions land somebody on an auth error or a missing tag with no in-repo path to `--devcontainer local` or to "make the package public" — the answer exists only in this PR body, which is the one place a future user will not look. One added clause at `:22` closes it for both flags at once. The flags themselves are stated squarely and honestly in the PR; it is the *code's* legibility that is short, not the PR's.
- **Non-blocking — `wf` itself cannot take the fallback.** Same spec line. `src/launch.rs:1049` looks only at the two default config locations, and `wf` execs `dl <ws> -- <cmd>` with no `--devcontainer`, so during the publish window (or while the package is private) a `wf`-driven isolated launch has no route to the variant — only a human typing `dl` directly does. Consistent with the spec, which promised only `dl`, but it narrows the mitigation the PR body offers for flag 2.
- **Non-blocking — "pulls layers from GHCR" stays unproven.** Same spec line. The measured "after" column is a locally-built tag, and the cold-download row is computed from layer digests rather than timed. That is the honest comparison available pre-merge and the PR says as much; noted only because the criterion cannot actually be closed until the tag exists and is pullable.

No scope drift: the `ci.yml` step, the test binary, and the `concurrency` block all follow existing house patterns.

## Verdict

**Request changes** — posted as a comment because GitHub refuses a formal review state on a same-account PR; this written verdict is the gate, not the GitHub state.

The spec is met, criterion by criterion, and the fallback path was executed faithfully — the Spec axis found no blocking gap. What blocks is on the Standards axis, and both items are small:

1. **`tests/devcontainer_prebuild.rs:311` — the reference guard passes while the workflow pushes the wrong image, and passes with the Push step deleted.** Both reproduced above, 7/7 green each time. The test the PR nominates as its most load-bearing does not hold the half it advertises. Collect every `ghcr.io/`-prefixed token (or every argument to `docker build -t` / `docker push`) rather than filtering on the expected repository, and assert the push command exists — then mutation-check it the way the config drift guard was.
2. **`AGENTS.md:110-116` and `:124-126` — the documented local check chain omits `cargo test --locked --test devcontainer_prebuild`** and still calls `skill_docs.rs` the sole exception under `tests/`. `ci.yml` disagrees with both.

Everything else above is non-blocking. The two operational flags are raised honestly in the PR, and the judgement asked for is on legibility: the flags are legible to a reviewer, not to a future user hitting the failure — a clause at `.devcontainer/devcontainer.json:22` would fix that and is worth taking while item 2 is being edited anyway.

Not merged, per the run's instruction.

blooop:
Fix pass for the review above, pushed as five commits ending at `554b869`.

**Blocking 1 — the reference guard.** Rebuilt to collect by what the commands *do* rather than by what the reference is expected to look like: every `docker push` argument, every `docker build -t` tag, and the unfiltered `ghcr.io/` sweep, plus a direct assertion that exactly one push of exactly the booted reference exists (replacing the `packages: write` assert that duplicated the permissions test and could not go red alone). Mutation-checked the way the drift guard was, all three red:

- push renamed to `ghcr.io/blooop/wayfinder-image:latest` → **red** (was green)
- Push step deleted → **red** (was green)
- build tag renamed to the bare `wayfinder-image:latest`, no registry prefix at all → **red**

Real workflow: 7/7 green.

**Blocking 2 — AGENTS.md.** The chain now includes `cargo test --locked --test devcontainer_prebuild`, and the prose names both offline binaries under `tests/` as the exceptions instead of calling `skill_docs.rs` the sole one.

**Legibility (taken).** The escape-hatch note in the default config now names the denied pull — GHCR private-by-default, and the tag not existing before the workflow's first run on `main` — as a reason to reach for `--devcontainer local`, and README's Isolation section gained a paragraph on the prebuilt image, so `ghcr`/`prebuil` now grep-hit in README, AGENTS.md and the config.

**Non-blocking, taken:**
- Login token through step `env:` per the `package.yml` pattern, with its reason restated in place.
- The `CLAUDE_VERSION` pin is now compared against `claude --version` output and fails the build on a mismatch — which also backstops the unpiped `curl | bash`.

**Non-blocking, deferred:**
- *`workflow_dispatch` ref guard* — deliberately not added: dispatching from the branch is the only route to publishing the **first** image during the publish window (the workflow file does not exist on `main` until this merges), and a `refs/heads/main` gate would close that route at the moment it is most needed. Foot-gun accepted; it requires push access.
- *YAML-parser-based assertions* (also Sourcery's first point) — a refactor with no behavior change, and a test-only `serde_yaml` dependency. The load-bearing half of the coupling is gone (the reference is now collected from the commands, not their spelling); the trigger/permission substring checks remain as they were.
- *Centralizing the ghcr literal* (Sourcery's second point) — deferred on purpose: the test's literal **is** the decision, and the duplication across config, workflow and test is what lets the test notice one of them moving. A shared constant would let all three drift together.
- *Helper overlap / `strip_jsonc_comments` strictness* — cleanup without behavior change, left for a dedicated pass.
- *`wf` itself having no route to `--devcontainer local`* — real, and out of this ticket's scope (the spec promised the fallback via `dl` only). Worth its own ticket if the publish window bites in practice.
- *"pulls layers from GHCR" proof* — cannot be closed pre-merge; the tag has to exist first.

Not merging, per the run's instruction.

blooop (commented):

**Re-review after the fix pass.** Fixed point: `git diff a3903fa...HEAD` (merge-base with `main`), head `554b869`. Spec: #150. Preflight green — all checks SUCCESS (release job skipped, expected on a PR). The two axes below were produced by independent agents in fresh contexts and are **not** merged or reranked.

Scope of this pass: the prior review's two blocking findings were already independently re-verified upstream of this review — both original mutations (push target renamed off the repository prefix; Push step deleted) now go red on exactly `the_workflow_publishes_exactly_the_reference_the_default_config_boots_from`, baseline 7/7 green — so the axes were pointed at what the fix pass *added*, judged on its own merits. One housekeeping note: despite the fix-pass agent dying mid-run, its summary comment did land on this PR (2026-08-14 11:17Z) and every claim in it was spot-checked true against the diff — the PR record is complete relative to what shipped.

## Standards

**No blocking findings.** The guard now holds the direction the first review broke; remaining holes are outside the drift class it claims to catch. Repo standards (AGENTS.md house style, package.yml conventions) are followed.

- **Non-blocking — a *deliberate second publish* that avoids both token patterns passes the guard.** `tests/devcontainer_prebuild.rs:307-336`, confirmed by two live mutants (applied, run, reverted): appending a `docker/build-push-action@v6` step pushing `docker.io/attacker/wayfinder:latest` stays 7/7 green (no adjacent `docker push` tokens, no `-t`, no `ghcr.io/` prefix), as does a line-continuation split `docker \` / `push docker.io/...`. Both require a **non-ghcr** target and are additions, not the accidental rename/delete drift the guard exists for — the covered direction is real: renaming the actual push argument lands in `pushed_references` unfiltered, deletion empties the set, and a second `ghcr.io/` tag anywhere reddens it. But the assert message at `:364` ("and no other") and the doc comment at `:302-306` ("renamed to anything at all still lands here") overclaim relative to what the tokenizer sees. Worth a comment naming the limit, per this file's own house style of stating what is and isn't asserted.
- **Non-blocking — false-red paths are all in the safe direction.** Legit refactors force a test edit rather than slipping through: `docker push "$IMAGE"` / `${{ env.IMAGE }}` indirection, a trailing-`\` wrap of the push argument, buildx `--push`, and the exact-spelling `branches: [main]` check at `:248` all fail red. Acceptable coupling.
- **Non-blocking nit — comment inaccuracy.** `tests/devcontainer_prebuild.rs:321-323` says the sweep catches "a login line"; the login token is bare `ghcr.io` (no slash) and is not collected. Harmless, mildly misleading.
- **Other fix commits, judged on their own merits — sound.**
  - `554b869` (`.devcontainer/Dockerfile:66-72`): the pin comparison genuinely fails the build (`exit 1`; `set -e` also covers a missing binary), the substring `case` is format-agnostic, and an installer that silently ignores its argument installs latest ≠ pin → red — exactly the self-verification the prior review asked after. Nit: a substring match would accept a version string merely *containing* `2.1.223`; negligible.
  - `ed5ade0` (`.github/workflows/devcontainer.yml:62-71`): mirrors package.yml's step-`env` secret pattern exactly; `--password-stdin` correct.
  - `44cadc3`/`276f4c0`: the AGENTS.md chain now matches ci.yml's order; README/config prose is accurate.
- Fowler baseline: nothing beyond trivial tokenizer duplication between the two collectors; not worth churn.

## Spec

**No spec violations found.** Every bullet of #150 is met on the sanctioned fallback path (*"If devpod ignores it, fall back to `"image":` + a `.devcontainer/local/devcontainer.json` variant … and say so in the PR"* — the probe is documented in the PR body, so the "keep the `build:` block, add `cacheFrom`" bullet is superseded).

- *"triggers are push to `main` filtered on `.devcontainer/**`, plus `workflow_dispatch`. No PR leg, no schedule"* — exactly so; no `pull_request`, no `schedule`.
- *"Plain `docker build .devcontainer` + push — not `devpod build`"* — literal.
- *"Registry `ghcr.io/blooop/wayfinder-devcontainer`, tag `latest` only, `GITHUB_TOKEN`"* — the only reference anywhere in the workflow; login reads `secrets.GITHUB_TOKEN` via step `env:` over stdin.
- *"workflow-level `contents: read`; `packages: write` on the build-push job only (package.yml's narrow-job pattern)"* — matches package.yml's actual shape; `packages: write` appears once, on the publishing job only.
- *"update the file's header comment, which currently says #43 is still open"* — replaced; now records the settlement and the `cacheFrom` measurement.
- *"check if claude's `install.sh` takes a version argument; pin via ARG if so"* — it does (`TARGET="$1"`, validated); the Dockerfile passes `bash -s -- "$CLAUDE_VERSION"` and now fails the build on a mismatch.
- *"No cargo warming"* — none anywhere in the diff.
- *"Before/after timing of that cold `devpod up` goes in the PR"* — present (34.9 s build → none; 10.4 s vs 10.6 s wall; digest-based download comparison). The *"pulls layers from GHCR"* leg remains unclosable pre-merge, as the PR states — carried finding, not new.
- Fix-pass claims spot-checked true: AGENTS.md names both offline binaries and includes `--test devcontainer_prebuild`; README gained the prebuilt-image paragraph; the config's escape-hatch comment names the denied pull (GHCR private-by-default, tag absent before the first `main` run).
- Beyond spec, factual: the 380-line test binary + ci.yml step, the `concurrency` block, the version-mismatch check, and the doc edits — all following house patterns.

## Verdict

**Approve.** Both prior blocking findings are fixed and were re-verified by mutation upstream of this review; the rebuilt guard was then attacked fresh and holds every accidental-drift direction it exists for, with its remaining blind spot (a deliberately added non-ghcr publish) outside that class and worth only a clarifying comment. The Spec axis found zero violations. The non-blocking items above — the overclaiming assert message/doc comment at `tests/devcontainer_prebuild.rs:302-306`/`:364`, the "login line" comment nit, the substring-match nit in the Dockerfile pin check — are recorded for a future pass, none gating.

Posted as a comment because GitHub refuses a formal review state on a same-account PR; this written verdict is the gate, not the GitHub state. Not merged, per the run's instruction.

PR #155, opened 2026-08-14
Model the cursor writes: staging chooses, and the empty list is pinned

### LINKED ISSUE #148: Model the cursor writes: staging a launch chooses a row, and an empty list is pinned
The narrow remainder of #90, split out so that ticket can close on the question it actually asked.

#90 asked what should happen to the cursor when the scope changes. That question was answered by #135 — `Scope` was deleted, and `App::enter`/`App::leave` now own level, query and cursor together, so there is no longer a scope change that can move the list behind the cursor's back. The query-path `Untouched` writes are pinned.

What #135 did not settle is *when a cursor write happens at all*. Three items, all in `src/app.rs`:

**1. The empty-list write is unpinned.** `move_sibling` sets `Cursor::Untouched` when there are no stops (`src/app.rs:665-671`), with a comment saying why: "pressing a key over an empty list is not a selection." Nothing holds it. Flipping it to `Cursor::Chosen(0)` leaves **390 tests passing** — verified by running it. Either the rule matters and a test should say so, or it does not and the comment is overstating.

**2. Staging a launch does not record choosing the row.** `request_launch` (`src/app.rs:810`) reads `self.cursor_stop()` and writes no `self.cursor` anywhere in its body. So a row can be launched from without ever having been *chosen*, and the two notions drift: the launch acts on a stop the cursor model still calls untouched. Decide whether staging is a choice — it plainly is to the user — and make the write follow.

**3. `pinned` carries a shape that can hold a contradiction.** `src/app.rs:781` builds `Option<(Option<StopKey>, usize)>`. The nested option means "pinned by identity" and "pinned by position" are two states inside one value rather than two arms, so a `(None, n)` and a `(Some(k), n)` are told apart by reading the inner option at the use site rather than by matching. This is the shape the house style says to avoid — where a sum type already says something, do not add a field that can disagree with it.

**Done when** all three are decided: each write is either pinned by a test or deleted with a reason, `request_launch`'s relationship to the cursor is explicit, and `pinned` is a type whose arms are the two things it means.

Closing #90 and #47 on this split — #47's own "Not yet specified" says its remainder *is* #90, and its destination shipped in 0.17.1.

blooop:
Closes #148.

Three slices, one per item in the ticket, all in the cursor model:

1. **The empty-list write is now pinned.** `move_sibling`'s `Cursor::Untouched` write over an empty list was held by nothing — flipping it to `Chosen(0)` left all 390 lib tests green. The new test walks the path where the rule is observable: sift to nothing, press `↓`, let a matching map land. A recorded `Chosen(0)` would put the cursor on the map header a sifted screen leads with, where `enter` stages the whole map instead of the row the query was typed for. The test fails under that mutation (verified).

2. **Staging a launch chooses the row it acts on.** `request_launch` read `cursor_stop()` and wrote no cursor, so a row could be launched from without ever being chosen. Every arm that opens the launch picker now records `Cursor::Chosen` at the stop it staged; refusals (blocked, done, nothing selected) and the group toggle still write nothing, because nothing was staged. Pinned by a test that stages from a defaulted cursor and asserts a fresher map sorting above does not steal the selection.

3. **`pinned` has its two arms.** The `Option<(Option<StopKey>, usize)>` in `replace_clusters` became a `Pinned` enum — `Identity(key, fallback)` and `Position(fallback)` — so pinned-by-identity and pinned-by-position are matched, not read out of a nested option. Behavior unchanged; the existing suite pins it.

All CI gates run locally: fmt, clippy (all targets, all features), lib/bins/examples tests (392), skill_docs, live_fetch common, all-targets build, doc.

## Summary by Sourcery

Clarify cursor behavior around cluster refreshes and launches so that selections are preserved correctly and not spuriously created over empty lists.

New Features:
- Ensure staging a launch explicitly chooses the row it acts on so subsequent refreshes keep the cursor on that ticket.

Bug Fixes:
- Prevent a down-arrow keypress over an empty, sifted list from being treated as a real selection that would pin the cursor to the map header after new data arrives.

Enhancements:
- Introduce a dedicated Pinned enum to distinguish cursor pinning by stop identity versus by position during cluster replacement.
- Add tests covering cursor behavior when sifting to an empty list and when staging a launch from a defaulted cursor position.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -208,6 +208,23 @@ enum Cursor {
     Chosen(usize),
 }
 
+/// How a chosen cursor rides out a cluster swap: two different holds, so two
+/// arms (#148). **Identity** is the normal one — the stop under the cursor,
+/// carried with the old position as the fallback should it vanish from the
+/// new order. **Position** is what remains when the cursor is chosen over an
+/// *empty* screen — a swap can empty the list under a choice — where there is
+/// no stop to name. A `None` inside a tuple used to say that by omission,
+/// which let "pinned by identity" and "pinned by position" share one shape
+/// whose halves were told apart by reading the inner option at the use site
+/// rather than by matching.
+enum Pinned {
+    /// Follow this stop wherever the new order puts it; fall back to the old
+    /// position, clamped, if it is gone.
+    Identity(StopKey, usize),
+    /// Nothing to follow: hold the bare position, clamped.
+    Position(usize),
+}
+
 #[derive(Debug)]
 pub struct App {
     /// The clusters on screen: every open map that has arrived, keyed by id.
@@ -780,21 +797,28 @@ impl App {
     pub fn replace_clusters(&mut self, clusters: BTreeMap<MapId, Map>) {
         let pinned = match self.cursor {
             Cursor::Untouched => None,
-            Cursor::Chosen(_) => Some((self.cursor_key(), self.cursor_pos())),
+            Cursor::Chosen(_) => Some(match self.cursor_key() {
+                Some(key) => Pinned::Identity(key, self.cursor_pos()),
+                None => Pinned::Position(self.cursor_pos()),
+            }),
         };
         self.clusters = clusters;
-        if let Some((anchor, old_index)) = pinned {
-            let new_order: Vec<StopKey> = self
-                .stops()
-                .iter()
-                .map(|at| self.stop_key(&at.stop))
-                .collect();
-            self.cursor = Cursor::Chosen(crate::refresh::preserve_cursor(
-                anchor.as_ref(),
-                old_index,
-                &new_order,
-            ));
-        }
+        let Some(pinned) = pinned else {
+            return;
+        };
+        let new_order: Vec<StopKey> = self
+            .stops()
+            .iter()
+            .map(|at| self.stop_key(&at.stop))
+            .collect();
+        self.cursor = Cursor::Chosen(match pinned {
+            Pinned::Identity(key, fallback) => {
+                crate::refresh::preserve_cursor(Some(&key), fallback, &new_order)
+            }
+            Pinned::Position(fallback) => {
+                crate::refresh::preserve_cursor(None, fallback, &new_order)
+            }
+        });
     }
 
     /// The first enter (#62): stage a launch of whatever the cursor is on by
--- a/src/app.rs
+++ b/src/app.rs
@@ -1599,6 +1599,37 @@ mod tests {
     }
 
     #[test]
+    fn a_key_over_an_empty_list_is_not_a_selection() {
+        // The sift has emptied the screen, so `↓` has nothing to land on. If
+        // that keypress were written down as a choice, the rows arriving next
+        // would find the cursor pinned to position 0 — the map header a sifted
+        // screen leads with — and `enter` there would stage the whole map
+        // rather than the row the query was typed for.
+        let mut app = fixture_app();
+        type_str(&mut app, "zzz");
+        assert!(app.stops().is_empty(), "the query should match nothing");
+
+        app.handle_key(key(KeyCode::Down));
+
+        let mut fresher = app.clusters.clone();
+        fresher.insert(
+            MapId::new(PROJECT, 99),
+            Map {
+                title: "Map: late arrival".to_string(),
+                last_activity: None,
+                tickets: vec![ticket(PROJECT, 200, "zzz sleeper", true, false, vec![])],
+            },
+        );
+        app.replace_clusters(fresher);
+
+        assert_eq!(
+            app.cursor_ticket().map(|t| t.number),
+            Some(200),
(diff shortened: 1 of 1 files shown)


blooop (commented):

Two-axis review of #155 against ticket #148, at merge-base `a3903fa`. Axes are independent and not reranked against each other.

## Standards

Repo standards consulted: `AGENTS.md` ("House style"); no other standards docs exist. Fowler 12-smell baseline applied; nothing tooling already enforces is reported.

1. **Duplicated Code / invariant-by-prose (non-blocking)** — `src/app.rs:860-866`, `885-891`, `927-933`: three byte-identical blocks of `self.cursor = Cursor::Chosen(self.cursor_pos());` followed by the same four-field `Overlay::PickLaunch` construction. The doc comment at `src/app.rs:835-841` states the invariant ("every arm that opens the picker records the cursor as Chosen there") as a convention a future arm can silently break. House style keeps illegal states unrepresentable; "picker open, cursor not written" is exactly such a state, and a private `open_picker(&mut self, staged: Staged)` helper would make the pairing structural instead of documented. The overlay duplication predates this diff, but the diff adds a fourth duplicated line to each copy.
2. **Primitive Obsession residue (non-blocking)** — `src/app.rs:814-820`: the only consumer of `Pinned` matches it solely to rebuild `Some(&key)` / `None` for `preserve_cursor` (`src/refresh.rs:432`), whose `(Option<&K>, usize)` parameters are the original clump — the "inner option read at the use site" the enum's own doc condemns moved from construction site to call site. The write side is genuinely fixed; a `Pinned::resolve(self, new_order: &[StopKey]) -> usize` would absorb the match without touching the generic, well-tested `preserve_cursor`.
3. **Missed-arm check: none missed.** `src/app.rs:1114` constructs `PickLaunch` but is the picker's own key handler re-storing an overlay taken by value, not an opening; `src/ui.rs:860` is render-only.
4. **Nit** — the write at `src/app.rs:860` (Map arm) is nearly a no-op since a header is unreachable by an untouched cursor (`cursor_pos`, `src/app.rs:588-597`), but uniformity across arms matches the stated invariant; no change requested.

Naming, test names, and comment quality match house style.

## Spec

All three ticket items and the Done-when are satisfied.

- **Item 1 — "Either the rule matters and a test should say so, or it does not."** Pinned with a test. `a_key_over_an_empty_list_is_not_a_selection` (`src/app.rs:1636-1665`) genuinely exercises the arm: `type_str("zzz")` resets the cursor to `Untouched` and empties `stops()` (asserted), so `Down` → `move_sibling` hits the empty-list branch at `src/app.rs:684-689` and nothing else writes. Under `Chosen(0)` the post-swap position 0 is the `Stop::Map(99)` header, so `cursor_ticket()` would be `None`, not `Some(200)` — observably failing. *Nit (non-blocking):* the test would still pass if the write were deleted outright rather than flipped; acceptable, since the ticket offered pin-or-delete and this pins the named mutation.
- **Item 2 — "Decide whether staging is a choice and make the write follow."** Explicit in the doc (`src/app.rs:835-841`) and total in the code: every picker-opening arm writes `Cursor::Chosen(self.cursor_pos())` first — Map (860), Project-on-own-screen (885), Ticket via `request_ticket_launch` (927) — and every refusal writes nothing: group toggle (848-853), blocked (906-913), done (920-922), no selection (895-898).
- **Item 3 — "`pinned` is a type whose arms are the two things it means."** `enum Pinned` (`src/app.rs:218-223`), `Identity(StopKey, usize)` vs `Position(usize)`, matched exhaustively at `src/app.rs:814-821`. Behavior vs the old tuple path is identical: same reads, same `preserve_cursor` calls, same early return for `Untouched`.
- **Write correctness:** `Cursor::Chosen(self.cursor_pos())` from `Untouched` is sound — `cursor_stop()` is defined as `stops().get(self.cursor_pos())`, and nothing between the read and the write mutates `plan()`'s inputs, so the pinned index is the row just read.

## Mutation verification

The PR's mutation claims were re-run in an isolated worktree rather than trusted:

- **Base branch (pre-PR):** flipping `move_sibling`'s empty-list write to `Chosen(0)` left **390/390** lib tests green — the ticket's premise reproduces.
- **PR branch, same mutation:** **391 passed, 1 failed** — exactly `a_key_over_an_empty_list_is_not_a_selection`. Claim verified.
- **PR branch, deleting the ticket-arm cursor write (`src/app.rs:927`):** **391 passed, 1 failed** — exactly `staging_a_launch_chooses_the_row_it_acts_on`. Claim verified.
- Baseline unmutated: 392 passed, matching the PR's stated 390 → 392.

## Verdict

**Approve.** No blocking findings on either axis; both new tests are verified to catch the mutations they were written for; CI is fully green including the real-dl contract job. The two Standards findings (picker-open helper, `Pinned::resolve`) are non-blocking refactors, fine to leave for a future ticket. Review state posted as a comment where GitHub refuses a same-account approval — this written verdict is the gate. Not merging; that stays human.

PR #157, opened 2026-08-14
The ctx handoff's guards guard what they claim

### LINKED ISSUE #133: The ctx handoff's guards don't yet guard what they claim
## Question

The re-verdict on [PR #131](https://github.com/blooop/wayfinder/pull/131#pullrequestreview-4891273518) approved the `ctx:` handoff with six non-blocking findings, deliberately not held against the merge. They are collected here so they are worked once rather than rediscovered one at a time. Two of them are guards that do not guard what they claim — the failure mode this map has now hit repeatedly.

### 1. `*_word`'s comment overclaims exhaustiveness

The `*_word` functions have wildcard-free `match`es, so a new variant stops the module compiling — that part is real. But the **iteration** in the matrix test is hand-written arrays. The Standards axis proved the gap by adding a `Launchable::Spiking` variant: it **compiles and greens** once its word is pinned, while never being launched in the matrix and never being required of the docs.

So the comment claims a completeness the tests do not deliver. Either make the iteration exhaustive (derive it from the type rather than restating the variants), or correct the comment to say what is actually enforced. Prefer the former — the whole point of the wildcard-free `match` was that the compiler catches new variants, and hand-written iteration arrays reintroduce exactly the drift it removed.

### 2. The docs guard binds the schema but not the contract prose

`tests/skill_docs.rs` now binds documented field names and vocabularies to what the types serialize — that was the fix for the guard that let nonsense docs pass. But the **contract prose** is still unbound: the precedence sentence (orient from ctx, verify live, **argv wins**) can be **inverted** with the whole suite green.

That sentence is the contract. A doc guard that pins the vocabulary but lets the rule invert is guarding the cheaper half.

### 3. `tests/live_launch_exec.rs` substring-scans for `"assignee"` / `"claim"`

Precisely the anti-pattern this PR's retry removed from the unit test, where the claim-free invariant became an exact field-name set check on parsed JSON. The live test still scans text for substrings — and, being a live test, it is the one file nobody runs, so the weaker assertion is also the least observed. Bring it in line with the unit test's shape.

### 4. `README.md` says "Every launch of a node…"

It excludes plain mode, contradicting the surrounding text in the same file and the PR's own test. A reader following the README would expect a block where none is emitted.

### 5. The scratch dir leaks into `$TMPDIR` on the shell-refusal path

The seam canary's per-call temp dir is removed on the normal path, but on the shell-refusal path it is left behind. Never in the repo — that was the blocking finding and it is fixed — but a long-lived host accumulates `wf-seam-*` directories.

### 6. #122 §5's `live_launch_exec.rs` change is missing from the departures list

The PR body's departures list is otherwise scrupulous, which is exactly why a gap in it costs more than the item itself. The binding clause ("no test anywhere may require ctx for a skill to proceed") **is** satisfied and stated in the body; only the departure label is missing. Note the Spec axis graded this **blocking** and the compiling reviewer graded it non-blocking — the disagreement is recorded in the review rather than hidden, and this ticket inherits it as a should-fix.

### Seams to test at

- **The matrix's iteration** (item 1): a test that fails when a variant exists but is not exercised — i.e. derive the iteration from the type so adding a variant cannot green without being launched. Reproduce the axis's `Launchable::Spiking` probe as the red case.
- **`tests/skill_docs.rs`** (items 2 and 4): extend it to bind the contract sentences it currently only extracts, so inverting the precedence rule or contradicting the plain-mode statement turns it red. Prove by mutation, under the **full** `--test skill_docs` selection, not a single-test run.
- **The claim-free invariant in `tests/live_launch_exec.rs`** (item 3): the exact-field-name-set assertion the unit test already uses. Note this file is excluded from the CI chain by design and has never been executed — say so rather than implying it now guards at runtime.
- **The canary's cleanup path** (item 5): assert the scratch dir is gone on the refusal path too, the way the normal path already asserts it is empty.

Every new or changed guard here must be **proven by mutation**, with the selection stated. This map has found three tests passing while guarding nothing, and two of the items above are more of the same.

blooop:
Closes #133.

The six findings #131's re-verdict deliberately did not hold against the merge, worked as one PR. Every changed guard is proven by mutation, with the selection stated; the manager's re-verification against post-#135 main was re-confirmed here before each fix.

### What changed

**Item 1 — the matrix iterates the type, not a restatement of it.** Reproduced the axis's probe first: on current main, `Launchable::Spiking` compiled and greened with its word pinned, never launched, never required of the docs. Now a single `every_variant!` list feeds both a wildcard-free `match` and the returned iteration, so the completeness proof and the loop cannot disagree; the launch matrix and the doc vocabularies iterate `TicketType::every()`, `Launchable::every()`, `Checks::every()`, `Review::every()`, `PrStatus::every_arm()`, `Aim::every_arm()`. Mutation (selections `--lib launch`, `--test skill_docs`): re-adding `Spiking` fails to compile in `every()` and in the matrix's stage inverse; force-listing it and mapping it onto an existing stage fails the matrix at runtime (`expected "stage":"spiking"`, wire says `"ready"`); the doc-vocabulary test then demands the word the doc never published.

**Item 2 — the contract prose is bound, not just the vocabulary.** Re-ran the mutation experiment on the post-#135 file first, as instructed: inverting the precedence sentence to "the block beats your arguments" left the full `--test skill_docs` selection green (7/7), so the finding still stood. The two normative sentences (accelerator/verify-live/discard-whole, and the precedence line) are now pinned whole by equality; the inverted doc fails the new test under the full selection, the true doc passes all 9.

**Item 3 — the live test asserts field names, not substrings.** The `"assignee"`/`"claim"` text scan is replaced with the unit tests' shape: every key of the parsed block must be one the v1 schema writes, the aim-independent core must be present, and the forbidden names are checked against the allowed universe. The file stays excluded from the CI chain by design and now says so at the assertion. Beyond the ticket's expectation that it never runs: it was executed here by name (`--test live_launch_exec -- enter_execs_the_agent`) — green against the real tracker, and red at the new assertion with the binary mutated to smuggle an `assignee` field into `LaunchContext`.

**Item 4 — the README promises the block only where a skill receives it.** "Every launch of a node" is now "Every **skill** launch", and the exclusions paragraph names the plain rows beside the creation rows. Both paragraphs are pinned by equality in `skill_docs`; the old wording fails the new test, and flipping the exclusion to "the plain rows carry one too" fails it again.

**Item 5 — the seam canary cleans up on the refusal path.** The scratch dir is now removed before the shell's exit status is asserted, and a test drives a real refused command (an unterminated quote) and asserts the `wf-seam-*` dir is gone. Red before the reorder, green after.

### Departures

- **Item 6, inherited from #131:** the binding clause of #122 §5 ("no test anywhere may require ctx for a skill to proceed") was satisfied and stated in #131's body, but its `live_launch_exec.rs` change was missing from that PR's departures list. Recorded here instead: that change was a departure from #122 §5's letter, and the Spec axis graded the omission blocking where the compiling reviewer graded it non-blocking.
- **The live assertion is bounded, not pinned.** The ticket asked for "the exact-field-name-set assertion the unit test already uses"; the live node varies (stage, PRs), so an equality pin would fail on tracker state. The assertion is subset-of-schema plus mandatory core instead — every key must be one the schema writes, which still fails on any new field.
- **`tests/skill_docs.rs` now also reads `README.md`.** The ticket routed items 2 and 4 at that seam; the file's charter is skill docs, and the README paragraphs it binds are the same ctx contract the skills consume.
- `Aim::every_arm()`/`PrStatus::every_arm()` carry one representative payload per data-carrying arm (checked at run time to be the arm it stands for); callers expand the payload grid themselves.

## Summary by Sourcery

Tighten ctx contract guarantees by driving both launch behavior and documentation from compiler-complete enumerations and literal contract text, and by strengthening live execution and seam cleanup tests.

New Features:
- Add compiler-checked enumeration helpers and per-type iteration APIs so launch matrices and doc vocabulary tests iterate the actual variants for ticket types, stages, PR status, checks, and review decisions.
- Introduce tests that pin key ctx contract sentences and README paragraphs by exact equality to ensure documented behavior matches the schema and runtime guarantees.

Bug Fixes:
- Ensure seam scratch directories are cleaned up on both successful and refused shell command paths to avoid leaking wf-seam-* directories on long-lived hosts.

Enhancements:
- Refactor launch matrix and ctx emission tests to use type-provided variant iterators and a stage inverse helper, ensuring every variant is both launched and validated on the wire.
- Strengthen live launch execution guards to assert context block field names against the v1 schema and enforce mandatory core fields, replacing brittle substring scans and expanding the forbidden-field checks.
- Document and enforce that ctx blocks are only emitted for skill launches and not for plain or creation rows, aligning runtime behavior with README promises.

Tests:
- Extend skill_docs tests to cover emitted vocabularies via type-driven iterators, literal contract prose, and README context guarantees.
- Add live execution tests that mirror unit-test schema shape assertions for ctx blocks and verify behavior when mutated binaries attempt to introduce forbidden fields.
- Add a seam test for refused commands to confirm scratch directory cleanup on panic paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -752,8 +752,8 @@ through `codex --dangerously-bypass-approvals-and-sandbox`. Both receive one
 prompt argument, so the route, numbers, and steering text cannot be split apart
 by the shell.
 
-**Every launch of a node also hands the agent what `wf` already knew**, as a
-`ctx: <json>` block between the skill's arguments and any steering suffix:
+**Every skill launch of a node also hands the agent what `wf` already knew**,
+as a `ctx: <json>` block between the skill's arguments and any steering suffix:
 
 ```
 /wf-review 124 ctx: {"v":1,"repo":"blooop/wayfinder","map":{…},"aim":{"ticket":{…,"prs":[…]}}} steer: <text>
@@ -766,7 +766,9 @@ never went through the picker, finds no block, and discovers exactly as before.
 The one thing it deliberately cannot say is whether the ticket is still yours to
 take — there is no assignee and no ticket status in the schema, so claiming stays
 a live call and a stale block cannot make an agent act on someone else's work.
-The creation rows carry no block at all: they name nothing that exists yet.
+The creation rows carry no block at all: they name nothing that exists yet, and
+the plain rows carry none either — a block is addressed to a skill, and plain
+runs none.
 
 The auto mode collapses the ticket rows on purpose: the launched session is a
 *manager*, and what it manages is the node's whole remaining lifecycle —
--- a/src/launch.rs
+++ b/src/launch.rs
@@ -601,6 +601,13 @@ pub enum Launchable {
 }
 
 impl Launchable {
+    /// Every launchable stage, compiler-complete
+    /// ([`crate::model::every_variant`]): the iteration the launch matrix and
+    /// the doc vocabulary run over, so a stage cannot exist unlaunched (#133).
+    pub fn every() -> Vec<Launchable> {
+        crate::model::every_variant!(Launchable: Ready, Building, InReview, NeedsAttention)
+    }
+
     /// Narrow a derived stage to one an agent can be launched on. Exhaustive
     /// on [`Stage`]: a new stage must decide whether it is launchable, or this
     /// stops compiling.
@@ -651,6 +658,25 @@ pub enum Aim {
         stage: Launchable,
         prs: Vec<PrLink>,
     },
+}
+
+impl Aim {
+    /// One value of every arm, compiler-complete
+    /// ([`crate::model::every_variant`]) — the wire only ever sees the tag,
+    /// so the ticket representative's payload is the doc guards' business to
+    /// vary, not this list's.
+    pub fn every_arm() -> Vec<Aim> {
+        crate::model::every_variant!(Aim:
+            Map,
+            Ticket => Aim::Ticket {
+                number: 1,
+                title: String::new(),
+                ticket_type: TicketType::Build,
+                stage: Launchable::Ready,
+                prs: vec![],
+            },
+        )
+    }
 }
 
 /// The map a launch was picked in: its identity *and* its title (#124).
@@ -3157,9 +3183,13 @@ mod tests {
     /// These four are the whole vocabulary the tracker doc publishes, written
     /// as literals traceable to it rather than derived from the types — a
     /// derivation would only prove serde agrees with itself. Being `match`es
-    /// with no wildcard is the other half: adding a type, a stage, a check
-    /// rollup or a review decision stops this module compiling until the new
-    /// word has been decided and pinned here.
+    /// with no wildcard is half of the guarantee: adding a type, a stage, a
+    /// check rollup or a review decision stops this module compiling until
+    /// the new word has been decided and pinned here. The other half is the
+    /// matrix below iterating [`TicketType::every`] and friends rather than
+    /// hand-written arrays, so the new variant is also *launched* — a probe
+    /// variant once compiled and greened with its word pinned while the
+    /// arrays never exercised it (#133).
     fn type_word(ticket_type: TicketType) -> &'static str {
         match ticket_type {
             TicketType::Build => "build",
@@ -3225,25 +3255,16 @@ mod tests {
     /// uses. Before this, only `ready`, `build` and a single open PR were ever
     /// emitted by any test, so every other word the tracker doc publishes
     /// rested on the doc's say-so.
+    ///
+    /// "Whole" is the type's own claim, not this test's: the iteration comes
+    /// from [`TicketType::every`] and friends, where the compiler holds the
+    /// list complete, so a new variant cannot green without a row here (#133).
     #[test]
     fn every_stage_type_and_pr_state_spells_itself_on_the_wire() {
-        let types = [
-            TicketType::Build,
-            TicketType::Research,
-            TicketType::Task,
-            TicketType::Grilling,
-            TicketType::Prototype,
-            TicketType::Untyped,
-        ];
-        let stages = [
-            (Stage::Ready, Launchable::Ready),
-            (Stage::Building, Launchable::Building),
-            (Stage::InReview, Launchable::InReview),
-            (Stage::NeedsAttention, Launchable::NeedsAttention),
-        ];
-        for ticket_type in types {
-            for (stage, launchable) in stages {
-                let ctx = ctx_of(&ticket_prompt(ticket_type, stage, &interactive("")))
+        for ticket_type in TicketType::every() {
+            for launchable in Launchable::every() {
+                let prompt = ticket_prompt(ticket_type, staged_at(launchable), &interactive(""));
+                let ctx = ctx_of(&prompt)
                     .expect("a ticket launch carries context")
                     .to_string();
                 let expected = format!(
@@ -3254,20 +3275,17 @@ mod tests {
                 assert!(ctx.contains(&expected), "expected {expected} in {ctx}");
             }
         }
-        let mut pr_states = vec![PrStatus::Draft, PrStatus::Merged, PrStatus::Closed];
-        for checks in [
-            Checks::Absent,
-            Checks::Pending,
-            Checks::Passing,
-            Checks::Failing,
-        ] {
-            for review in [
-                Review::NotRequired,
-                Review::Required,
-                Review::Approved,
-                Review::ChangesRequested,
-            ] {
-                pr_states.push(PrStatus::Open { checks, review });
+        let mut pr_states = Vec::new();
+        for arm in PrStatus::every_arm() {
+            match arm {
+                PrStatus::Open { .. } => {
+                    for checks in Checks::every() {
+                        for review in Review::every() {
+                            pr_states.push(PrStatus::Open { checks, review });
+                        }
+                    }
+                }
+                settled => pr_states.push(settled),
             }
         }
         for status in pr_states {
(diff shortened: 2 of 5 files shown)


blooop (commented):

Two-axis review of #157 against ticket #133, from a fresh context. One note up front: the launch record said the build agent died before writing its report, so this review was primed to find a body missing or thin relative to the diff. It is neither — the body accounts for every changed file, states its mutation selections, and carries a departures section; every mutation claim in it that this review reproduced held. The diff was still judged on its own, with the body treated as claims to verify rather than narrative to trust.

## Standards

- **Repo standards (AGENTS.md "House style") — met overall.** Comments carry the why, illegal states stay unrepresentable, no new lint `allow`s, README edited alongside behavior. The `every_variant!` design (exhaustive-match closure feeding the same list as the `Vec`, duplicate variants dying on denied `unreachable_patterns`) is a genuinely sound "iterate the type, not a restatement" mechanism, and replacing five hand-written arrays with it is the right refactor. Nothing blocking.
- **should-fix — Duplicated Code, under-acknowledged: `tests/live_launch_exec.rs:631-681`.** The live test hand-copies *three* things from `the_handed_context_cannot_speak_about_the_claim` (`src/launch.rs:3084`, `3146-3178`): the `keys_of` helper, the 14-key v1 schema set, and the 6-name forbidden list. The comment at `tests/live_launch_exec.rs:222-225` owns up to the helper copy only; the schema set and forbidden list are equally hand-synced, and drift there is caught only when this CI-excluded file is run by name. This also cuts against the PR's own thesis — the schema set is exactly a "restatement of the type". Since integration tests link the lib, deriving the schema keys from serializing one richest ctx via the pub API (or a `#[doc(hidden)]` shared const) would remove all three copies; at minimum the comment should own all three.
- **nit — self-referential assertion: `tests/live_launch_exec.rs:669-681` (and its twin `src/launch.rs:3166-3178`).** The forbidden loop asserts against the `schema` literal declared ~20 lines above, not against the wire keys — a constant checked against a constant. It guards future edits of the adjacent literal, which is legitimate, but a comment saying so would stop a reader thinking it tests the block.
- **nit — inverse map written twice: `src/launch.rs:3306-3321`.** `staged_at` restates `Launchable::parse` backwards. Both are wildcard-free so drift can't compile silently, but a one-line round-trip assert (`Launchable::parse(staged_at(l)) == Some(l)`) inside the helper would bind them for free.
- **nit — prose pinned wider than the invariant: `tests/skill_docs.rs:369-384`.** The README test equality-pins the entire paragraph, including non-normative sentences, so any typo touch there is a two-file edit. Convincing for the two normative sentences; the rest is collateral. Accepted-tradeoff territory.
- **observation:** `every`/`every_arm` are production-pub solely so `tests/skill_docs.rs` can reach them; unavoidable for integration tests and well-documented.

No blocking findings.

## Spec

- **Item 1 — "derive the iteration from the type rather than restating the variants" / "Reproduce the axis's `Launchable::Spiking` probe as the red case": met.** `every_variant!` (`src/model.rs`) feeds one list into both a wildcard-free `match` and the returned `Vec`, so omission is a compile error and the vec cannot diverge; the matrix and doc vocabularies now iterate `TicketType::every()` and friends. The overclaiming comment was corrected.
- **Item 2 — "extend it to bind the contract sentences… so inverting the precedence rule… turns it red. Prove by mutation, under the full `--test skill_docs` selection": met.** Both normative sentences are pinned by equality in `the_contract_sentences_read_exactly_as_the_contract_rules`.
- **Item 3 — "the exact-field-name-set assertion the unit test already uses… say so rather than implying it now guards at runtime": met, via declared departure.** The substring scan is replaced with `keys_of` subset-of-schema plus a mandatory core; the CI exclusion is stated at the assertion. The bounded-not-pinned departure is legitimate: the live stage and PRs vary, and subset-of-schema still fails closed on any new field (forbidden ∉ schema ∧ keys ⊆ schema ⟹ forbidden ∉ keys). Minor: the allowed-set and `keys_of` are hand-copied restatements of the unit side — acknowledged in-comment, but it is the drift pattern item 1 exists to kill (the Standards axis grades this a should-fix).
- **Item 4 — README "excludes plain mode": met.** "Every skill launch" plus the plain-rows sentence, both equality-pinned in `skill_docs`.
- **Item 5 — "assert the scratch dir is gone on the refusal path too": met.** `remove_dir_all` now precedes the status assert; `a_refused_command_leaves_no_scratch_behind` drives a real refusal and asserts absence. **Should-fix:** against the spec's "proven by mutation, with the selection stated", item 5's body says only "Red before the reorder, green after" — no selection named, where items 1–3 state theirs.
- **Item 6 — "only the departure label is missing": met.** The Departures section records the #122 §5 `live_launch_exec.rs` departure and the blocking/non-blocking disagreement, per the ticket's "this ticket inherits it as a should-fix".
- **Departures section: legitimate.** Bounded live assertion — justified above; `skill_docs` reading README.md — the ticket itself routed item 4 "at that seam"; the representative-payload note is sound and enforced by the macro's runtime `matches!` check.

Fully met: items 1, 2, 3, 4, 6. Item 5 met in substance, should-fix on the unstated selection.

## Mutation verification — every changed guard seen red

The ticket's closing rule is that a guard not seen to fail is the defect under review, so this review re-ran the mutations independently rather than trusting the body. Baseline first: `--lib launch` 85/85, full `--test skill_docs` 9/9, and `--test live_launch_exec -- enter_execs_the_agent` green against the real tracker (8.95s). Then, each mutation applied in an isolated worktree and reverted:

| Mutation | Selection | Result |
|---|---|---|
| Bare `Launchable::Spiking` variant (the ticket's probe) | `--lib launch` (build) | **4 compile errors**: `every()`'s macro match, the route match, `stage_word`, `staged_at` |
| `Spiking` force-listed, word pinned, `staged_at` mapping it to `Stage::Ready` | `--lib launch` | **red at runtime**: `expected "stage":"spiking"`, wire says `"ready"` |
| — same mutation — | full `--test skill_docs` | **red**: documented `stage` vocabulary must now contain `"spiking"` |
| Precedence sentence inverted ("the block beats your arguments") | full `--test skill_docs` | **red** at the new equality test, 8 passed 1 failed |
| README reverted to "Every launch of a node" | full `--test skill_docs` | **red** |
| README exclusion flipped ("the plain rows carry one too") | full `--test skill_docs` | **red** |
| Seam cleanup reordered back after the success assert | `--lib a_refused_command_leaves_no_scratch_behind` | **red**: "the refusal path must clean its scratch too" |
| `assignee: u32` smuggled into `LaunchContext` | `--test live_launch_exec -- enter_execs_the_agent` | **red**: `"assignee" is not a field the v1 schema writes` — against the live tracker |

Every guard the PR adds or changes was observed failing under the mutation it claims to catch, under the selection the body states (item 5's selection, unstated in the body, is `--lib a_refused_command_leaves_no_scratch_behind` — recorded here).

## Verdict

**Approve** (posted as a comment: the reviewing account authored the PR, so GitHub refuses the approve state — this written verdict is the gate's input). No blocking findings on either axis. CI is green including the real-dl contract job. Not merging — that stays human.

Should-fixes to carry forward, none gating:
1. The live test's schema set, forbidden list, and `keys_of` are three hand-copies of the unit side with only one acknowledged (`tests/live_launch_exec.rs:631-681`).
2. Item 5's mutation selection is not stated in the PR body (now recorded above).

Issue #161, opened 2026-08-14
Which wf-side stages enter the trend, and where does wf bench?


blooop:
## Question

Grilling: which wf-side stages enter the trend, and where does wf bench — its own CI, or riding devlaunch's?

The wf side may be too small and too TUI-entangled to bench headlessly at honest fidelity: the interval of interest starts at a human keypress inside a drawn picker. Candidate positions to grill:

- **Ride devlaunch's trend entirely**: the stamped handoff stage *is* wf's number; wf's own CI benches nothing. Simplest; but it only measures launches the devlaunch bench makes, which have no real picker in front of them — the handoff stage there measures the stamp mechanics, not wf.
- **A wf-side micro-bench**: time keypress-to-exec with a scripted TUI (or a headless launch path if one exists), tracked in wf's CI with the same alert-don't-gate shape. Honest about wf's own cost, but new machinery for a stage measured in tens of milliseconds — unless prewarm-miss is in the interval, where it's tens of seconds.
- **Track the discriminator, not the duration**: prewarm hit-rate as the wf-side DX number (a counter over real launches, surfaced locally), leaving all wall-clock to devlaunch's trend.

Also settle: does #121's open fog item ("is wf's own startup worth attention — never profiled") get answered by this map's first measurements, and fed back there?

blooop:
**breadcrumb:** grilling started. Code facts established: prewarm is fire-and-forget at staging (app.rs:949, only 'fired' is recorded), so 'hit' is unobservable by wf — the falsifying event lives in dl, same shape as #128. Two human pauses sit inside keypress→exec (mode/steer overlay, checkout picker). timing.py's epoch excludes interpreter startup, so a final-keystroke t0 measures exec+interpreter — real, unmeasured cost.

blooop:
## Resolution

Grilled 2026-08-14. Three decisions, the first two grounded in code facts established at the start (breadcrumb above):

[...]

PR #162, opened 2026-08-14
wf stamps the launch t0 and hands it across the exec seam

### LINKED ISSUE #160: wf stamps the launch t0 and hands it across the exec seam
## Question

Build: wf stamps the launch t0 and hands it across the exec seam.

The instant a human commits to a launch (the keypress that resolves to an exec of `dl`), wf records a monotonic-honest wall-clock t0 and passes it to the child per the seam contract owned by the devlaunch vocabulary ticket (blooop/devlaunch#194) — an env var, set only on real launches. dl reports the `handoff` stage from it; a launch without the stamp has no handoff stage.

To carry with it (same seam, near-free once the stamp exists): whether the launch **hit the prewarm** — the single biggest felt-latency discriminator wf knows and nothing records. Whether that rides the same env var, a second one, or the `ctx:` block is this ticket's main design point; the `ctx:` block is claim-free by charter (#122), and "prewarm hit" is a claim about the past that stays true, so it is admissible there — but the consumer is a *timing* reader, not the agent, which argues for the env seam.

Blocked by the devlaunch vocabulary ticket: the variable name and semantics are its to mint. Do not invent them here first.

Seams: the launch-path tests pin the stamp's presence on a launch exec and its absence everywhere else; the docs guard extends to the seam contract the same way the ctx: docs guard works (#133's pattern).

blooop:
Closes #160

`wf` now stamps what it did on the way to a launch and hands it to the `dl` it becomes: `DEVLAUNCH_HANDOFF_T0` (the keystroke that resolved to this exec) and `DEVLAUNCH_PREWARM_FIRED_AT` (when staging fired this node's `dl <ws> up`, if it fired one), both Unix epoch seconds in `date +%s.%N`'s spelling, in the environment of the exec'd process. The names and the format are devlaunch#194's to mint; this side is the writer.

Per #161's resolution and the ticket's spec amendment: **the prewarm travels as its firing time in a second variable, never as a hit claim and never in the `ctx:` block.** `wf` fires a prewarm and forgets it, so whether the warm-up helped is visible only to the `dl` that then had to run the launch — hit, partial and miss are its to derive from the arm it took. An absent stamp is an absent measurement, never a zero.

### Where each stamp is minted

- **The keystroke** in the picker's launch arm — the one place a keypress has already resolved to an exec and nothing is left waiting on a human. Taken *before* the loader shutdowns, so `wf`'s own hand-over cost (the shutdowns, the cache write, the skill refresh, the exec) falls inside the gap the far side measures rather than being lost between the two programs.
- **The prewarm's firing time** at the spawn, not at the claim. A node whose launch is host-only is still claimed — it plans nothing now and would plan nothing later — but recording an instant for it would name a `dl <ws> up` that never went out. The session's record therefore grew from "which workspaces were prewarmed" to "what happened to each, and when".

### Only where the exec *is* a `dl`

A host launch execs the agent itself. Nothing there reads these, and one left standing in an agent session's environment would be picked up by every unrelated `dl` that session went on to run, each reporting a hand-over measured from a keystroke hours old. That arm carries neither variable and *clears* both, so a stamp `wf` inherited cannot travel on under `wf`'s name.

### Tests, at the seams the ticket named

- **The launch path's stamping decision** — asserted as name→value pairs rather than through a spawn, so absence is as checkable as presence: the keystroke travels, a prewarm nobody fired does not, a host launch carries nothing, and the values are instants and nothing else.
- **The picker's launch arm** — its two existing launch probes (a `dl` above the floor, one below) now also assert that the isolated run stamps an instant inside the run, and that the host run stamps nothing.
- **End to end**, `tests/live_launch_exec.rs` claim 6: the run opts into the prewarm, and the `dl` shim records the *environment* it was handed — the only place this is observable at all, since a stamp is not in an argv and the exec is where it is applied. Both stamps are instants inside the run, the prewarm's precedes the keystroke's (the two-step launch, seen from the far side of the exec), and the run's non-launch `dl` invocations — the `--version` probe and the prewarm's own `up` — carry neither.
- **The docs guard**, #133's pattern: the seam's variable names are read off the binary and compared to the README's in both directions, and the claim-free rule is pinned by equality on its own sentences. Mutating either side turns them red.

### Checked

`cargo fmt --check`, `clippy --all-targets --all-features`, `--lib --bins --examples`, `skill_docs`, `devcontainer_prebuild`, `cargo doc`, all under `-D warnings`; `pixi run suite` (hermetic, no `dl` on PATH); and the live binaries that cover what changed — `live_launch_exec`, `live_fetch`, `live_discovery`, `live_streaming_startup`. `live_devlaunch` needs the pixi devlaunch environments and is left to the contract workflow, which this diff does not touch the argv side of.

## Summary by Sourcery

Stamp keystroke and prewarm timing information into the environment of devlaunch execs so dl can measure handoff timing, while ensuring host launches carry no stamps and documenting the seam contract.

New Features:
- Track per-workspace prewarm outcomes and firing instants within the session and expose them to launches for inclusion in exec environment stamps.
- Introduce explicit handoff metadata for launches, including keystroke and prewarm timestamps, and propagate these as DEVLAUNCH_* environment variables only to devlaunch-based execs.

Enhancements:
- Refine prewarm bookkeeping from a simple set to a richer map that distinguishes warmed workspaces from those with no container to warm.
- Expose Launch.workspace and add helper APIs to compute and serialize handoff stamps consistently for both runtime and documentation checks.
- Strengthen picker and launch integration so the keystroke instant is captured before shutdowns and exec, ensuring `wf`'s own handover cost is measured.

Documentation:
- Add README and AGENTS documentation describing the exec seam environment stamps, their meaning, and the claim-free and host-launch-clearing rules.

Tests:
- Extend unit and live exec tests to cover handoff stamping behavior, prewarm recording semantics, absence-on-host and non-launch invocations, and to guard README seam documentation against drift.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -138,6 +138,18 @@ call for those tests (a machine with devlaunch installed and one without must
 not take different paths through the same test) and it leaves the fixtures
 unchecked against the program they describe. They have been wrong twice.
 
+There is a fifth thing `wf` hands `dl`, and it is not an argument: the launch
+`exec` sets two environment stamps of what `wf` itself did — the keystroke that
+resolved to the exec, and when this node's prewarm fired, if one did (#160,
+names and format minted by devlaunch#194). Two rules go with them. They are set
+only where the exec *is* a `dl`, so a host launch carries neither and clears
+both rather than leaving an inherited one in an agent's environment; and neither
+is ever a claim about how the launch went — a hit, a partial and a miss are
+`dl`'s to observe from the arm it takes, and `wf` is gone by then. The seam is
+observable only through a real exec, so its end-to-end guard is claim 6 of
+`tests/live_launch_exec.rs`, and the variable names the README publishes are
+compared to the ones the binary sets in `tests/skill_docs.rs`.
+
 `pixi.toml` is here for that and nothing else: it installs a chosen `devlaunch`
 and `tests/live_devlaunch.rs` asks it the questions `wf` asks it. **Pixi does
 not build this crate** — the compiler is still rustup's, and adding `rust` to
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -143,7 +143,8 @@ There is a fifth thing `wf` hands `dl`, and it is not an argument: the launch
 resolved to the exec, and when this node's prewarm fired, if one did (#160,
 names and format minted by devlaunch#194). Two rules go with them. They are set
 only where the exec *is* a `dl`, so a host launch carries neither and clears
-both rather than leaving an inherited one in an agent's environment; and neither
+both rather than leaving an inherited one in an agent's environment — as does
+every other `dl` `wf` starts, which is what `launch::unstamped` is for; and neither
 is ever a claim about how the launch went — a hit, a partial and a miss are
 `dl`'s to observe from the arm it takes, and `wf` is gone by then. The seam is
 observable only through a real exec, so its end-to-end guard is claim 6 of
--- a/README.md
+++ b/README.md
@@ -997,7 +997,10 @@ A **host** launch carries neither, and clears both rather than passing on one it
 inherited. The reader is the `dl` a launch becomes; a host launch becomes the
 agent instead, and a stamp left standing in an agent session's environment would
 be picked up by every unrelated `dl` that session went on to run, each reporting
-a hand-over measured from a keystroke hours old.
+a hand-over measured from a keystroke hours old. The same clearing covers every
+`dl` `wf` starts and does not become — the version probe, the prewarm's `dl <ws>
+up`, and `wf reap`'s listing and removals — because `wf` run inside a workspace
+starts with the stamps of the launch that built it already in its environment.
 
 ### `wf reap` — clearing away finished tickets
 
--- a/README.md
+++ b/README.md
@@ -967,6 +967,38 @@ Container lifecycle is `dl`'s: `wf` never stops or rebuilds one, and could not
 during a session — it `exec`s and is gone, so there is no `wf` left to observe
 an agent exiting. `dl <ws> stop`, `dl <ws> rm` and `dl --ls` are yours to run.
 
+### What a launch tells `dl` about itself
+
+**An isolated launch hands `dl` two stamps of what `wf` itself did**, in the
+environment of the process it becomes:
+
+| variable | what it stamps |
+| --- | --- |
+| `DEVLAUNCH_HANDOFF_T0` | the keystroke that resolved to this exec — the moment nothing is waiting on a human any more |
+| `DEVLAUNCH_PREWARM_FIRED_AT` | when staging fired this node's `dl <ws> up`, if it fired one |
+
+Both are Unix epoch seconds, spelt the way `date +%s.%N` prints them, and the
+names and the format are `dl`'s to mint rather than `wf`'s — `wf` is the writer,
+and a writer that invented its own names would be handing the reader nothing.
+The gap from `DEVLAUNCH_HANDOFF_T0` to `dl`'s own start is the only measurement
+of the exec there is, since everything either side of it is inside one program
+or the other.
+
+**Neither stamp is a claim about how the launch went.** `wf` fires a prewarm and
+forgets it — the process that fired it is gone long before the container is
+ready — so whether the warm-up helped, was still running, or saved this launch
+nothing is visible only to the `dl` that then had to run the launch, and that is
+`dl`'s to report from the arm it took. What `wf` sends is when it acted, twice,
+and nothing else. An absent stamp is an absent measurement and never a zero: a
+launch nobody prewarmed sets no `DEVLAUNCH_PREWARM_FIRED_AT`, rather than
+setting it to something that would read as an instant warm-up.
+
+A **host** launch carries neither, and clears both rather than passing on one it
+inherited. The reader is the `dl` a launch becomes; a host launch becomes the
+agent instead, and a stamp left standing in an agent session's environment would
+be picked up by every unrelated `dl` that session went on to run, each reporting
+a hand-over measured from a keystroke hours old.
+
 ### `wf reap` — clearing away finished tickets
 
 A workspace per ticket means workspaces accumulate as fast as tickets are
--- a/src/app.rs
+++ b/src/app.rs
@@ -9,6 +9,7 @@
 
 use std::cmp::Reverse;
 use std::collections::{BTreeMap, BTreeSet};
+use std::time::SystemTime;
 
 use ratatui::crossterm::event::{KeyCode, KeyEvent, KeyModifiers};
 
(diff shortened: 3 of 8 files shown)


blooop (commented):

Two-axis review at fixed point `952b56b` (merge-base with main); diff `git diff 952b56b...wayfinder/wayfinder-160`. Preflight: all checks green (incl. the devlaunch-contract job), mergeable, diff non-empty. Spec = ticket #160 including the 2026-08-14 amendment comment (second env var carries the firing time; never in `ctx:`; never a hit claim).

## Standards

Repo standards (AGENTS.md house style — reasoning in doc comments, illegal states unrepresentable, no bool that can disagree with a sum) are met. The new types are total: `Handoff` (`launch.rs:1423`) has no invalid combination, `stamps` (`launch.rs:1791`) returns a `None` per variable and `exec` applies it as `env_remove`, so the host arm cannot leak an inherited stamp, and `Prewarmed`/absent-from-map covers three real states with no flag beside them. No boolean blindness reintroduced.

- **`src/app.rs:1010` — Mysterious Name (non-blocking).** `claim_prewarm` no longer claims: it took `&mut self` and inserted; it now takes `&self` and only reports a candidate. Two consecutive calls both return `Some`; the once-per-node rule is enforced solely by `warm`'s insert at `app.rs:996`. The name and its doc ("the once-per-node rule … testable") still describe the old contract.
- **`src/app.rs:972` and `:989` — Middle Man + Speculative Generality (non-blocking).** `prewarm` and `warm` are near-synonyms on the same type, the former a one-line pass-through whose only content is reading the env gate. And `warm`'s `Option<Prewarmed>` return is discarded at both production call sites (`app.rs:882`, `:949`) — it exists only so the tests can observe the record.
- **`src/app.rs:320` — doc rationale vs. house style (non-blocking).** `Prewarmed`'s comment defends itself against "an `Option<SystemTime>` beside a claim flag", which nothing proposed; the real alternative is `BTreeMap<String, Option<SystemTime>>`. The type does earn its weight (it stops `get` yielding a nested `Option`) — the doc just never says the thing that justifies it, which is what AGENTS.md asks doc comments to carry.
- **Shotgun Surgery on the seam's prose (non-blocking).** The claim-free rule and the host-clears-both rule are each restated in `AGENTS.md:141`, `README.md:970-1000`, `src/launch.rs:1406` (`Handoff`), `src/launch.rs:1779` (`stamps`), and comments in `src/launch.rs:4085`, `src/picker.rs:826`, `tests/live_launch_exec.rs:310` — with two README paragraphs pinned verbatim as string literals in `tests/skill_docs.rs:557`. Reworded once, that is a seven-site edit with a test failure gating it.
- **`tests/live_launch_exec.rs:310` — Duplicated Code (non-blocking).** `dl_asides` is the exact negation of `dl_launches` (`:287`), re-spelling the `" -- "` predicate; one extracted `is_launch(record)` would keep the two definitions from parting.
- **`src/picker.rs:170` — legibility nit (non-blocking).** `Handover(Box<Launch>, Handoff)` threads an unnamed pair through three destructurings; a struct variant would name both halves at each site.

Standards verdict: 6 non-blocking

## Spec

**"wf records a monotonic-honest wall-clock t0 and passes it to the child per the seam contract ... set only on real launches"** — **met**. `src/launch.rs:1423-1476` mints `DEVLAUNCH_HANDOFF_T0` from `SystemTime::now()`; `src/launch.rs:1791-1804` returns it only for `Isolation::Devlaunch`; `src/launch.rs:1855-1868` applies it to the child `Command`, never to `wf`'s own env.

**Format satisfies dl's actual parser** — **met**. `epoch_seconds` (`src/launch.rs:1472`) emits `{secs}.{nanos:09}`. dl's `_stamp`/`_read_stamps` do `float(strip())` + `isfinite` and need a non-negative gap; t0 is taken before the shutdowns/exec so the gap is positive, and `head_start = keystroke − prewarm_fired ≥ 0` holds because the prewarm fires on the first enter and t0 on the second (`tests/live_launch_exec.rs` claim 6 asserts the ordering).

**"the prewarm information travels as a second env var carrying the firing time, never a hit claim and never in the `ctx:` block"** — **met**. `PREWARM_FIRED_VAR` (`src/launch.rs:1463`) carries an instant only; nothing touches the `ctx:` block; no hit/partial/miss appears on the wf side.

**"wf stamps only its own past actions"** — **met**. `src/picker.rs:406` stamps in the launch arm before `loaders.shutdown()`; `src/app.rs:994` records `Prewarmed::At` only where an argv was actually spawned; `prewarm_fired` (`src/app.rs:1023`) maps `Nothing`/absent to `None`.

**"tests pin the stamp's presence on a launch exec and its absence everywhere else"** — **partially met, and this is the blocking finding**. Presence/absence is pinned at three seams, but absence on wf's *non-launch* `dl` children is only true of a clean environment: `spawn_detached` (`src/launch.rs:2055`) and the `dl --version` probe inherit `wf`'s env verbatim, so an inherited `DEVLAUNCH_HANDOFF_T0`/`DEVLAUNCH_PREWARM_FIRED_AT` rides onto the prewarm `dl <ws> up` and the probe. Only the launch `exec` clears. This is the exact hazard the PR body's own deviation 1 names ("every unrelated `dl` that agent ran would report a fictional handoff stage") — closed for the exec path, left open for the detached children.

**"the docs guard extends to the seam contract (#133's pattern)"** — **met**. `tests/skill_docs.rs` adds a two-directional name check off `Handoff::variables()` plus two whole-paragraph equality pins.

**Declared deviations:** (1) stamps only on dl execs, `env_remove` otherwise — licensed; (2) t0 minted in the picker's launch arm — licensed; (3) docs guard on README — licensed; (4) `claim_prewarm` reshaped — unlicensed but the direct consequence of "staging remembers *when* it fired". All PR-body claims verified; one imprecision: `Prewarmed::At(SystemTime::now())` is taken after `spawn_detached` returns, a sub-millisecond conservatism.

Spec verdict: gaps: [inherited stamps pass through `spawn_detached` to wf's non-launch `dl` children]

## Verdict

**Request changes.** One blocking finding: wf's non-launch `dl` children (the prewarm `up` via `spawn_detached`, the `dl --version` probe) inherit `DEVLAUNCH_HANDOFF_T0`/`DEVLAUNCH_PREWARM_FIRED_AT` from wf's own environment instead of clearing them — the fictional-handoff hazard the PR body itself names, enforced in tests only under a clean environment. Fix: scrub both variables (`env_remove`) on every `dl` child wf starts that is not the launch exec, and pin it with a test that runs wf with the stamps pre-exported. The six Standards findings and deviation 4 remain non-blocking.

blooop (commented):

**Re-review of the blocking finding after `53175c7`: resolved — Approve.**

The scrub is a single constructor (`launch::unstamped`, src/launch.rs:1465) reading the names off `Handoff::variables()` — no new literal sites — and it now builds every dl child wf does not become: the `--version` probe, `spawn_detached`'s `sh` (prewarm `up`), and `wf reap`'s `--ls --json` and `rm`, the last two a third leaking site the fix's red test surfaced beyond the review. Pinned twice: the pty run pre-exports both stamps so the `dl_asides` assertion is now about wf rather than about a clean rig (verified red-then-green), plus a hermetic unit test asserting the child's whole env delta is exactly the two removals (`get_envs`), which runs in CI where the live tests are compile-only. `claim_prewarm` → `workspace_to_warm` rider included; other non-blocking findings deliberately left. CI green on head `53175c7`.

**Verdict: Approve.**

PR #164, opened 2026-08-15
A bare cargo test is green, and the live tests have a runner

### LINKED ISSUE #42: Make cargo test pass inside the container
## Question

Make `cargo test` pass inside the devcontainer, and decide what the live tests become.

This repo's tests are not hermetic and they are not `#[ignore]`d, so they run by default and they reach outside:

- `tests/live_fetch.rs` hardcodes `blooop/wayfinder` map **#1** and needs network plus an authenticated `gh`. Read-only, so N containers running it at once is safe — confirm that holds.
- `tests/live_zellij_launch.rs` starts **real zellij sessions** (20 references). Inside a container this should be *better* than on the host, where two concurrent runs collide on session names — verify that the container's PID/socket namespace actually delivers that, and that zellij runs at all with no TTY.
- `live_discovery.rs`, `live_poll.rs`, `live_streaming_startup.rs` — establish what each needs (a git checkout with a GitHub `origin`? a writable `~/.cache/wf`?) and whether a devpod workspace's checkout satisfies it.

Decide: do these stay as-is, gain `#[ignore]`, or split into a hermetic set plus a live set behind a feature or env gate? Whichever way, `cargo test` inside a fresh workspace must end green and the reason must be recorded.

blooop:
Closes #42

A bare `cargo test` in a fresh checkout was red — the contract tests failed
closed with 15 panics — because nothing separated the tests that need only a
compiler from the 25 that need network, an authenticated `gh`, a pty, or a
chosen `devlaunch`. Built to the decisions in the ticket's grilling resolution,
not to the ticket body: two of the five files it names (`live_zellij_launch.rs`,
`live_poll.rs`) were deleted in 9ae2897, so the whole zellij branch of the
question was moot.

## What changed

- **All 25 live tests get `#[ignore]` with a per-file reason string.** An
  env-var skip was rejected because a skip that reports as a pass is a green
  nobody can trust; a feature gate was rejected because CI's
  `cargo test --all-targets --no-run` is the only thing keeping these tests from
  bit-rotting and a gated-out test compiles nowhere. Test bodies are untouched —
  the assertions about the real tracker stay exactly as they were (decision 3:
  asserting reality is the point of a live test).
- **The pixi `contract` tasks pass `--ignored`.** They are the callers that
  supply what `live_devlaunch` needs, so they are the ones that ask for it.
  Worth naming the near-miss: without the flag those tasks select nothing and
  libtest prints `ok` for the empty selection, so `devlaunch-contract.yml` would
  have stayed green while checking nothing.
- **New `.github/workflows/live.yml`** runs the eight gh-live tests on push to
  `main` and on `workflow_dispatch`, serially, with `GH_TOKEN: ${{ github.token }}`
  — every `gh` call in this repo is a read-only GET, so no secret is needed.
  Deliberately not on pull requests: two of these tests assert the tracker's
  present contents and one asserts wall-clock budgets, so a red run means the
  world moved, which should not block an unrelated merge. `live_devlaunch` stays
  out of it — it has `devlaunch-contract.yml`.
- **The devcontainer says why zellij is in it, truthfully.** Its justification
  cited `tests/live_zellij_launch.rs`, deleted a while ago. Zellij stays, for
  interactive use in a workspace, and the comment now says that.
- **README and AGENTS.md** state what `cargo test`, `cargo test -- --ignored`
  and each of the two live workflows covers — including that `-- --ignored`
  selects both halves of the gated set and only one half can pass on a given
  machine.

## Verified locally

| check | result |
| --- | --- |
| `cargo test --locked` (bare) | green, **32 ignored** — see the note below |
| `cargo test --locked --all-targets --no-run` | compiles, bit-rot guard intact |
| `ci.yml`'s `--test live_fetch -- common::` step | green, 3 passed / 1 filtered |
| the whole AGENTS.md check chain, `-D warnings` on both | green |
| `pixi run lint`, `pixi run suite` | green |
| `pixi run -e default contract` | **17 passed** — the `--ignored` change proven end to end |
| the 4 gh-live files, `-- --ignored --test-threads=1` | **8 passed, 0 failed** (rc=0) |

**The ignored count is 32, not the 25 the resolution predicted.** Seven tests in
`src/` were already `#[ignore]`d before this branch — the probe children that
only mean anything under recording shims. 25 of the 32 are the live set this PR
gates; nothing was double-counted and nothing else changed. Those seven no-op
under `cargo test -- --ignored` (they check `probe::is_child()` first), so
documenting that command as a way to reach the live set is safe.

**One environmental failure, pre-existing and not fixed here.**
`pixi run -e floor contract` fails one of its 17 on the machine this was built
on: `the_dl_under_test_is_the_one_the_environment_installed` compares the
resolved `dl` against `CONDA_PREFIX`, and on this host `cargo` is a pixi-global
trampoline (`~/.pixi/bin/cargo`) that overwrites `CONDA_PREFIX` with its own
`rust` environment for every process below it. The `dl` under test is in fact
the right one. Reproduced identically with `main`'s `tests/live_devlaunch.rs`
and `main`'s command form, so it predates this branch; CI takes `cargo` from
rustup rather than a trampoline, which is why `devlaunch-contract.yml` is green
there. Weakening the assertion would have been the wrong fix — it is the one
that makes the version pin mean anything.

## Post-merge acceptance step

`workflow_dispatch` only becomes available once `live.yml` is on the default
branch, so "green on a manual dispatch" — the last line of the resolution's
acceptance list — cannot be demonstrated from this PR. The push-to-`main`
trigger fires on merge and is the first real run either way.

## Summary by Sourcery

Gate all live integration tests behind #[ignore] and add targeted workflows and documentation so default test runs are hermetic while still exercising live paths in CI and via pixi contract tasks.

New Features:
- Add a dedicated live GitHub workflow that runs the gh-based integration tests against the real tracker on pushes to main and manual dispatch.
- Document the behavior and usage of live tests and gated test runs in README and AGENTS, including how cargo test and cargo test -- --ignored split the suite.

Bug Fixes:
- Ensure pixi devlaunch contract tasks actually execute the live_devlaunch tests by selecting ignored tests instead of silently running nothing.
- Make a bare cargo test in a fresh checkout or devcontainer succeed by marking all live tests as ignored with explicit reason strings.

Enhancements:
- Extend CI test coverage by explicitly running skill_docs and devcontainer_prebuild tests, and adding a compile-only all-targets no-run step as a bit-rot guard.
- Clarify the devcontainer Dockerfile rationale for shipping gh, the agent CLI, and zellij, decoupling zellij from now-removed tests and framing it as an interactive workspace tool.

Documentation:
- Update README and AGENTS to explain CI coverage, the hermetic default test suite, the gated live test set, and the responsibilities of live.yml and devlaunch-contract.yml.

Tests:
- Mark all live GitHub and devlaunch contract tests with #[ignore] annotations that describe their environmental requirements, keeping their bodies intact while controlling when they run.
- Run live devlaunch contract tests serially via pixi tasks and ensure they are only executed in environments that satisfy the WF_CONTRACT_* requirements.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -9,10 +9,22 @@
 # could save. If that number ever grows into the minutes, this is the file to
 # revisit — see blooop/wayfinder#39 for the measurements.
 #
-# What it *does* carry is the three binaries `wf` shells out to and discovers on
-# PATH — `gh`, `zellij`, and the agent CLI — because `cargo test` here is not
-# hermetic: tests/live_fetch.rs drives `gh` against the real tracker and
-# tests/live_zellij_launch.rs starts real zellij sessions.
+# What it *does* carry is three binaries, for two different reasons.
+#
+# `gh` and the agent CLI are things `wf` itself shells out to and discovers on
+# PATH, so a workspace without them cannot run `wf` at all. They are not here
+# for the tests: `cargo test` in here is hermetic and stays green with no
+# network, because every test that reaches a real GitHub is `#[ignore]`d with a
+# reason (see the Checks section of README.md). Running that gated set inside a
+# workspace is a thing you can choose to do — `cargo test -- --ignored` — and
+# `gh` being present is what makes the choice available.
+#
+# `zellij` is here for a different reason and it is worth saying so, because
+# this comment used to justify it by a test file that no longer exists: `wf`
+# stopped launching zellij sessions in 9ae2897 and no code or test in this repo
+# names it now. It stays for interactive use — a workspace is somewhere a
+# person works, and splitting the terminal inside one is worth carrying a
+# single static binary for.
 FROM mcr.microsoft.com/devcontainers/rust:1-bookworm
 
 # Pinned rather than "latest" so an image rebuild is a deliberate version bump
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,11 @@
 name: CI
 
 # Formatting, lints, tests and docs — everything that can be judged with
-# nothing but a compiler and this checkout. Two other workflows cover what
-# cannot: `package.yml` proves the recipe still installs and runs, and
+# nothing but a compiler and this checkout. Three other workflows cover what
+# cannot: `package.yml` proves the recipe still installs and runs,
 # `devlaunch-contract.yml` runs `wf` against a real `dl` at three chosen
-# versions. All three are kept apart on purpose, and for one reason — a rattler
+# versions, and `live.yml` runs the gh-live tests against the real tracker
+# after a merge. All four are kept apart on purpose, and for one reason — a rattler
 # build and a conda solve each take minutes and download a toolchain of their
 # own, so folding `cargo fmt` into either would make the cheapest possible
 # failure the slowest one to hear about.
@@ -116,8 +117,8 @@ jobs:
         run: "cargo test --locked --test live_fetch -- common::"
 
       # Compiled but not run: a test that no longer builds is rot, and this is
-      # what catches it at the commit that caused it rather than the next time
-      # somebody runs them by hand.
+      # what catches it at the commit that caused it rather than at the next
+      # `live.yml` run after a merge.
       - name: The live tests still compile
         run: cargo test --locked --all-targets --no-run
 
--- a/.github/workflows/live.yml
+++ b/.github/workflows/live.yml
@@ -1,0 +1,88 @@
+name: live
+
+# The four test binaries that need a real GitHub: `wf`'s fetch, discovery,
+# streaming startup and launch paths, run against this project's own tracker
+# rather than a fixture. `ci.yml` compiles them and never runs them; without
+# this workflow nothing ever did, and the only thing standing behind them was
+# somebody remembering to run them by hand.
+#
+# `tests/live_devlaunch.rs` is the fifth gated binary and is deliberately not
+# here: it wants a chosen `devlaunch` from a pixi environment, which is
+# `devlaunch-contract.yml`'s whole job.
+#
+# ## Not on pull requests, on purpose
+#
+# This is a non-blocking leg and the trigger list is what makes it one. Two of
+# these tests assert the tracker's present contents — that this repo has more
+# than one open map, that the first one has at least seven tickets — and
+# `live_streaming_startup` asserts wall-clock budgets, which are the first
+# thing to flake on a shared runner under a noisy neighbour. All three are
+# right to assert what they assert (asserting reality is the point of a live
+# test), and none of them is a reason to block an unrelated merge. Running
+# after the merge instead means a red run here says "the world moved" at a
+# moment when that is information rather than an obstacle.
+on:
+  push:
+    branches: [main]
+  # The trigger a human pulls after changing the fetch or launch path, when
+  # they want the answer before the merge rather than after it. Same reason
+  # `ci.yml` carries one, plus this one: a leg that only ever fires on `main`
+  # is a leg nobody can aim at a branch.
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  # As in `ci.yml`. A superseded run on `main` has nothing to say that its
+  # successor will not say better, and the tests here are the slow ones.
+  group: live-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  RUSTFLAGS: -D warnings
+  CARGO_TERM_COLOR: always
+
+jobs:
+  live:
+    name: the live tests, against the real tracker
+    runs-on: ubuntu-latest
+    env:
+      # Every `gh` call this repo makes is a read-only API GET, so the token
+      # the workflow is handed anyway is enough and no secret needs to exist
+      # for this to run. `tests/common` turns a missing one into a diagnostic
+      # rather than a confusing failure, which is why that module is the one
+      # part of `tests/` `ci.yml` already runs.
+      GH_TOKEN: ${{ github.token }}
+    steps:
+      # The checkout is load-bearing beyond supplying the source:
+      # `live_discovery` and `live_launch_exec` resolve this directory's git
+      # `origin` and require it to be `blooop/wayfinder`, so they are testing
+      # the runner's own checkout as much as anything.
+      - uses: actions/checkout@v7
+
+      - name: Install the Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache cargo's registry and the target directory
+        uses: Swatinem/rust-cache@v2
+        with:
+          key: live
+
+      # `--ignored` is how the gated set is selected at all: each of these
+      # tests carries an `#[ignore]` naming what it needs, so a bare
+      # `cargo test` skips them and this is the caller that supplies it. Named
+      # binary by binary rather than `--tests`, so adding a live binary is a
+      # deliberate line here instead of a silent enrolment.
+      #
+      # `--test-threads=1` because `live_streaming_startup`'s budgets are
+      # wall-clock: run beside the pty harness on two shared cores they measure
+      # the runner's load rather than `wf`'s startup.
+      - name: The live tests
+        run: >-
+          cargo test --locked
+          --test live_fetch
+          --test live_discovery
+          --test live_streaming_startup
+          --test live_launch_exec
+          -- --ignored --test-threads=1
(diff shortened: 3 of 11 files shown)


blooop (commented):

Two independent fresh-context review axes (Standards, Spec), run in parallel; findings are not merged or reranked across axes. Fixed point: merge-base `f402b2e` with `main`; diff reviewed as `git diff f402b2e...fbe0620`. Spec = ticket #42 **plus** its grilling-resolution comment (5301862240), which overrides the ticket body where they differ.

## Standards

Verified against a fresh clone: bare `cargo test --locked` is green offline — **478 passed / 32 ignored, exit 0** — and all 25 `#[ignore]`s land where claimed (17 `live_devlaunch`, 4 streaming, 2 launch_exec, 1 each fetch/discovery); every `#[test]` in `tests/live_*.rs` carries one, and the 7 `src/` probe children are guarded by `probe::is_child` (`src/probe.rs:588`), so `-- --ignored` doesn't misfire. Dockerfile claim checks out: `9ae2897` is "delete zellij", and no code or test names it; the interactive-use justification is honest and the deleted-test citation is gone.

Findings:

- **`tests/live_streaming_startup.rs:57-59`** — comment still says "this file is compiled but never run by CI, which makes a wildcard in it the least observed thing in the tree". `live.yml` is precisely what falsifies that. *non-blocking*
  *[Resolved during review — orchestrator commit efed34e: now "runs only after a merge (`live.yml`), which would make a wildcard in it the slowest-observed thing in the tree".]*
- **`.github/workflows/ci.yml:4-10`** — the header enumerates its siblings ("Two other workflows cover what cannot… All three are kept apart on purpose"); `live.yml` is now a fourth and isn't named, though `live.yml` names `ci.yml` in both directions. Same for `ci.yml:119-120` ("the next time somebody runs them by hand"). *non-blocking*
  *[Resolved during review — orchestrator commit efed34e: header now censuses all four workflows; the compile-guard comment points at `live.yml`'s post-merge run.]*
- **`AGENTS.md:103`** — "CI runs four things" while `ci.yml` now runs eight steps; README's parallel sentence was fixed in this diff, AGENTS.md's wasn't. *non-blocking*
  *[Resolved during review — orchestrator commit efed34e: countless phrasing ("the checks below") so it cannot rot again.]*
- **`.github/workflows/live.yml:29`** — `permissions: contents: read` leaves `issues` at none, but every call is a GraphQL issues/search read. Worth confirming before merge: the leg only fires post-merge, so a wrong guess surfaces late and red. *non-blocking, verify*
  *[Resolved during review — orchestrator commit efed34e: `issues: read` granted explicitly, with a comment saying why it is granted rather than gambled on.]*
- **`AGENTS.md:153` / `README.md:190`** — "four devlaunch versions": four *environments*, three versions (`default` has none, which is the point of that feature); `ci.yml:6` says three. *nit*
  *[Resolved during review — orchestrator commit efed34e: both now say "four pixi environments — three pinned versions plus the unpinned `default`".]*
- **`live.yml:16`** — "Two of these tests assert the tracker's present contents"; it's three (`live_discovery.rs:20`, `live_streaming_startup.rs:66`, `live_fetch.rs:31`), as `tests/common/mod.rs:41-46` itself lists. *nit*
  *[Resolved during review — orchestrator commit efed34e: "Three of these tests… and `live_streaming_startup` asserts wall-clock budgets besides"; README's parallel sentence fixed the same way.]*
- **Reason strings** — "asserts wall-clock budgets" is true of 2 of the 4 in `live_streaming_startup` (`:134`, `:209`); `live_fetch.rs:15` omits the mature-map precondition its own module doc states. *nit*
  *[Accepted with reason: the strings are the resolution's own suggested wording, verbatim, and a reason string names the file-level requirement class, not each assertion; the module docs carry the detail.]*
- **`concurrency:`** — sound, keep. Mirrors `ci.yml`, and superseding is right for tests asserting present reality. Unlike `devlaunch-contract.yml:40` the group omits `event_name`, so a dispatch on `main` cancels an in-flight push run — same question, later answer wins.
- **Declined guard test — recommend the ticket.** Nothing in CI runs the live binaries unfiltered (`ci.yml` stops at `--lib --bins --examples` plus two named binaries and `-- common::`; the AGENTS chain matches), so a live test added without the attribute breaks the offline-green promise with no check anywhere, and with it missing also never runs in `live.yml`. Precedent is strong (`devcontainer_prebuild.rs`, and `live_devlaunch.rs:1583` already guards its own source). Not over-machinery.
  *[Resolved during review — filed as #165.]*
- **Doc duplication** — the 25/8/17 split is narrated in README, AGENTS.md, `live.yml`, `pixi.toml` and the Dockerfile; the counts appear in two. Shotgun surgery on the next live test. *non-blocking*
  *[Accepted with reason: each site addresses a different reader (workflow author, contributor, task runner), and #165's guard will catch the count that matters going stale; consolidating the prose is its own churn for no behavior.]*

Axis verdict: **Comment** (all findings resolved in-review, ticketed, or accepted with reason; nothing blocking remains)

## Spec

**Verdict: compliant.** Every one of the resolution's five decisions is implemented to the letter. Reproduced independently in a fresh clone at `wayfinder/wayfinder-42` (merge-base `f402b2e`):

- **(a) Bare run.** `cargo test --locked` → **478 passed, 0 failed, 32 ignored, exit 0**, exactly the predicted number. Each live test prints its reason inline (e.g. `ignored, live: needs network + gh; asserts wall-clock budgets`), satisfying "green, with the skip visible, never a dishonest green". The diff adds exactly 25 `#[ignore]` lines (17 devlaunch / 4 streaming / 2 launch_exec / 1 fetch / 1 discovery), and the reason strings are verbatim the ones the resolution suggested per file. The other 7 are the pre-existing probe children in `src/` (4 lib + 3 bin) — untouched by this diff, and I confirmed they *pass* under `--ignored` rather than fail, so documenting that flag is safe.
- **(b) Bit-rot guard.** `cargo test --locked --all-targets --no-run` → exit 0, all five live binaries linked. Decision 1's stated reason for rejecting a feature gate still holds.
- **(c) Live legs.** `cargo test --locked --test live_fetch --test live_discovery -- --ignored --test-threads=1` → 2 passed, 0 failed against the real tracker (my clone's origin is `blooop/wayfinder`).
- **(d) Decision 3** — "Repo-state assertions stay as-is… test bodies untouched": every hunk under `tests/` is a single added attribute line. No body, assertion, or import changed.
- **live.yml vs decision 2's letter:** triggers are exactly `push: branches: [main]` + `workflow_dispatch` (no `pull_request`); `GH_TOKEN: ${{ github.token }}`; the four `--test` flags in the named order; `-- --ignored --test-threads=1`; `live_devlaunch` absent and its exclusion documented.
- **Decision 4:** zellij install retained; the comment no longer cites the deleted `live_zellij_launch.rs` and now justifies it as interactive use.
- **Decision 5:** README testing section and AGENTS.md both state what each command covers.
- **Contract tasks:** both pixi tasks gained `-- --ignored`; PR CI shows all four contract environments still `running 17 tests / 17 passed`.

Judgement calls, all **non-blocking**:
- `concurrency:` in live.yml is unrequested but mirrors `ci.yml`'s existing block; affects no decision.
- The README "Checks list" rewrite goes beyond the five decisions, but the old list was wrong (said "four things", omitted three steps). I diffed it against `ci.yml`: the new list matches all seven commands in order.
- **25 vs 32 ignored (nit):** acceptance said "25 ignored". Intent met, and AGENTS.md states the split explicitly ("25 of the 32… the other seven being the probe children"). README says only "there are 25 of them" without naming the 32 a reader will actually see — the one place the docs are less than fully explicit.
  *[Resolved during review — orchestrator commit efed34e: README now carries the 25-of-32 split at that sentence.]*
- One of five bare runs showed a single failure in the `src/` bins binary (~12s, outside this diff); did not reproduce in four further runs.

Axis verdict: **Approve**

## Verdict

**Approve.** Spec approves with all five decisions independently reproduced (bare run, bit-rot guard, two live legs, attribute-only diff hunks, workflow letter). Standards' findings were doc-accuracy class: resolved during review in orchestrator commit efed34e (hermetic suite re-run green: 478 passed / 32 ignored; `cargo fmt --check` clean), the missing structural guard filed as #165, two nits accepted with reason. No blocking findings remain. GitHub may refuse the approve state on a same-account review; this written verdict is what the gate reads.

Two notes for the maintainer, outside the diff: the builder hit a pre-existing environmental failure in `pixi run -e floor contract` on this host only (the pixi-global cargo trampoline overwrites `CONDA_PREFIX`; CI, which takes cargo from rustup, is green) — worth knowing before running contract envs locally. And GitHub is reporting one low-severity Dependabot alert on the default branch (dependabot/1), unrelated to this PR.

PR #166, opened 2026-08-15
the prebuild guard claims only what it holds

### LINKED ISSUE #158: The prebuild reference guard says 'and no other' but only holds the drift directions
Non-blocking residue from PR #152's re-review (ticket #150), recorded so it does not decay in a PR comment.

The rebuilt guard (822e554) holds every **accidental-drift** direction it exists for — renaming the real push argument, deleting the Push step, and adding a second `ghcr.io/` tag all go red. Its blind spot is a **deliberately added** publish to a non-ghcr registry: a `docker/build-push-action` step pushing to `docker.io/...` stays 7/7 green.

That is sabotage rather than drift, so the guard is doing its job — but the assert message ("and no other") and the doc comment claim more than the mechanism delivers. Either narrow the wording to the class actually guarded, or widen the guard to collect every publish-shaped step regardless of registry.

Two smaller nits from the same review ride along: the "login line" comment, and a substring version-match in the pin check.

blooop:
Closes #158

Per the grilling resolution on the ticket: **narrow the wording, do not widen the guard.**

The blind spot #152's re-review found — a deliberately added non-ghcr publish — is sabotage rather than drift, and sabotage takes push access, which is also enough to edit the test. Widening the tokenizer (continuation-aware parsing, `build-push-action` steps, buildx `--push`) would buy nothing durable, since the next publish shape escapes it again, while adding false-red surface on legitimate workflow refactors. So the claims move to meet the mechanism instead.

Three text changes and one behavioural one:

- The **"and no other"** assert message gains the limit it actually holds: accidental drift of the ghcr publish — a renamed argument, a deleted step, a second `ghcr.io/` tag. A publish deliberately added to another registry is stated as out of sight and out of scope.
- **`pushed_references`** claimed a push target "renamed to anything at all" still lands there. The continuation-split mutant (`docker \` newline `push …`) showed that is false as written, so the claim is scoped to renames of the existing `docker push` command's argument, with the token-matching limit spelled out.
- **`published_references`** claimed the sweep catches "a login line". It does not — the login names the registry host bare (`docker login ghcr.io`), no slash, so the `ghcr.io/` prefix never matches and the token is never collected. Corrected to what the sweep really sees, and pointed at `the_push_authenticates_with_the_runs_own_token`, which is what actually guards that step.
- **Rider, executable:** the Dockerfile pin check compared with a substring `case`, so a version merely *containing* the pin satisfied it — `2.1.2234` passed a pin of `2.1.223`, as did `12.1.223`. Replaced with word-boundary matching:

  ```sh
  printf %s "$installed" | grep -qwF "$CLAUDE_VERSION" \
    || { echo "pinned CLAUDE_VERSION=$CLAUDE_VERSION but installed: $installed" >&2; exit 1; }
  ```

  Boundary matching stays agnostic about the shape of the version line, which matters because the three tools pinned in this file each print a different one (`2.1.223 (Claude Code)`, `gh version 2.97.0 https://…`, `zellij 0.44.3`) — parsing a field out would be a fourth thing to keep in step with three upstreams.

## Verification

The pin check only runs inside `docker build`, so its new form was checked directly in a shell against both tools' real output shapes: the exact pin, a bare version, and a name-prefixed version all accept; `2.1.2234`, `12.1.223` and `2.1.22` all refuse. The old `case` form accepts the first two of those three, which is the bug.

All 7 prebuild tests stay green, and the full AGENTS.md chain (fmt, clippy, suite, skill docs, prebuild, doc — both `-D warnings` flags set) is clean. The guard was re-mutated to confirm the wording change left it real rather than vacuous: renaming the workflow's push target still reddens `the_workflow_publishes_exactly_the_reference_the_default_config_boots_from`.

As the resolution states, the two mutants that escaped #152's re-review are **not** expected to redden — the wording now says as much, which was the point.

## Summary by Sourcery

Clarify and narrow the devcontainer prebuild guard’s assertions to match what the workflow tokenizer actually observes, and tighten the devcontainer tool version pin check.

Bug Fixes:
- Ensure the devcontainer CLAUDE_VERSION pin check rejects versions that only contain the pinned value as a substring, preventing false positives for mismatched versions.

Enhancements:
- Refine pushed and published reference collection comments to accurately describe the scope and limitations of the workflow analysis, including exclusions like registry logins and non-docker-push publish mechanisms.
- Clarify the workflow publish assertion message so it explicitly scopes its guarantee to accidental drift of the ghcr publish rather than deliberate publishes to other registries.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -77,11 +77,12 @@ RUN set -eux; \
 #
 # Matched on a word boundary rather than as a substring, because a substring
 # match accepts a version it was never pinned to: `2.1.2234` contains `2.1.223`,
-# as does `12.1.223`. Boundary matching refuses both while staying agnostic about
-# the shape of the line the version arrives in — the three tools pinned in this
-# file each print a different one (`2.1.223 (Claude Code)`, `gh version 2.97.0
-# https://…`, `zellij 0.44.3`), so parsing a field out would be a fourth thing to
-# keep in step with three upstreams.
+# as does `12.1.223`. Boundary matching refuses both. It is not fully
+# shape-proof — `.` and `-` are non-word characters, so a `2.1.223-beta` would
+# still match, and a `v`-prefixed `v2.1.223` would be refused (loudly, the safe
+# direction) — but it stays agnostic about the line the version arrives in
+# (today: `2.1.223 (Claude Code)`), where parsing a field out would be one more
+# format to keep in step with the upstream installer.
 USER vscode
 RUN set -eux; \
     curl -fsSL https://claude.ai/install.sh | bash -s -- "$CLAUDE_VERSION"; \
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -74,13 +74,19 @@ RUN set -eux; \
 # argument would leave this pin buying nothing with the build still green. The
 # comparison also backstops the unpiped `curl | bash` — a truncated script that
 # installed something else fails here.
+#
+# Matched on a word boundary rather than as a substring, because a substring
+# match accepts a version it was never pinned to: `2.1.2234` contains `2.1.223`,
+# as does `12.1.223`. Boundary matching refuses both while staying agnostic about
+# the shape of the line the version arrives in — the three tools pinned in this
+# file each print a different one (`2.1.223 (Claude Code)`, `gh version 2.97.0
+# https://…`, `zellij 0.44.3`), so parsing a field out would be a fourth thing to
+# keep in step with three upstreams.
 USER vscode
 RUN set -eux; \
     curl -fsSL https://claude.ai/install.sh | bash -s -- "$CLAUDE_VERSION"; \
     installed="$(/home/vscode/.local/bin/claude --version)"; \
-    case "$installed" in \
-      *"$CLAUDE_VERSION"*) ;; \
-      *) echo "pinned CLAUDE_VERSION=$CLAUDE_VERSION but installed: $installed" >&2; exit 1 ;; \
-    esac
+    printf '%s' "$installed" | grep -qwF "$CLAUDE_VERSION" \
+      || { echo "pinned CLAUDE_VERSION=$CLAUDE_VERSION but installed: $installed" >&2; exit 1; }
 ENV PATH=/home/vscode/.local/bin:$PATH
 USER root
--- a/tests/devcontainer_prebuild.rs
+++ b/tests/devcontainer_prebuild.rs
@@ -301,9 +301,19 @@ fn only_the_publishing_job_can_write_a_package() {
 
 /// Every argument to `docker push` in the workflow, in order. Collected by
 /// what the command *does*, not by what the reference is expected to look
-/// like: a push target renamed to anything at all — another repository,
-/// another registry, a bare local tag — still lands here, where a filter on
-/// the expected name would let it vanish from the comparison entirely.
+/// like: rename the existing push's argument to anything at all — another
+/// repository, another registry, a bare local tag — and it still lands here,
+/// where a filter on the expected name would let it vanish from the comparison
+/// entirely.
+///
+/// That rename is the whole of the scope, and the wording used to claim more
+/// (#158). This matches three whitespace-separated tokens, so it sees a push
+/// only where `docker push <ref>` is written as three tokens: splitting the
+/// command across a line continuation (`docker \` then `push …`) hides it, as
+/// does publishing by some other mechanism entirely — a `build-push-action`
+/// step, `buildx --push`, a `curl` to the registry API. Those are additions
+/// rather than drift, and adding one takes push access to this repository,
+/// which is also enough to edit this file.
 fn pushed_references(workflow: &str) -> Vec<String> {
     workflow
         .split_whitespace()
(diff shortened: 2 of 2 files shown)


blooop (commented):

Two independent fresh-context review axes (Standards, Spec), run in parallel; findings are not merged or reranked across axes. Fixed point: merge-base `3a54273` with `main`; diff reviewed as `git diff 3a54273...d082b2a`. Spec = ticket #158 plus its grilling-resolution comment (5301886226).

## Standards

Ran the check chain at `3a54273...HEAD`: `cargo fmt --all --check` clean, `cargo clippy --all-targets --all-features --locked` (RUSTFLAGS=-Dwarnings) clean, `cargo test --locked --test devcontainer_prebuild` **7/7 green**.

**Blocking — the new assert message under-claims, in the direction the PR set out to fix.** `tests/devcontainer_prebuild.rs:383-386`: "A deliberately added publish to another registry is outside this tokenizer's sight". It is not. Mutation-checked — appending `run: docker push quay.io/blooop/wayfinder-devcontainer:latest` fails this very test:
```
left: ["ghcr.io/…:latest", "quay.io/…:latest"]
```
`pushed_references` is registry-agnostic by construction, and `-t`/`--tag` collection catches a non-ghcr build tag too. The accurate limit is the one `pushed_references`' own new paragraph states (`:311-316`): a publish through a *mechanism* it cannot see — line continuation, `build-push-action`, `buildx --push`, `curl`. Swapping "to another registry" for "by a mechanism this tokenizer does not see" makes the two texts agree and both true.
*[Resolved during review — orchestrator commit bae6d2f: the message now reads "guards drift of any publish the tokenizers can see — a renamed argument, a deleted step, a second tag, whatever the registry. A publish added by a mechanism they do not see (a line continuation, a build-push-action step) is outside it". This also corrects the grilling resolution's own imprecision — the blind spot was registry-framed there and is mechanism-shaped in fact; correction noted on the ticket.]*

**Non-blocking — limit prose placed inconsistently.** The other two limits went onto doc comments (`:308`, `:332`); this one went into the failure message, leaving the test's own doc comment (`:354-361`) unchanged. The message a red build prints is now mostly about what did *not* break. House style ("doc comments carry the design decisions", AGENTS.md) points at `:354`.
*[Resolved during review — orchestrator commit bae6d2f: the test's doc comment now carries the full limit paragraph; the message keeps a compact form.]*

**Non-blocking — `.devcontainer/Dockerfile:80` "refuses both" is right, "word boundary" reads stronger than it is.** Verified: `2.1.2234` no, `12.1.223` no, `2.1.223 (Claude Code)` yes. But `.` and `-` are non-word chars, so `3.2.1.223` and `2.1.223-beta` both match, and `v2.1.223` does *not* (fails loud — safe direction, but not the claimed shape-agnosticism).
*[Resolved during review — orchestrator commit bae6d2f: the comment now states the residual shapes explicitly ("not fully shape-proof… `2.1.223-beta` would still match, `v2.1.223` refused loudly, the safe direction").]*

**Shell: correct.** `printf '%s' | grep -qwF … || { …; exit 1; }` — the `||` left side is errexit-exempt as intended, quoting is right, stderr message preserved, `exit 1` fails the RUN (verified rc=1 under `set -eux`). Better than the `case` it replaces (no herestring needed in `sh`).

**Nit** — `Dockerfile:81-84` justifies not field-parsing by "the three tools pinned in this file each print a different one", but only `claude` goes through this comparison; `gh`/`zellij` are pinned via download URL. Argument holds by analogy, overstated as fact.
*[Resolved during review — orchestrator commit bae6d2f: rewritten to speak only of the one output actually compared.]*

Axis verdict: **Comment** (the blocking wording corrected in-review; nothing blocking remains)

## Spec

Compliant. The diff is one commit (`d082b2a`) touching exactly two files — `tests/devcontainer_prebuild.rs` (three doc/message texts) and `.devcontainer/Dockerfile` (comment + the one-line pin check). I read every hunk: neither `pushed_references` nor `published_references` has a single body line changed, so no collector behaviour moved, as the resolution required ("comment/message text only plus the one-line pin-check change").

Reproduced independently in a fresh clone at merge-base `3a54273`:

1. `cargo test --locked --test devcontainer_prebuild` → 7/7 green.
2. Renaming the workflow's real `docker push` argument (`.github/workflows/devcontainer.yml:88`) reddens `the_workflow_publishes_exactly_the_reference_the_default_config_boots_from`. The prose narrowing did not quietly narrow the guard. Appending a `docker/build-push-action` step pushing `docker.io/evil/...` leaves 7/7 green — which is now what the assert message and doc comment *say*, per "the two escaped mutants … are *not* expected to redden".
3. Shell check with pin `2.1.223`: new `grep -qwF` refuses `2.1.2234` and `12.1.223`, accepts `2.1.223 (Claude Code)`; the old `case` accepted all three. Exactly the form the rider named.
4. The corrected login claim is true against the workflow: line 71 is `docker login ghcr.io -u …` — bare host, no slash, never collected by the `ghcr.io/` prefix sweep; `the_push_authenticates_with_the_runs_own_token` does cover it. No stale copy of the old claim survives elsewhere in the repo.

Judgement calls, all sound:

- **nit** — the extended Dockerfile comment is seven lines, but acceptance allows comment text and it records the superstring reasoning where the next editor will read it.
- **nit** — the `v2.1.223` false-red is real (I confirmed `grep -qwF` refuses it), but it is inherent to the spec-named form, and `claude --version` prints `2.1.223 (Claude Code)` today.
- **non-blocking** — no fixture for the pin check is correct: it only runs during an image build, and acceptance capped the change at one line.

Axis verdict: **Approve**

## Verdict

**Approve.** Spec approves with the guard's non-vacuity re-mutated independently and every hunk read. Standards' one blocking finding — the new wording claimed a registry-shaped blind spot where the real one is mechanism-shaped — was corrected during review in orchestrator commit `bae6d2f` (7/7 prebuild tests re-run green, fmt clean), along with the placement, Dockerfile-comment precision, and analogy nits. That finding also corrects the grilling resolution's own framing, noted on the ticket. No blocking findings remain. GitHub may refuse the approve state on a same-account review; this written verdict is what the gate reads.

PR #167, opened 2026-08-15
The picker opens one way, and the hold resolves itself

### LINKED ISSUE #156: Make the cursor-write invariant structural, and let Pinned resolve itself
Non-blocking findings from PR #155's review (ticket #148), recorded so they do not decay in a PR comment.

1. **The cursor-write + `PickLaunch` block is triplicated** across the three picker-opening arms of `request_launch`. #148 made the rule explicit in a doc comment; an `open_picker` helper would make it **structural** — a fourth opening arm added later cannot then forget the write, which is the failure the ticket existed to prevent.
2. **`Pinned` is destructured back into `Some(&key)`/`None` at the `preserve_cursor` call site**, so the shape the enum removed reappears one line later. A `Pinned::resolve` method would finish what the sum type started.

Both are shape work with behavior unchanged; the existing suite pins the behavior.

blooop:
Closes #156

Two shape refactors recorded from PR #155's review. Behaviour is unchanged and the existing suite is what pins it: **478 passed / 32 ignored before, 478 passed / 32 ignored after, with no test edited**. That is the acceptance criterion for this ticket — a test that had to move would have meant behaviour moved with it.

## `open_picker`: the invariant is structural now

#148 established that staging chooses the stop it acts on — the picker cannot open without the cursor being written as `Cursor::Chosen`, or a refresh re-derives "the top" and carries the cursor off the very row whose launch is being picked. That rule was enforced by a doc comment on `request_launch` asking each of the three opening arms (map, project, ticket) to remember a five-line block. The block is one `open_picker` call now, so a fourth stop worth staging reaches the picker with its row already chosen, or does not reach the picker at all.

The doc comment moved with the code it describes: the invariant paragraph is on `open_picker`, and `request_launch` keeps a one-sentence pointer rather than restating it. The helper's prose is explicit that it is the *opening* site, as distinct from the launch-picker key handler, which re-seats an overlay already up and correctly does not write the cursor.

## `Pinned::resolve`

`Pinned` was built as a sum type in #155 to stop "pinned by identity" and "pinned by position" sharing one shape told apart by reading an inner `Option`. It was then matched straight back into `Some(&key)` / `None` at the `preserve_cursor` call site. `resolve` moves that match onto the type, so the two arms stay two arms down to the one call that needs them to be one, and the call site is a single expression.

## Judgement call: no new tests

The instinct is to add a test that the map and project arms also choose their row — `staging_a_launch_chooses_the_row_it_acts_on` only exercises the ticket arm. I did not, for two reasons. First, it would pin the same claim a third time without covering the failure mode this ticket exists to prevent: a test can only assert about arms that exist, and the arm that forgets the write is by definition the one nobody wrote a test for. The structure is the guard; a test is not. Second, it would be new behavioural coverage smuggled into a behaviour-preserving diff, which is exactly the confusion the "no test changed" criterion is there to make visible.

`Pinned::resolve` is likewise a pure move of an expression, already covered through `replace_clusters` by the empty-list and reorder tests, and directly by `refresh::preserve_cursor`'s own unit tests.

A reviewer who disagrees on either point is disagreeing about coverage the ticket did not ask for, not about this diff's correctness — worth saying out loud rather than leaving implicit.

## Checks

Full AGENTS.md chain with `RUSTFLAGS=-D warnings` and `RUSTDOCFLAGS=-D warnings`: `cargo fmt --all --check`, `cargo clippy --all-targets --all-features --locked`, the three test commands, `cargo doc --no-deps --all-features --locked`. All green.

## Summary by Sourcery

Refactor launch staging and cursor preservation logic without changing behaviour.

Enhancements:
- Introduce a Pinned::resolve helper to encapsulate cursor preservation based on pinned identity or position.
- Centralize launch picker initialization into an App::open_picker helper to ensure the cursor is marked as chosen whenever the picker opens.
- Update documentation comments to describe the staging invariant and the picker opening site in one place, keeping request_launch as a thin delegator.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -228,11 +228,8 @@ enum Pinned {
 
 impl Pinned {
     /// Where the hold lands once `new_order` is the order on screen.
-    ///
-    /// Each arm answers for itself rather than being taken apart into the
-    /// `Option` it was made to replace: [`crate::refresh::preserve_cursor`] is *told*
-    /// what is pinned here, so the two shapes stay two shapes all the way down
-    /// to the one call that needs them to be one.
+    /// Each arm answers for itself, so the two shapes stay two shapes all
+    /// the way down to the one call that needs them to be one.
     fn resolve(&self, new_order: &[StopKey]) -> usize {
         match self {
             Pinned::Identity(key, fallback) => {
@@ -868,7 +865,7 @@ impl App {
     /// has no blockers and no stage, and a finished one is not drawn.
     ///
     /// Staging **chooses the stop it acts on** (#148), and every arm that
-    /// opens the picker goes through [`App::open_picker`] to do it — the
+    /// opens the picker goes through [`App::open_launch_picker`] to do it — the
     /// refusals do not, because nothing was staged, so nothing was chosen.
     fn request_launch(&mut self) -> Outcome {
         match self.cursor_stop() {
@@ -888,7 +885,7 @@ impl App {
                 // resumable as a build one — and rather more worth resuming.
                 let staged = self.resumable(staged, id.number);
                 self.prewarm(&staged);
-                self.open_picker(staged);
+                self.open_launch_picker(staged);
                 Outcome::Continue
             }
             // One stop, two screens, and the screen decides — which is the
@@ -906,7 +903,7 @@ impl App {
                     self.enter(&repo);
                     return Outcome::Continue;
                 }
-                self.open_picker(Staged::project(&repo));
+                self.open_launch_picker(Staged::project(&repo));
                 Outcome::Continue
             }
             Some(Stop::Ticket(row)) => self.request_ticket_launch(&row),
@@ -942,7 +939,7 @@ impl App {
             Some(staged) => {
                 let staged = self.resumable(staged, ticket.number);
                 self.prewarm(&staged);
-                self.open_picker(staged);
+                self.open_launch_picker(staged);
                 Outcome::Continue
             }
         }
@@ -961,7 +958,7 @@ impl App {
     /// one statement pair in one function precisely so that they cannot come
     /// apart: a new stop worth staging reaches the picker through here, and
     /// arrives with its row already chosen.
-    fn open_picker(&mut self, staged: Staged) {
+    fn open_launch_picker(&mut self, staged: Staged) {
         self.cursor = Cursor::Chosen(self.cursor_pos());
         self.overlay = Overlay::PickLaunch {
             candidate: staged.default_candidate(),
--- a/src/app.rs
+++ b/src/app.rs
@@ -226,6 +226,25 @@ enum Pinned {
     Position(usize),
 }
 
+impl Pinned {
+    /// Where the hold lands once `new_order` is the order on screen.
+    ///
+    /// Each arm answers for itself rather than being taken apart into the
+    /// `Option` it was made to replace: [`crate::refresh::preserve_cursor`] is *told*
+    /// what is pinned here, so the two shapes stay two shapes all the way down
+    /// to the one call that needs them to be one.
+    fn resolve(&self, new_order: &[StopKey]) -> usize {
+        match self {
+            Pinned::Identity(key, fallback) => {
+                crate::refresh::preserve_cursor(Some(key), *fallback, new_order)
+            }
+            Pinned::Position(fallback) => {
+                crate::refresh::preserve_cursor(None, *fallback, new_order)
+            }
+        }
+    }
+}
+
 #[derive(Debug)]
 pub struct App {
     /// The clusters on screen: every open map that has arrived, keyed by id.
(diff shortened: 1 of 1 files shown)


blooop (commented):

Two independent fresh-context review axes (Standards, Spec), run in parallel; findings are not merged or reranked across axes. Fixed point: merge-base `09eaa8c` with `main`; diff reviewed as `git diff 09eaa8c...a871f13`. Spec = ticket #156.

## Standards

**Checks (all green).** `cargo fmt --all --check` clean. `RUSTFLAGS=-Dwarnings cargo clippy --all-targets --locked` clean. `cargo test --locked`: **478 passed / 32 ignored / 0 failed**, exactly as expected. Also ran `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --all-features --locked` (AGENTS.md flags this as the check that lies without the env var, and this diff adds a new intra-doc link) — clean. Diff is `src/app.rs` only, every hunk above line 1355 where `#[cfg(test)]` begins: **no test file or test module touched**, and none needed to be — coverage drives the picker through keys, not by constructing the overlay.

**"The only place it opens" — verified.** Repo-wide there are exactly two `Overlay::PickLaunch` assignments: `src/app.rs:966` (the helper) and `src/app.rs:1183` (the re-seat at the tail of `handle_pick_launch_key`). Nothing in `tests/` or `examples/` constructs one. The builder's reading holds, and it is not incidental: `handle_pick_launch_key`'s own pre-existing doc (`src/app.rs:1101-1105`) already declares "the invariant is that nothing here may touch them [query and cursor]". Two independent docs, one fact, no contradiction.

**Refactor lands.** `fn open_picker(&mut self, staged: Staged)` — minimal, no speculative parameters; `resumable`/`prewarm` correctly stay at the call sites, since the `Project` arm (`src/app.rs:906`) does neither. Returning `Outcome` would shave three lines but fuse a state mutation into key-handler control flow — the `()` boundary is right. `Pinned::resolve` is behavior-identical (`&key`→`Some(key)`, `fallback`→`*fallback`) and its doc states the postcondition honestly.

**non-blocking** — `src/app.rs:231-235`: `resolve`'s doc re-tells the rationale the enum doc already carries at `src/app.rs:212-221` ("an `Option` inside a tuple… told apart at the use site"). #148's invariant moved cleanly to one place; this one was duplicated rather than moved. Trimming to the first sentence would fix it.
*[Resolved during review — orchestrator commit 789d67b: trimmed to the postcondition plus one sentence; the rationale lives only on the enum.]*

**nit** — `src/app.rs:233` is 86 chars, the longest doc line in the file, which otherwise hand-wraps to ~79. `rustfmt.toml` calls hand-wrapping out explicitly.
*[Resolved during review — orchestrator commit 789d67b: the trim removed the line.]*

**nit** — `open_picker` names one of two pickers (`PickLaunch`/`PickCheckout`); `open_launch_picker` disambiguates.
*[Resolved during review — orchestrator commit 789d67b: renamed at all five sites; fmt/clippy/suite re-verified green (478/32).]*

Axis verdict: **Approve**

## Spec

**Compliant.** Both ticket points are delivered, and behaviour-unchanged is reproduced independently.

**Reproduced** (clone of `wayfinder/wayfinder-156`, merge-base `09eaa8c`):
- `cargo test --locked` at HEAD: **478 passed / 0 failed / 32 ignored**. Same command at the merge-base: **478 / 0 / 32**. Identical.
- Diff is `src/app.rs` only, +49/−37. All six hunks land between lines 228 and 974; `#[cfg(test)] mod tests` starts at 1355, and `#[test]` count is 69 on both sides. Zero test hunks — the ticket's "the existing suite pins the behavior" is honoured literally.

**Point 1** — "the cursor-write + `PickLaunch` block is triplicated across the three picker-opening arms": all three arms (`src/app.rs:891`, `:909`, `:945`) now call the helper, and the `Cursor::Chosen(self.cursor_pos())` write is the first statement inside it (`:964`). The invariant doc comment moved onto the helper; `request_launch` keeps a pointer. Satisfied.

**Point 2** — "`Pinned` is destructured back into `Some(&key)`/`None` at the `preserve_cursor` call site": the destructure is gone; the call site is `pinned.resolve(&new_order)` (`:855`), and the match lives in `Pinned::resolve` (`:236`). Grepped crate-wide: the only other `Pinned` site is the *construction* at `:843-844` (`Option` → `Pinned`), which is the sum type's intended entry boundary, not a re-widening. No other destructure exists. Satisfied.

**non-blocking — "structural" is overclaimed in prose.** `pub enum Overlay` (`:70`) has public variants, so a fourth arm can still write `self.overlay = Overlay::PickLaunch { … }` directly and skip the write. The nearest in-file template for doing so is the re-seat site at `:1183`, which correctly has *no* cursor write — the likeliest copy-paste source is the one that omits the thing. The helper strictly improves the default path and does what the ticket named ("an `open_picker` helper would make it structural"), so spec is met; but the guard is convention plus doc comment, not the type system. The PR body's "cannot forget" reads stronger than what the types enforce.
*[Noted for the record rather than changed: the ticket's own word was "structural" and the helper is what it asked for; making it type-enforced (private variant / constructor-only overlay) would be new shape work beyond this ticket's scope. This paragraph is the honest statement of the limit.]*

**Judgement call (no new tests): sound.** The ticket frames both items as "shape work with behavior unchanged; the existing suite pins the behavior," and a new per-arm assertion would be new coverage inside a diff whose acceptance criterion is that no test moved. The builder's reasoning — a test cannot cover an arm nobody wrote — is correct and correctly scoped.

Axis verdict: **Approve**

## Verdict

**Approve.** Both axes approve independently: Spec with behavior-unchanged reproduced on both sides of the merge-base and both shape claims verified crate-wide; Standards with the full check chain green and the refactor judged minimal. The three Standards residues were resolved during review in orchestrator commit `789d67b` (fmt/clippy/suite re-verified: 478 passed / 32 ignored); the "structural is convention, not types" limit is recorded above rather than papered over. GitHub may refuse the approve state on a same-account review; this written verdict is what the gate reads.

PR #168, opened 2026-08-15
Guard: every test in tests/live_*.rs carries #[ignore], structurally

### LINKED ISSUE #165: Guard: every test in tests/live_*.rs carries #[ignore], structurally
Non-blocking finding from PR #164's review (ticket #42), recorded so it does not decay in a PR comment.

Nothing in CI runs the live binaries unfiltered — ci.yml stops at `--lib --bins --examples` plus two named hermetic binaries and `-- common::`, and live.yml selects with `--ignored`. So a future live test added *without* the attribute breaks the offline-green promise (#42's whole point) with no check anywhere, and — missing the attribute — also never runs in live.yml: unobserved in both directions. AGENTS.md states the rule in prose only.

**Done when:** a config-as-behavior test (the `devcontainer_prebuild.rs` pattern; `live_devlaunch.rs:1583` already guards its own source the same way) walks `tests/live_*.rs` and asserts every `#[test]`/`#[tokio::test]` is paired with `#[ignore]`, and ci.yml runs it. One test, one workflow line.

> *This ticket was generated by AI during review follow-through.*

blooop:
Closes #165

The offline-green promise (#42) — a bare `cargo test` green in a fresh checkout with no network, no authenticated `gh`, no `devlaunch` — rests entirely on every test in `tests/live_*.rs` carrying `#[ignore]`. Nothing checked that. Unignored, a live test breaks the bare run on any machine lacking what it wants; and being unignored it is also never selected by `live.yml`, which runs the gated set with `--ignored`. Unobserved in both directions, with the rule stated in AGENTS.md prose and enforced nowhere.

`tests/offline_green.rs` now walks the live sources as text and asserts every `#[test]`/`#[tokio::test]` sits in an attribute group that also carries `#[ignore]` — either order, bare or with a reason string, doc comments and `#[cfg]`/`#[allow]` interleaved freely. `ci.yml` runs it beside the two hermetic binaries already named there.

## Red before green

The defect this guards is a missing attribute, so the red step is a mutation of the sources rather than of the code. In the working tree only, never committed:

1. Appended an ungated `#[tokio::test] async fn temporary_ungated_live_test_for_the_red_step()` to `tests/live_discovery.rs` → `every_test_in_a_live_source_is_gated_behind_ignore` **FAILED**, naming `live_discovery.rs:72 temporary_ungated_live_test_for_the_red_step`.
2. Reverted → **green**.
3. Appended a test gated the *other* way round — `#[ignore]` first, bare, with a doc comment, a blank line, `#[cfg(unix)]` and a `#[tokio::test(flavor = "current_thread")]` between it and the `fn` → **green**, so the guard reads the group and not a fixed layout.
4. Appended `#[tokio::test] async fn all_on_one_line() {}` — a shape the scanner cannot pair → the *second* test **FAILED** with "contains 2 test attributes but the scanner paired only 1". The unparseable shape is loud, not silent.
5. Reverted, re-ran the whole trace after the clippy-driven refactor of the scanner, red and green both reproduced.

## Judgement calls worth a reviewer's eye

- **Its own file, not joined to `devcontainer_prebuild.rs`.** This repo runs one binary per contract — `skill_docs.rs` is the skill snippets, `devcontainer_prebuild.rs` is the devcontainer/publish pair — and each is named for the thing it holds. Live-test gating is a third contract sharing nothing with the second but its hermeticity; folding it in would make that file's name a lie and its failure output ambiguous about which promise broke.
- **The name is deliberately not `live_*`.** Under that glob the walk would walk itself and demand its own gating, which would gate the guard out of the very run it exists to protect. `offline_green` names the promise (#42's phrasing, and the README's) rather than the mechanism; the file header says why the obvious name was refused.
- **Line-based scanning, no `syn`.** A proc-macro toolchain to read an attribute this repo writes one way in a rustfmt-normalised file is a dependency out of proportion to the contract. The cost is a scanner with limits, so the limits are named in the doc comment (an attribute split across lines; `#[test] fn x()` on one line; a macro-generated test) — and the first two are *checked*, by `every_test_attribute_in_a_live_source_is_one_this_guard_can_see`, which compares what was paired against the test attributes present. A shape it cannot read is a red run asking to be taught, not a quiet pass. The third is beyond any textual guard and this repo writes no such macro.
- **Two vacuity assertions, on purpose.** Empty source list, or zero tests found, both fail loudly — a guard that walks nothing passes everything, and the rename that would cause it is exactly the kind of change nobody would think to re-check this file for.
- **Declined: a test asserting `ci.yml` names this binary.** It is the same class of drift, and tempting given this repo's "an assertion no workflow runs is not an assertion" line — but `skill_docs` and `devcontainer_prebuild` are not guarded that way either, and the ticket asked for one test and one workflow line. Flagging rather than smuggling it in.
- **Two words changed in existing `ci.yml` comments.** "The other offline resident" and "The other config-as-behavior binary" become "An" and "Another": with a third such binary they were about to be false.
- **Docs touched only where they enumerate.** AGENTS.md's check chain and its "two binaries under `tests/` are exceptions" (now three), plus the "a new live test needs the attribute and a workflow line" bullet — the attribute half is now checked, the workflow half still is not, and it says so. README's CI command list and its "three things under `tests/` are not gated" (now four). The `25`/`32` counts elsewhere are unchanged and stay true: the new tests are hermetic, not ignored.

## Checks

Full chain per AGENTS.md with both `-D warnings` flags: `fmt --check`, `clippy --all-targets --all-features --locked`, `cargo test --locked`, `cargo doc`. Bare `cargo test --locked` is **480 passed, 0 failed, 32 ignored** — 478 before, plus these two, which run by default because they are hermetic.

## Summary by Sourcery

Add an offline test binary that enforces the structural gating of all live tests behind #[ignore] and integrate it into the standard test chain and CI.

New Features:
- Introduce tests/offline_green.rs as a hermetic binary that scans tests/live_*.rs and asserts every test function is gated with #[ignore].

CI:
- Wire the offline_green test binary into the CI workflow alongside existing hermetic binaries, clarifying comments to reflect multiple offline/config-as-behavior residents.

Documentation:
- Update AGENTS.md and README.md to describe the offline_green guard, its role in enforcing #[ignore] on live tests, and to adjust counts of ungated offline binaries under tests/.

Tests:
- Add two structural tests that verify all live test functions are ignored and that the line-based scanner sees every test attribute in live sources.
- Ensure missing or unreadable test shapes in live sources fail loudly, preventing silent drift from the offline-green contract.


blooop (commented):

Two independent fresh-context review axes (Standards, Spec), run in parallel; findings are not merged or reranked across axes. Fixed point: merge-base `2996766` with `main`; diff reviewed as `git diff 2996766...fbb2bd8`. Spec = ticket #165.

## Standards

**Gates.** `cargo fmt --check` clean; `cargo clippy --all-targets --all-features --locked -D warnings` clean; `cargo test --locked` = **480 passed / 32 ignored**, as expected. Mutation check passed: appending a bare `#[test] fn probe_plainly_ungated()` to `tests/live_discovery.rs` reddens the guard with `live_discovery.rs:72 probe_plainly_ungated`; reverted.

**Census accuracy — verified against reality.** 25 test attributes and 25 `#[ignore]`s across `tests/live_*.rs` (matches AGENTS.md:139 "All 25"); 32 ignored total; README:132-140's chain matches `ci.yml`'s step order exactly; `live.yml:92` really selects with `--ignored`, as the new comment claims. The two sibling uniqueness comments (`ci.yml:91`, `:98`) were correctly demoted from "the other" to "an/another". Doc culture is well met: the module doc carries the design decisions (why line-based not `syn`, why not named `live_*`) and states what is *not* asserted.

**Non-blocking — a silent hole exists.** The stated limits (`tests/offline_green.rs:107-117`) are not the real limits. Attribute-shaped text inside a multi-line string literal is scanned as an attribute. Verified empirically: a `const S: &str = "…\n#[ignore = \"live: network\"]";` sitting directly above a real `#[test] fn` makes the guard report that ungated test as gated, and the companion self-check stays green. Block comments are safe only because the closing `*/` line clears the run. No live file embeds such a snippet today, so plausibility is low — but the fourth limit deserves naming alongside the three.
*[Resolved during review — orchestrator commits ac741f2/b9a3023: the limit is now the fourth bullet, stated with its direction ("the count check below cannot catch that direction") and its mitigation (a live file quoting attribute text is a review problem before a scanner problem).]*

**Non-blocking.** The walk is top-level `tests/live_*.rs` only. Tests reached through `mod common;` (three deliberately unignored `#[test]`s compiled into every live binary) and a dir-style `tests/live_x/main.rs` target are outside it. Correct today, but the exclusion is load-bearing and undocumented.
*[Resolved during review — orchestrator commits ac741f2/b9a3023: `live_sources`' doc now names both exclusions and what adding a dir-style target would demand.]*

**Nit.** `#[cfg_attr(all(), ignore = "…")]` is reported ungated with advice to "Add `#[ignore = …]`" — loud, so safe, but misleading (verified).
*[Resolved during review — the limits list now carries the cfg_attr bullet: reported ungated on purpose, this repo gates unconditionally.]*

**Nit.** `tests_in(text)` recomputed purely to restate the value `assert_eq!` already prints; `tests.join(&name)` re-joined when `path` is in hand.
*[Resolved during review — bound once (`paired`), `read_to_string(&path)`.]*

**Nit.** `ci.yml:117` "The one part of `tests/` that needs neither network nor `gh`" is now false with three siblings above it; `ci.yml:91` comment left unrewrapped; AGENTS.md's chain omits the `--test live_fetch -- common::` line ci.yml carries (pre-existing).
*[Resolved during review — comment reworded ("offline like its three siblings above"), rewrapped, and the chain gained the `common::` line.]*

Axis verdict: **Approve**

## Spec

**Verdict: compliant.** Every done-when clause of #165 is satisfied literally, and I reproduced the builder's full trace independently rather than taking the PR body's word for it.

**Reproduced**
- `cargo test --locked` → **480 passed / 32 ignored** (aggregated across all binaries), matching the claim; the two new tests are hermetic, so the `25`/`32` census lines in `AGENTS.md:137,139` and `README.md:156` stay true (`grep -c "^#\[ignore" tests/live_*.rs` = 25).
- (a) Appended an ungated `#[tokio::test]` to `tests/live_fetch.rs` → `every_test_in_a_live_source_is_gated_behind_ignore` **FAILED**, naming `live_fetch.rs:97 an_ungated_live_test`. Reverted → green.
- (b) Adversarial gated shapes — `#[ignore]` before `#[tokio::test]`; doc comment, ordinary comment and `#[cfg(unix)]` interleaved between `#[tokio::test(flavor = "multi_thread")]` and `#[ignore]`; `pub fn` — all **green**, and the self-check paired all three. Reverted.
- (c) `#[tokio::test] async fn x() {}` on one line → the companion **FAILED** loudly ("contains 2 test attributes but the scanner paired only 1"). I also traced the split-attribute case by hand: it too lands in the loud path, not a silent pass. Tree left clean.
- ci.yml: the step runs the binary (`cargo test --locked --test offline_green`) and sits between "Devcontainer prebuild contract" and "The live fixture's diagnostic" — exactly among the sibling hermetic steps.
- Self-exclusion is **sound**: `live.yml`'s enrolment is per-file, so `offline_green` cannot be swept in; and not being `live_*` keeps `live_sources()` from demanding the guard gate itself out of the run it guards. The module header states this reasoning.

**Findings**
- *non-blocking — ruling on two tests vs. "One test, one workflow line."* Within scope. The line reads as one guard binary and one CI step, both literally met. The second `#[test]` is not a second contract: without it the primary assertion has a silent false-pass (an unparseable shape reports as gated), which is the same unobserved-omission failure #165 exists to close. It earns its place.
- *nit — declined ci.yml pinning.* Sound scope discipline. Neither `skill_docs.rs` nor `devcontainer_prebuild.rs` pins its own enrolment, so pinning here would be a one-off. The residual drift is repo-wide, not this ticket's; a separate ticket at most, and the PR flags it rather than smuggling it.
- *nit.* The walk is top-level `tests/live_*.rs` only — exactly what the ticket names.

Axis verdict: **Approve**

## Verdict

**Approve.** Both axes approve independently: Spec with the full red→green trace re-run (guard, adversarial shapes, self-check tripwire) and every census figure verified; Standards with the gates green and one real discovery — the string-literal blind spot — which is now named in the scanner's own limits list along with the walk-scope exclusions (orchestrator commits `ac741f2` + `b9a3023`, fmt/clippy/guard re-verified). No blocking findings remain. GitHub may refuse the approve state on a same-account review; this written verdict is what the gate reads.

Issue #169, opened 2026-08-18
An update ships a skill; the machine that updated should have it


blooop:
## Destination

A `pixi global update wf` on a machine that has already run `wf skills install`
leaves that machine with the **complete** set of skills this build ships — not
the set that existed the day install was last run. Done looks like: update wf,
launch a node, and a skill the release newly ships answers, in the container as
well as on the host.

## Notes

Single-ticket map: known work, no fanning.

Reported from a second machine: `pixi global update wf`, then an isolated launch
into a devcontainer, then `Unknown command: /wf-one`. The cause is not the
container mount — it is that `skills::refresh` (`src/skills.rs:572`), the
self-heal every launch runs, skips every skill whose link is not already
`Link::Current`. An update rewrites the bundle in the pixi prefix and touches
nothing under `~/.claude`, so a skill that is new in a release (`wf-one`) or was
renamed by one (`tdd` → `wf-tdd`, #104) has no link on that machine and never
gets one. The launch says nothing, and the failure surfaces seconds later inside
the agent as an unknown command.

Calls for `wf-tdd` then `wf-review`. The standing rules refresh must keep:
never touch a real directory it did not create, never touch a link it cannot
prove is its own, and stay a complete no-op on a machine that never opted in.

## Left outside this ticket

Two things the review of #170 turned up that its scope does not cover, recorded
here rather than grown into tickets on a single-ticket map:

- **A launch never calls `sweep`.** `refresh` iterates `BUNDLED`, so a link an
  older name left behind — #104's `tdd` — is invisible to it and stays dangling
  until someone types `wf skills install`. Healing the missing is not the same as
  clearing the orphaned.
- **`is_ours` proves link *shape*, not authorship.** It matches a hand-made link
  straight into the recorded `WF_SKILLS_DIR` checkout, and any foreign path
  ending `share/wf/skills`. Harmless where it was written (`sweep`, after a typed
  command) and worth a second look now that the launch path repoints links with
  nobody having typed anything.

blooop:
Ticket #170 is merged as #171 (`9ffba5a`) and the map closes with it: a launch now heals a missing or stale skill link, so `pixi global update wf` alone leaves a machine with the skills the build ships. The two findings left outside the ticket are recorded in the body above. Shipping in 0.18.0 (#172).

PR #171, opened 2026-08-18
A launch heals a missing or stale skill link, not just a stale copy

### LINKED ISSUE #170: A launch heals a missing or stale skill link, not just a stale copy
## Work

`skills::refresh` (`src/skills.rs:572`) is the self-heal `wf` runs on the way
into every agent (`src/main.rs:417`). It keeps the *contents* of the installed
copies in step with the bundle they were installed from, and it refuses to
consider a skill at all unless its link is already `Link::Current`:

```rust
if !source.is_dir() || link_state(name, &target.links.join(name)) != Link::Current {
    continue;
}
```

So the set of skills a machine has is frozen at whatever `wf skills install`
last saw. `pixi global update wf` rewrites the bundle in the pixi prefix and
writes nothing under `~/.claude`; a skill the new build ships but this machine
has no link for stays unlinked through every launch that follows, silently. The
report that opened this: update on a second machine, launch a node into its
devcontainer, `Unknown command: /wf-one`. The same hole swallowed the #104
rename — after it, `tdd`'s link is stale and `wf-tdd` has none.

Make a launch heal the link, not just the copy:

- A **missing** link for a bundled skill gets its copy written and its relative
  link created, exactly as `install` does.
- A **stale** link is repointed only when `is_ours` proves `wf` wrote it — which
  covers the pre-#110 absolute link into the pixi prefix, the shape that
  resolves on the host and dangles in the container.
- Everything else is left exactly as it is. `Link::Unmanaged` — a real
  directory, chezmoi's or a hand-edit's — is never deleted, and a link `wf`
  cannot prove is its own is never rewritten.
- The prompts still come from `installed_from`, the recorded source, never from
  this build's bundle: an ordinary launch has no standing to undo a
  `WF_SKILLS_DIR` install, and healing a link must not become the loophole that
  does.
- A machine with no `.source` record — one that never ran `wf skills install` —
  stays a complete no-op.
- Creating a link is worth one line on stderr; silence is what made this
  invisible. Re-copying contents stays quiet, as now.

Done when:

- A test walks the reported sequence — install from a bundle without `wf-one`,
  swap in a bundle with it, `refresh` — and asserts the link now exists, is the
  relative `../wf-skills/wf-one`, and still resolves through the `remount`
  helper that stands in for the container's home.
- A test covers the renamed-skill shape: a stale link `wf` wrote is repointed,
  and a stale link it did not write is left alone.
- `refresh_touches_nothing_it_was_not_asked_to_own`,
  `a_launch_refreshes_the_prompts_that_were_installed_not_its_own` and
  `a_real_directory_blocks_rather_than_being_deleted` still hold — the last one
  now for `refresh` as well as `install`.
- The README's skills section says a launch keeps the *set* in step, not only
  the contents, so the documented promise and the code agree.

blooop:
Closes #170. Map: #169.

`skills::refresh` — what `wf` runs on the way into every agent — kept the
*contents* of the installed copies in step with the bundle they came from and
refused to look at a skill whose link was not already the one it writes. So the
set of skills a machine had was frozen at whatever `wf skills install` last saw:
`pixi global update wf` rewrites the bundle in the pixi prefix and writes nothing
under `~/.claude`, and a skill the new build ships stayed unlinked through every
launch that followed. The report that opened the ticket was `Unknown command:
/wf-one` inside a devcontainer, seconds after a launch, with nothing pointing
back at the update — and the same hole swallowed #104's rename.

A launch now heals the link as well as the copy:

- A **missing** link for a skill the recorded source ships gets its copy written
  and its relative `../wf-skills/<name>` link created, exactly as `install` does
  — so it still resolves inside the container, which is the only reason the
  indirection exists.
- A **stale** link is repointed only where `is_ours` proves `wf` wrote it, the
  same test `sweep` removes links by. That covers the pre-#110 absolute link into
  the prefix, the shape that resolves on the host and dangles in the container.
- Everything else is left as it is. A real directory another tool owns, a link
  `wf` cannot claim, and a machine with no `.source` record are all untouched.
- The prompts still come from `installed_from`, never from the launching build's
  bundle: healing a link must not become the loophole that undoes a
  `WF_SKILLS_DIR` install.
- Creating or repointing a link is one line on stderr; re-copying contents stays
  quiet. `refresh` reports which of the two it did (a new `Healed`) and the
  launch path owns the wording, the way `install`'s `Outcome` already works,
  and names the agent whose directory was healed — which is why
  the two existing `refresh` assertions read `Healed::Copied` where they read a
  bare name before. Their behaviour is unchanged.

Tests (red before green, one slice each): a launch links a skill the update newly
ships and it survives the `remount` that stands in for the container's home; a
launch tells a link it created apart from a copy it rewrote; a stale link `wf`
wrote is repointed and one it did not is left where it points.
`a_real_directory_blocks_rather_than_being_deleted` now pins that restraint on
the `refresh` path too — the path nobody typed anything to start.

README's skills section now says a launch keeps the *set* in step, not only the
contents, so the documented promise and the code agree.

`origin/main` was merged in after the first push, having gained a second agent
(Codex) in the meantime. The three conflicting files were resolved on their
merits: `src/main.rs` taken from main whole, because the `refresh_skills` this
branch had edited there now lives in `src/picker.rs` and takes an `Agent`; the
stderr line re-applied at that new home; and the prose in `src/skills.rs` and
the README brought into main's agent-neutral voice — a launch refreshes one
agent's skills directory, and the symptom is an unknown-skill prompt rather
than a Claude slash command.

## Summary by Sourcery

Keep each agent's installed skill set synchronized at launch by healing missing and wf-owned stale links without overwriting externally managed resources.

New Features:
- Make launches create missing skill links for skills shipped by the recorded source and repoint stale links previously created by wf.
- Report newly created or repointed skill links on stderr while keeping routine copy refreshes quiet.

Bug Fixes:
- Prevent updated builds from leaving newly shipped skills unavailable until a manual install, including links that resolve on the host but fail inside containers.

Enhancements:
- Preserve user- or tool-owned directories and links, and continue refreshing from the recorded installation source rather than the launching build's bundle.

Documentation:
- Document that launches synchronize both the installed skill set and skill contents, including automatic link healing.

Tests:
- Add coverage for newly shipped skills, link reporting, stale wf-owned links, and protection of externally owned links and directories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,18 @@ copy    /home/you/.claude/wf-skills (what the links point at, so a devcontainer 
   wf-review       not a link — another tool owns this one
 ```
 
+What a launch brings back in step is the **set** of skills as well as their
+contents. A skill this build ships that your `~/.claude/skills` has no link for
+gets one, and a link `wf` wrote that still points into a package prefix — the
+shape that resolves on your host and dangles in the container — is repointed at
+the copy. So `pixi global update wf` is the whole update: there is no
+`wf skills install` to remember, and no release where a skill exists everywhere
+except on the machine you are launching from. Creating a link is worth a line on
+stderr, because a launch that silently did not was how that lasted a release.
+
 `wf` never deletes a real directory it did not create — if chezmoi or a
 hand-edit owns one, it says so and leaves it, and the other four still install.
+A link it cannot prove it wrote is left alone the same way, healed by nothing.
 Set `WF_SKILLS_DIR` to install a checkout's prompts instead of the package's,
 which is what you want while you are editing them: the copy **remembers where it
 was made from**, so every launch re-copies from your working tree and an edit is
--- a/README.md
+++ b/README.md
@@ -109,13 +109,14 @@ copy    /home/you/.codex/wf-skills (what the links point at)
 ```
 
 What a launch brings back in step is the **set** of skills as well as their
-contents. A skill this build ships that the launched agent's skills directory
-has no link for gets one, and a link `wf` wrote that still points into a package
-prefix — the
+contents. A skill gets a link when this build ships it *and* the source the copy
+was made from carries it — both, because the link has to name something and the
+prompt behind it comes from that source, not from the build doing the launching.
+For a package install those are the same thing, so a release that adds a skill
+adds it here. A link `wf` wrote that still points into a package prefix — the
 shape that resolves on your host and dangles in the container — is repointed at
-the copy. So `pixi global update wf` is the whole update: there is no
-`wf skills install` to remember, and no release where a skill exists everywhere
-except on the machine you are launching from. Creating a link is worth a line on
+the copy. So `pixi global update wf` is the whole update, with no
+`wf skills install` to remember afterwards. Creating a link is worth a line on
 stderr, because a launch that silently did not was how that lasted a release.
 
 `wf` never deletes a real directory it did not create — if chezmoi or a
--- a/src/main.rs
+++ b/src/main.rs
@@ -421,21 +421,40 @@ async fn main() -> Result<()> {
     }
 }
 
-/// Bring the installed skill copies back in step with the bundle they were
-/// installed from.
+/// Bring the installed skills back in step with the bundle they were installed
+/// from — their contents, and which of them this machine has links for at all.
 ///
 /// Best-effort, and deliberately silent when there is nothing to do: a machine
 /// with no home directory to resolve, and one that never ran
-/// `wf skills install`, are not worth a word on the way into an agent. A copy
-/// that could not be *written* is different — the agent is about to run a
-/// prompt that is not the one that was installed — so that one is said out
-/// loud.
+/// `wf skills install`, are not worth a word on the way into an agent. Two
+/// things are worth one. A copy that could not be *written*, because the agent
+/// is then about to run a prompt that is not the one that was installed. And a
+/// link that had to be *created*, because that means this launch changed which
+/// prompts the machine has — the thing that used to happen only at an install
+/// you typed, and that a release's worth of launches did silently not do
+/// (#170).
 fn refresh_skills() {
     let Ok(target) = skills::Target::resolve() else {
         return;
     };
-    if let Err(err) = skills::refresh(&target) {
-        eprintln!("wf: could not refresh the installed skills: {err:#}");
+    match skills::refresh(&target) {
+        Err(err) => eprintln!("wf: could not refresh the installed skills: {err:#}"),
+        Ok(healed) => {
+            for (name, what) in healed {
+                match what {
+                    skills::Healed::Copied => {}
+                    skills::Healed::Linked { was: None } => {
+                        eprintln!("wf: linked the {name} skill, which this build ships and this machine had not");
+                    }
+                    skills::Healed::Linked { was: Some(old) } => {
+                        eprintln!(
+                            "wf: repointed the {name} skill, which still linked to {}",
+                            old.display()
+                        );
+                    }
+                }
+            }
+        }
     }
 }
 
--- a/src/skills.rs
+++ b/src/skills.rs
@@ -33,7 +33,8 @@
 //! resolves under both home directories. The copy is then the thing this module
 //! has to keep honest, and it does that in three places rather than trusting it:
 //! [`install`] rewrites it and records where it came from, [`refresh`] brings it
-//! back in step with *that* source at every launch, and [`status`] reports it as
+//! back in step with *that* source at every launch — links and all, so a build
+//! that ships a new skill needs no second command — and [`status`] reports it as
 //! [`State::Outdated`] when it has drifted — so a stale copy is something that
 //! gets *seen*, rather than a prompt that quietly runs a release behind.
 
@@ -546,8 +547,26 @@ pub fn install(bundle: &Bundle, target: &Target) -> Result<Vec<(String, Outcome)
     Ok(done)
 }
 
-/// Bring the copies back in step with the bundle they were installed from, and
-/// name the ones that moved.
+/// What a launch put right on one skill's behalf — reported rather than
+/// printed, for the same reason [`Outcome`] is: the caller owns the wording.
+///
+/// The distinction is which of the two is worth interrupting a launch to say.
+/// Rewriting a copy is the errand [`refresh`] runs every time and would be
+/// noise; creating a link changes *which prompts this machine has*, and a
+/// release went by with that happening never rather than silently, which is
+/// the same kind of invisible (#170).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Healed {
+    /// The copy behind an existing link was rewritten from the recorded source.
+    Copied,
+    /// A link was created, or repointed from `was` — a link `wf` could prove it
+    /// wrote itself.
+    Linked { was: Option<PathBuf> },
+}
+
+/// Bring the installed skills back in step with the bundle they were installed
+/// from — their contents *and* which of them this machine has links for at all
+/// — and name what moved.
 ///
 /// This is what keeps "a copy" from meaning "a copy that goes stale". `wf` is
 /// the thing that launches the agent, so the copy is rewritten by the very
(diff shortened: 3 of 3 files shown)


blooop (commented):

`wf-review` on head `01f97ca6bc60b7692fd41cc5fef8ae46c493b245` (branch `wayfinder/wayfinder-170`), fixed point `git diff b7718f96...01f97ca6` — README.md, src/picker.rs, src/skills.rs. Spec is #170, under map #169. Preflight: all four checks green (release job skipped as expected), PR `MERGEABLE`. The two axes were run as independent subagents in fresh context and are reported unmerged and unranked against each other.

## Standards

Overall this is careful, in-voice work: the doc comments carry the *why* (#170's failure mode, why `installed_from` and not this build's bundle, why the copy is written before the link), the restraint on links `wf` cannot prove it wrote is tested from both sides, and `refresh` reuses `is_ours` rather than inventing a second permission rule. Clippy (`-D warnings`, all targets) and `cargo test --lib` are green at head. Nothing here is **blocking**.

**1. Vacuous assertion — `src/skills.rs:1058-1061`** (highest-value fix). `a_launch_links_a_skill_the_update_newly_ships` deletes the bundle at :1053, so `refresh` early-returns at :622 (`installed_from(target).filter(is_dir)` → `None`) and the final `is_empty()` can never fail. Verified empirically: inserting `remove_dir_all(target.links())` immediately before that assert leaves the test passing with *every* link gone. The comment claims "the launch after that has nothing left to heal", which the code does not check. Fix is one move: assert idempotence before `remove_dir_all`, or assert the early-return honestly.

**2. Duplicated link-write mechanism — `src/skills.rs:658-666` vs `install` at `:544-551`.** Eight near-identical lines, down to both `with_context` strings (`"cannot replace the link {}"`, `"cannot link {} → {}"`). The policy around them differs legitimately, but the mechanism does not; this module already factors mechanism out (`recopy`, `copy_tree`, `clear`, `prune_mirror`), so a `write_link(target, name, replacing: bool)` fits house style and removes a two-site edit if the link shape ever changes again (which #110 already did once).

**3. `Option<Option<PathBuf>>` as a control flag — `src/skills.rs:633-645`.** Three meanings in a nested `Option`, decoded by a prose comment, plus a `continue` escaping a value-producing `match`. In a module that mints a named sum type for every such distinction (`Link` vs `State` at :266-288, `Outcome`, `Healed`) this reads as the odd one out, and `clippy::option_option` — pedantic, enabled at `Cargo.toml:38` — would reject the same type in a signature. A local `enum Relink { Leave, Create, Repoint(PathBuf) }` would delete the comment. The `copy_moved: bool` at :648 is *not* a violation: it states an independent fact no sum type here already carries.

**4. `Healed` vs `Outcome` is the right call**, not speculative generality: reusing `Outcome` would give `refresh` a return type with three variants it can never produce, which is exactly the illegal state the module already avoids by keeping `Link` and `State` apart.

**5. Prose nits.** The picker string (`src/picker.rs:593-596`) and README (:112) are accurate — the linked *set* does come from `BUNDLED`, i.e. this build — though the prompt's contents come from the recorded source; the README's "no release where a skill exists everywhere except on the machine you are launching from" (:117) overclaims for a `WF_SKILLS_DIR` install whose checkout lacks the new skill, and "the same test `sweep` removes links by" (`skills.rs:602`) is the same predicate with a different bundle argument. `README.md:113-115` wraps raggedly ("prefix — the" on its own line).

## Spec

Verdict: the ticket is met in behaviour. Two findings, **neither blocking**; one test assertion is dead.

**1-6 — met.** A `Link::Missing` bundled skill gets `Some(None)` → `recopy` runs *before* the symlink, and the link written is literally `Target::link_target(name)` = `../wf-skills/<name>` (`src/skills.rs:604-668`), pinned by the new test's `read_link` assert. `Link::Unmanaged` and un-provable `Link::Stale` `continue` *before* `copy_moved`/`recopy`, so "never deleted, never written over" holds on the refresh path too, and `a_real_directory_blocks_rather_than_being_deleted` now pins it with three asserts including `!target.mirror().join("wf-tdd").exists()`. `installed_from(target).filter(|s| s.is_dir())` early-returns, so "A machine with no `.source` record … stays a complete no-op" holds. The healed link points at a copy written from the *recorded* source, never this build's bundle. `src/picker.rs:583-604`: `Healed::Copied => {}` (quiet), one `eprintln!` per link. All 18 `skills` tests green.

**7 — `is_ours` is a shape heuristic, not proof.** Spec: *"a link `wf` cannot prove is its own is never rewritten."* Two probes at head: a hand-made link into the recorded `WF_SKILLS_DIR` checkout (`parent == bundle`) and a foreign link under `othertool/share/wf/skills` (`ends_with`) are **both** silently repointed and reported as `repointed …`. The first is realistic — the README itself describes linking straight at a checkout. Non-blocking: the shapes are genuinely indistinguishable from pre-#110 `wf` links, the outcome is a copy of that same source, and `sweep` already used the predicate. Safe cases confirmed: `Link::Stale(PathBuf::new())` → `parent()` is `None` → left alone; bare-filename targets → `Some("")` → left alone.

**8 — the last assert in `a_launch_links_a_skill_the_update_newly_ships` is vacuous.** `std::fs::remove_dir_all(&bundle.path)` precedes it, so `installed_from(…).filter(is_dir)` is `None` and `refresh` early-returns `Ok(Vec::new())` regardless. Proved by inserting `assert!(installed_from(&target).filter(|s| s.is_dir()).is_none())` at that point — it passes. Move the idempotence assert above the `remove_dir_all`. The rest of the test (relative link shape + resolution through `remount`) genuinely walks the reported sequence.

**9 — README.** It does say the **set** (spec: *"says a launch keeps the set in step, not only the contents"*). But "A skill **this build ships** that the launched agent's skills directory has no link for gets one" reads as inaccurate against *"The prompts still come from `installed_from`, the recorded source, never from this build's bundle"* — a link is created only where `BUNDLED` **and** the recorded source both carry the skill. Minor wording fix.

**Merge.** `src/main.rs` is byte-identical to `origin/main` (the branch's only change there had relocated); both stderr lines survive at their new home in `picker.rs`, now naming `agent.label()`; the README paragraph survives, agent-neutralised; every branch test and doc change is present; every Codex/agent-neutral change from main is intact. Nothing dropped either way.

**Multi-agent.** `install` covers `Agent::all()`; a launch heals `launch.agent()` only. That is the right reading of *"the self-heal `wf` runs on the way into every agent"* — healing the directory you are entering fixes the reported symptom, and the other agent heals on its own next launch. Not a spec gap.

## Reviewer's notes on the two axes

Kept out of both axis reports so neither is reranked by the other:

- **The axes disagree about the README's "this build ships"** (Standards §5 calls it accurate, Spec §9 calls it inaccurate). Both are looking at a real seam: the *set* healed is `BUNDLED ∩ installed_from`, so "this build ships" is necessary but not sufficient. Standards' sharper version of the same point — that "no release where a skill exists everywhere except on the machine you are launching from" overclaims under a `WF_SKILLS_DIR` install — is the wording worth fixing. Non-blocking either way.
- **A launch still does not `sweep`.** `skills::sweep` is called only from `run_skills(install: true)` (`src/main.rs:276`). #170 frames #104's rename as the same hole — and this PR closes the half it names (`wf-tdd` had no link → now created), but the residue half (a link named `tdd`, no longer in `BUNDLED`, dangling) is invisible to `refresh` because it iterates `BUNDLED`, and still needs a typed `wf skills install`. The ticket's "Done when" list does not ask for sweep-on-launch, so this is scope worth recording rather than a miss — possibly a follow-up on #169.

## Verdict

**Comment** — no blocking findings on either axis. Standards passes; Spec passes. `--comment` rather than `--approve` because GitHub refuses a self-authored approval, which is the normal case here; the written verdict is the gate.

Both axes independently landed on the same single concrete defect, and it is a test defect rather than a behaviour one: the final assertion of `a_launch_links_a_skill_the_update_newly_ships` (`src/skills.rs:1058-1061`) is dead, because the bundle is deleted first and `refresh` early-returns. The idempotence it claims to pin is in fact pinned by `a_package_update_is_picked_up_by_install_and_by_refresh` for the `Copied` path, but nothing pins it for the new `Linked` path. Worth one line moved before merge; not worth blocking a green, correct fix.

Everything the ticket named as an invariant holds, and holds on the launch path specifically: `Link::Unmanaged` `continue`s before anything is written, the no-`.source` machine is a total no-op, and the healed link points at a copy of the recorded source. The one honest caveat is that `is_ours` proves *shape*, not authorship — reusing `sweep`'s predicate is the right call and its risk is pre-existing, but the launch path is now the place nobody typed anything to start, so the two matching shapes Spec §7 lists are worth knowing about.

blooop:
Both non-blocking items from the review are closed on `87b879f`; the two recorded as follow-ups under #169 (sweep-on-launch, and `is_ours` proving shape rather than authorship) are untouched, as are the Standards suggestions about factoring the link write and the nested `Option` — those are refactors this stage deliberately leaves to a later one.

**The dead assertion.** Correct, and worse than dormant: it read as pinning the new `Linked` path's idempotence while pinning nothing at all. It has moved above the `remove_dir_all`, where the recorded source is still on disk and the healed link is the only reason the second launch is a no-op. Checked against the reviewer's own probe — deleting the link again at that point now fails the assert, where before deleting *every* link left it green.

**The README.** Taken as the Standards axis framed it: the healed set is what this build ships *and* what the recorded source carries, so the paragraph says both and no longer claims anything about a skill a `WF_SKILLS_DIR` checkout does not have. The overclaiming clause about "no release where a skill exists everywhere except on the machine you are launching from" is gone rather than reworded, and the ragged wrap with it.

Full chain green under `-D warnings` (427 lib, plus `skill_docs`, `devcontainer_prebuild`, `offline_green`, `live_fetch -- common::`), and all three PR checks pass on the new head. Not merged.

Issue #73, opened 2026-08-06
Credential posture and config-injection mechanics


blooop:
## Question

Settle wf's credential posture and the config-injection mechanism, following the findings of [Injecting the agent and credentials into arbitrary devcontainers](https://github.com/blooop/wayfinder/issues/69):

1. **Trust tiers**: is the default env-token injection (`claude setup-token`, `gh auth token` via `--secrets-file`) with a per-container `~/.claude` volume — and is a live host-mount tier offered at all for trusted repos? Who marks a repo trusted, and where does that flag live?
2. **Injection mechanics**: `devcontainer up --mount` cannot express `readonly` or `containerEnv`; getting those needs `--override-config`, which *replaces* the repo's `devcontainer.json` — so wf would need to parse JSONC and merge. Does wf take on a JSONC parse+merge capability, or accept the simpler rw-mount + `--additional-features` route?

The two halves are one decision: the mechanics bound which postures are expressible. The chosen mechanism is what the prototype must exercise.

blooop:
**breadcrumb:** Trust tiers settled: **one posture only**. Every repo gets revocable env tokens (`claude setup-token`, `gh auth token`) via `--secrets-file` plus a per-container named volume at `~/.claude` with `CLAUDE_CONFIG_DIR`. No live host-credential mount tier at any trust level, so no trust flag, no repo-trust registry, and wf needs no config file (it has none today — only the derived cache at `~/.cache/wf/projects.json`).

blooop:
**breadcrumb:** Binary route: **`--additional-features`** (`ghcr.io/anthropics/devcontainer-features/claude-code:1.0`) — no host mount, so no host-poisoning vector, no `readonly` to express, and therefore **no JSONC parse+merge in wf**; also fixes the musl/Alpine case the bind-mount route fails. **Fallback still live:** rw bind-mount of the host `versions/` tree via `--mount`, to be reconsidered if the prototype finds feature-install build time unacceptable. Prototype ([#70](https://github.com/blooop/wayfinder/issues/70)) must measure cold-build and warm-`up` cost of the feature route.

blooop:
## Resolution

**One posture, live read-write mount. wf takes on no JSONC merge.**

### 1. Trust tiers — there are none

There is exactly **one** credential posture for every repo, so there is no trust flag, no
repo-trust registry, and no wf config file (wf has none today — only the derived cache at
`~/.cache/wf/projects.json`, `src/projects.rs:163`).

[...]

Issue #182, opened 2026-08-22
Fix the 0.19.2 review findings


blooop:
## Destination

Every medium-severity finding from the 0.19.2 code review is fixed and merged — the reap PR-cap deletion hazard, silent GraphQL truncation, non-atomic projects.json writes, the reap running-guard TOCTOU, the skills-mirror adoption pruning, and the duplicate-row count/cursor bugs — plus the low-severity sweep and the two test/CI gaps that let defects hide (test-binary enrolment, red-run alerting).

## Notes

- **Execution is in scope.** Every ticket on this map is a build slice: drive it through `wf-tdd` and gate it through `wf-review`, per the lifecycle. There are no open design questions — the review settled the approach each ticket states.
- The findings come from the 2026-08-22 review of 0.19.2 (four independent reviewers over all of `src/`, `tests/`, and the workflows; each finding re-verified against the code). Full report: https://claude.ai/code/artifact/730921b5-f8ad-4959-b0d8-16dcdca443aa
- House rules apply and will bite: `main.rs` and `reap.rs` carry source-text guard tests that fire on edits near the reap path, and `reap.rs` byte-compares its PR query block against `fetch.rs`'s — a change to one query text moves the other. Read AGENTS.md before the first slice.
- Each ticket cites its evidence as `file:line` against the 0.19.2 tree (commit 1716e6f); re-locate before editing if the tree has moved.

## Decisions so far

All nine shipped in **v0.19.4** — built test-first, reviewed in fresh context, merged 2026-08-22.

- [Reap survives a truncated PR rollup](https://github.com/blooop/wayfinder/issues/183) — a linked-PR page the tracker says is incomplete is unreadable evidence, never a reap *(agent)*
- [The screen says when a map fetch was truncated](https://github.com/blooop/wayfinder/issues/184) — label count can no longer un-map a map; truncation reaches the count line *(agent)*
- [projects.json saves are atomic and single-seamed](https://github.com/blooop/wayfinder/issues/185) — temp-plus-rename behind one read-modify-write seam; a failed startup save reports instead of aborting *(agent)*
- [Reap re-checks running right before it removes](https://github.com/blooop/wayfinder/issues/186) — both deleting paths re-take the listing after the prompt answers *(agent)*
- [The skills mirror is pruned only when wf made it](https://github.com/blooop/wayfinder/issues/187) — pruning requires the `.source` record wf wrote; an adopted directory is refused, not swept *(agent)*
[...]

PR #195, opened 2026-08-22
The README states the tracker-content injection posture

### LINKED ISSUE #193: The README states the tracker-content injection posture
## Question

State the tracker-content injection posture in the README, beside the #73 trade.

Issue titles — attacker-writable on any repo that accepts issues — are embedded verbatim in the ctx JSON handed to agents run with `--dangerously-skip-permissions` (`launch.rs:117–122`, `1820–1827`), and `wf-auto` means that text can steer an unattended run. The mechanical escaping is airtight (single-argv throughout, proven by the real-`sh` canary test at `launch.rs:4440–4533`) and the README is already frank that the container is not a security boundary — but that section covers hostile *code* in a repo you launch into, not hostile *tracker content* steering the agent. The threat model deserves a written stance even if the answer is "same posture as #73: don't point wf at a tracker you have not read."

Deliverable: a README paragraph in the isolation/security section naming the surface (issue titles and bodies reach the agent's prompt), what is and is not defended (argv/JSON escaping yes, semantic injection no), and the operator guidance — plus the `skill_docs.rs`-style binding if one fits, so the paragraph cannot silently drift from what the launch actually embeds.

blooop:
Closes #193

## What the README was missing

The isolation section was already frank that the container is not a security
boundary — but that paragraph is about hostile **code** in a repo you launch
into. Nothing anywhere said what happens with hostile **prose** on the tracker
you launch *from*, and a launch embeds tracker-authored titles into the prompt
of an agent started with permissions bypassed. Under `wf-auto` that text can
steer a run nobody is reading along with. The mechanical escaping is airtight
and already proven end to end; what was undocumented was that airtight escaping
says nothing about the meaning.

Three paragraphs now say it, in the isolation section directly after the #73
trade they extend:

- **what reaches the agent** — the map's title and the ticket's title verbatim
  in the `ctx:` block, and the wider surface beyond it, since the skill's own
  first move is to read the ticket body and its comment trail;
- **what is and is not defended** — one argv entry, one JSON string, a real
  `sh` rebuilding a launch's argv byte for byte; and nothing at all about what
  the text says;
- **the operator guidance** — #73's answer one layer out, plus the observation
  that the picker is where the check is cheapest, because every title that will
  reach the prompt is on the screen you pick from.

## The binding

Documentation-shaped ticket, so the test-first angle is the one the ticket
suggested: `tests/skill_docs.rs`, where every other README↔binary binding
already lives and which CI already runs. Three guards, all red first:

1. **Paragraph identity**, pinned by equality through the existing
   `readme_paragraph` helper — the same treatment the exec seam's claim-free
   rule gets. A security stance sampled by phrase survives a sentence that
   reverses it, and these three claims only mean anything together.
2. **The central factual claim, read off the binary.** A launch is built
   through the production path with injection-shaped prose planted in both
   tracker titles, and the block is asserted to carry it back **unaltered**.
   Worth recording what the first draft got wrong: it asserted the prose
   survived into the prompt's *bytes*, and a title's double quote rides as
   escaped JSON — so it was measuring the serializer rather than the claim. The
   claim is about what the reader parses back out, which is exactly where
   attacker text and agent instruction stop being distinguishable.
3. **A closed world over the block's strings**, which is what makes this a
   drift guard rather than a spot check. Every string a launch hands over is
   either constrained by shape (a repo slug, a closed-vocabulary word — nobody
   writes prose into those) or one of the free-text fields the paragraph names.
   A future field carrying an issue body, a comment or a label into the prompt
   satisfies neither arm and fails naming itself, until the posture accounts
   for it.

Both arms were checked to bite rather than assumed to: planting prose in an
unnamed field fails with `` `repo` carries "owner/name with prose", which is
neither a constrained identifier nor a field the injection posture paragraph
names``, and narrowing the README's promise to drop the map title turns both
the identity test and the closed-world test red.

## Not in scope

No production behavior changes. Mitigating semantic injection is explicitly
out of scope on the map — agents reading tracker content with permissions
bypassed is the product's premise. This states the stance and guards it against
drift; it does not redesign it.

## Checks

The full AGENTS.md chain, green with `RUSTFLAGS=-D warnings` and
`RUSTDOCFLAGS=-D warnings`: fmt, clippy `--all-targets --all-features`,
`--lib --bins --examples` (430 + 24), `skill_docs` (20), `devcontainer_prebuild`,
`toolchain_pin`, `offline_green`, `live_fetch -- common::`, and `cargo doc`.

## Summary by Sourcery

Document and test the security posture for tracker-authored content injected into agent prompts.

Enhancements:
- Document the tracker-content injection posture, clarifying that tracker-authored titles and subsequent ticket content reach the agent verbatim while escaping protects transport mechanics rather than meaning.
- Add documentation-to-runtime drift guards that verify tracker prose survives launch parsing unchanged and that every free-text context field is explicitly covered by the README posture.

Documentation:
- Explain the tracker content exposed to agents, the limits of escaping, and the requirement to review tracker entries before launching workflows.

Tests:
- Add tests pinning the README security guidance and validating the launch context's free-text fields against the documented injection surface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1268,8 +1268,9 @@ gives away as much again. Don't point `wf` at a repo you have not read.
 
 **The tracker's text is an input to the agent, and only its mechanics are
 defended.** A launch embeds what the picker read straight into the prompt it
-execs: the map's title and the ticket's title, verbatim, in that `ctx:` block.
-The skill's own first move is then to read the ticket — body, comment trail and
+execs: the map's title and the ticket's title, verbatim, in the
+[`ctx:` block](#launching) it hands the skill. The skill's own first move is
+then to read the ticket — body, comment trail and
 all — so the surface is wider than the block: everything anyone can write on a
 tracker you point `wf` at reaches an agent running with permissions bypassed,
 and under `wf-auto` there is nobody reading along. Titles are the part `wf`
@@ -1284,13 +1285,13 @@ quote. What nothing here can check is what the text *says*. A title that reads
 like an instruction is still just a title, and an agent that reads it is being
 talked to by whoever wrote it.
 
-So the posture is [#73](https://github.com/blooop/wayfinder/issues/73)'s, one layer out: **don't point `wf` at a
-tracker you have not read.** That trade was about hostile code in a repo you launch into;
-this is hostile prose on a ticket you launch from, and the same answer covers
-both, because a run holding your `~/.claude` and your `GH_TOKEN` is one you had
-to decide to start. Reading the rows is the whole check, and the picker is
-where it is cheapest: every title that will reach the prompt is on the screen
-you pick from.
+So the posture is [#73](https://github.com/blooop/wayfinder/issues/73)'s, one
+layer out: **don't point `wf` at a tracker you have not read.** That trade was
+about hostile code in a repo you launch into; this is hostile prose on a ticket
+you launch from, and the same answer covers both, because a run holding your
+`~/.claude` and your `GH_TOKEN` is one you had to decide to start. Reading the
+rows is the whole check, and the picker is where it is cheapest: every title
+that will reach the prompt is on the screen you pick from.
 
 ### Working while you are away
 
--- a/README.md
+++ b/README.md
@@ -1266,6 +1266,32 @@ open ([#73](https://github.com/blooop/wayfinder/issues/73)); the repos this
 exists to serve want `network=host`, X11 and device passthrough anyway, which
 gives away as much again. Don't point `wf` at a repo you have not read.
 
+**The tracker's text is an input to the agent, and only its mechanics are
+defended.** A launch embeds what the picker read straight into the prompt it
+execs: the map's title and the ticket's title, verbatim, in that `ctx:` block.
+The skill's own first move is then to read the ticket — body, comment trail and
+all — so the surface is wider than the block: everything anyone can write on a
+tracker you point `wf` at reaches an agent running with permissions bypassed,
+and under `wf-auto` there is nobody reading along. Titles are the part `wf`
+itself hands over, and a title is writable by anyone who can open an issue.
+
+**The escaping is airtight and the meaning is not.** The block is one argv
+entry built by a JSON serializer, quoted once more for the single seam that
+reparses it, so no title can end the argument early, add a flag, or reach a
+shell — a real `sh` rebuilds a launch's argv byte for byte in the test suite,
+against a title carrying a single quote, a command substitution and a double
+quote. What nothing here can check is what the text *says*. A title that reads
+like an instruction is still just a title, and an agent that reads it is being
+talked to by whoever wrote it.
+
+So the posture is [#73](https://github.com/blooop/wayfinder/issues/73)'s, one layer out: **don't point `wf` at a
+tracker you have not read.** That trade was about hostile code in a repo you launch into;
+this is hostile prose on a ticket you launch from, and the same answer covers
+both, because a run holding your `~/.claude` and your `GH_TOKEN` is one you had
+to decide to start. Reading the rows is the whole check, and the picker is
+where it is cheapest: every title that will reach the prompt is on the screen
+you pick from.
+
 ### Working while you are away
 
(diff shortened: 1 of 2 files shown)


blooop (commented):

Two-axis review of #195 at head 27b41bd (merge-base 4135446, `git diff 4135446...HEAD`, README.md +27 / tests/skill_docs.rs +211). Preflight: diff non-empty, all checks green. Each axis ran as an independent subagent; findings are not merged or reranked across axes. Both axes independently ran the 20-test `skill_docs` binary against the branch tree and saw it green.

## Standards

**Findings (nits only, ordered by severity):**

1. **Duplicated Code (minor)** — `tests/skill_docs.rs:892-901`: `launch_with_planted_prose` re-implements the `split_once(" ctx: ")` extraction (expect message included) that `launched_block` (lines 280-287) already owns. If the prompt's `ctx:` marker ever moves, two extraction sites must move with it. Mitigating: this helper needs both the prompt (for failure messages) and the block from *one* launch, which `launched_block` cannot give it, and the file has precedent for annotated small duplication (`keys_of`) — though this instance carries no such annotation. A one-line comment or a shared `(prompt, block)` splitter would close it.
2. **Duplicated lead string (trivial)** — `tests/skill_docs.rs:600` and `844-845` both spell the magic lead `"**The tracker's text is an input"`; the pin test could call `posture_paragraph()`. Drift fails both tests loudly anyway, so purely cosmetic.

**Candidates refuted before posting:** the README text equality-pinned inside the test is the repo's explicit house pattern (same treatment as the exec-seam and ctx-contract pins), not duplication; the phrase-sampling at lines 995-999 is safe because the sampled paragraph is separately equality-pinned whole; `string_leaves`'s array/scalar arms are exercised (`prs[0]`), not speculative; reading the block through the binary's own launch path is the file's stated design rule, not feature envy.

**Axis recommendation: pass-with-nits.** No blocking findings.

## Spec

**The diff delivers the spec (ticket #193).**

- *"a README paragraph in the isolation/security section... beside the #73 trade"* — delivered: three paragraphs (README.md:1269-1294 on the branch) sit immediately after the #73 "not a security boundary" paragraph, in the same isolation section.
- *"naming the surface (issue titles and bodies reach the agent's prompt)"* — delivered, and more precisely than the spec's parenthetical: titles via the `ctx:` block, bodies/comment trails via the skill's first read ("The skill's own first move is then to read the ticket — body, comment trail and all"). Matches what `src/launch.rs` actually embeds (the two titles only).
- *"what is and is not defended (argv/JSON escaping yes, semantic injection no)"* — delivered ("The escaping is airtight and the meaning is not").
- *"the operator guidance"* — delivered ("don't point `wf` at a tracker you have not read", plus the picker-screen observation).
- *"the `skill_docs.rs`-style binding... so the paragraph cannot silently drift from what the launch actually embeds"* — substantively delivered: `tracker_prose_reaches_the_agent_unaltered` goes through the production path (`wf::launch::plan` → `agent_argv()`) and checks the *parsed* block, and the closed-world test walks every string leaf of a real launch's block.

**One minor finding**, against *"so the paragraph cannot silently drift from what the launch actually embeds"*: the closed-world test's non-marker arm accepts any string matching `is_ascii_alphanumeric() || "._/-"`. Its doc comment claims "a branch name ... fails here" (tests/skill_docs.rs:961-964), but a realistic branch name (`fix/ignore-previous-instructions`) is entirely within that charset and would pass silently — the guard discriminates by fixture *value*, not field provenance, so identifier-shaped free text could slip through if a future author picks a bland fixture value. Also, only the ticket-aim shape is walked; a payload added to a map-stop launch would escape. Both are hypothetical future drift, not present-day gaps — every field the block carries today is covered — so non-blocking. The overclaiming doc comment is the actionable part.

Absence of any mitigation is correctly out of scope per the parent map (#182).

**Axis recommendation: pass-with-nits.** No blocking findings.

## Verdict

**Approve.** No blocking findings on either axis. The nits (the duplicated `ctx:` extraction, the duplicated lead string, and the closed-world doc comment overclaiming what the charset arm catches) are optional polish and do not gate the merge. Posted as a comment because GitHub refuses an approve state from the PR author's own account; this written verdict is the gate.

PR #196, opened 2026-08-22
Every offline test binary is enrolled or the build fails

### LINKED ISSUE #189: Every offline test binary is enrolled or the build fails
## Question

Guard test-binary enrolment: a `tests/*.rs` that no workflow runs must fail the build.

`ci.yml` names its offline test binaries one by one, and `offline_green.rs` guards only `tests/live_*.rs` (`offline_green.rs:43`). A new `tests/foo.rs` without a `live_` prefix compiles under the `--no-run` step and is *run by nothing, flagged by nothing* — the exact "an assertion nothing runs is not an assertion" hole this repo closes everywhere else (ci.yml:105 says so), reopened one meta-level up. AGENTS.md:194 admits the workflow half is unchecked even for live tests.

Fix shape: extend `offline_green.rs` (or a sibling files-as-behavior test) to enumerate `tests/*.rs` and require each to be either `live_`-prefixed (and thus `#[ignore]`-guarded and named in `live.yml`/`devlaunch-contract.yml`) or named in a `cargo test --test <name>` line of `ci.yml` — read off the workflow file the way `devcontainer_prebuild.rs` already reads workflows. Include the self-check-on-the-checker the house style expects.

blooop:
Closes #189

`ci.yml` names its offline test binaries one by one, and `offline_green.rs` guarded only the *attribute* half of enrolment — every test in `tests/live_*.rs` carries `#[ignore]`. The other half was prose in AGENTS.md, which said in as many words that "the workflow half is still on you".

So a `tests/foo.rs` with no `live_` prefix wanted no `#[ignore]`, compiled cleanly under the `--all-targets --no-run` step, and was selected by no workflow: **run by nothing and flagged by nothing.** Compiling kept it looking tended, which is why the gap was invisible. That is the "an assertion nothing runs is not an assertion" rule this repo invokes at four separate `ci.yml` steps, reopened one meta-level up — a whole binary of assertions quietly not running rather than one test.

## What now fails the build

Four guards, in the same file and for the same reason the one already there is: the seam is source text, because no run of the tests can see a binary that never runs.

- **Every test binary under `tests/` is named by a command something runs.** Both forms cargo auto-discovers — `tests/<name>.rs` and `tests/<name>/main.rs`. Three files count as runners: `ci.yml`, `live.yml`, and a `pixi.toml` task the contract workflow calls. The failure lists the orphans and what each runner is for.
- **No `--test` argument names a binary that does not exist.** The other direction, and the one a rename leaves behind. Such a line does make `cargo test` fail — but only in the workflow holding it, and `live.yml` runs after a merge while the contract workflow's red runs already mean the other repo moved. Read here it lands at the commit that renamed the file.
- **The pixi manifest is a runner only while a workflow reaches it.** `pixi.toml` has no trigger of its own; `devlaunch-contract.yml` calling `contract` is the entire reason a binary named there is run. Drop that call and the sweep would go on reporting those binaries as enrolled — the exact lie this file exists to catch, wearing the guard's own badge. So the delegation is asserted.
- **A binary's name stays the one cargo derives from its path.** A `[[test]]` table can call any file anything, which would part the two facts the sweeps compare and report *both* directions wrong at once: the file demanded and never enrolled, the name enrolled and matching no file.

## Self-checks on the checker

Both sweeps ask whether a name appears, so anything that over-reports enrolment turns the forward sweep into a green light wired to nothing. Two limits of the scan are pinned on text written inside the test rather than trusted against the real files — a comment enrols nothing (these workflows are mostly prose, and the prose names test binaries), and `--test-threads` is not a binary called `threads` — plus the direct check that a name no file writes comes back unenrolled.

One more, which turns an existing admission into a failure: `live_sources`' doc comment said a directory-style `tests/live_x/main.rs` would sit outside its walk, "or the guard goes quietly blind to it". Now that a walk enumerates every real target, the two are compared. A file the ignore walk never opens is a file it reports as gated — the quiet direction, so it fails loudly instead.

## Proving preventive guards

On a clean tree all of these pass, which makes "it passes" worth nothing. Each was run red against the tree state it exists to reject before being trusted: an unenrolled `tests/ghost_binary.rs` **and** a `tests/ghost_dir/main.rs` (both reported); a stale `--test renamed_away` in `ci.yml`; the comment filter deleted from the scan; the contract calls swapped out of `devlaunch-contract.yml`; a hand-declared `[[test]]` in `Cargo.toml`; and a `tests/live_ghost/main.rs`. Then reverted.

The `[[test]]` case was the most instructive: staging it renamed the target and `cargo test --test offline_green` stopped resolving at all, so `ci.yml`'s own step would have failed. That is the hazard the guard describes, demonstrated by accident.

## Docs moved because the code made them false

`AGENTS.md`'s "The workflow half is still on you" is no longer true, so it changes here rather than later, and gains a section stating the wider rule — it holds for any `tests/*.rs`, not only the live ones. `ci.yml`'s step name and `offline_green`'s module doc each described the attribute half only.

Worth noting what this cannot do for itself: it reports its own orphaning, which is worthless — an unenrolled `offline_green` is a guard nothing runs. Only `ci.yml`'s existing step makes it speak, and the module doc says so.

Full local chain green, `cargo fmt` first, with `RUSTFLAGS=-D warnings` and `RUSTDOCFLAGS=-D warnings`.

## Summary by Sourcery

Enforce that every discovered test binary is both correctly named and executed by an enrolled workflow, failing the build when test coverage is silently omitted.

New Features:
- Add build-time validation that every test binary under `tests/` is enrolled by an executing CI, live-test, or contract workflow command.
- Validate both directions of test enrollment so stale workflow references and untracked test binaries fail early.
- Verify contract task reachability, Cargo test-target naming assumptions, and directory-style live test coverage.

Bug Fixes:
- Prevent test binaries from compiling successfully while being run by no workflow.
- Prevent ignored or stale workflow enrollment references from going undetected until later workflow execution.

Enhancements:
- Strengthen `offline_green` with source-based enrollment checks and self-tests for parser behavior and guard blind spots.
- Update repository guidance and guard documentation to describe the enforced enrollment contract.

CI:
- Make the existing offline guard CI step enforce enrollment of every test binary.

Documentation:
- Document that all test binaries must be explicitly enrolled by a workflow or reachable contract task.

Tests:
- Add coverage for orphaned binaries, stale `--test` arguments, comment and option parsing, unreachable Pixi tasks, hand-declared test targets, and directory-style live binaries.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,14 +123,24 @@ jobs:
       - name: The Rust pin has one home
         run: cargo test --locked --test toolchain_pin
 
-      # The guard on the exclusion every step around it depends on: the live
-      # binaries are safe to leave out of the unit-test step only because each
-      # of their tests carries `#[ignore]`, and a new one that forgets it is
-      # invisible both here and in `live.yml`, which selects with `--ignored`.
-      # This walks `tests/live_*.rs` and reads the attributes off the source,
-      # which is the only place the omission is visible — running the tests
-      # cannot see it, since running them is the symptom.
-      - name: The live tests are all gated
+      # The guard on the exclusion every step around it depends on, and on the
+      # enrolment that makes the steps above mean anything. The live binaries
+      # are safe to leave out of the unit-test step only because each of their
+      # tests carries `#[ignore]`, and a new one that forgets it is invisible
+      # both here and in `live.yml`, which selects with `--ignored`. Reading the
+      # attributes off the source is the only place that omission is visible —
+      # running the tests cannot see it, since running them is the symptom.
+      #
+      # It also walks every test binary under `tests/` and fails on any that no
+      # command in this file, in `live.yml` or in a `pixi.toml` task the contract
+      # workflow calls ever runs (#189). That closes the hole one level above the
+      # rest of this job: the `--no-run` step below keeps a new `tests/foo.rs`
+      # compiling, so it looks tended while running nowhere — a whole binary of
+      # assertions nobody is told about, which is the "an assertion no workflow
+      # runs is not an assertion" rule the steps above each invoke, applied to
+      # the steps themselves. This step is the reason the guard is not its own
+      # first victim.
+      - name: The live tests are gated and every test binary is enrolled
         run: cargo test --locked --test offline_green
 
       # The live fixture's own diagnostic — offline like its three siblings
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -189,9 +189,36 @@ shims. Three consequences worth knowing before you touch any of this:
   `#[ignore]` it makes a bare `cargo test` fail on any machine lacking what it
   wants; with it and nothing else, it is a test that never runs anywhere.
   `live.yml` names its test binaries one by one for that reason — enrolment is
-  a deliberate line, not a side effect. The attribute half is checked rather
-  than remembered: `tests/offline_green.rs` walks `tests/live_*.rs` and fails
-  on any test that is missing it. The workflow half is still on you.
+  a deliberate line, not a side effect. Both halves are checked rather than
+  remembered, and by the same file: `tests/offline_green.rs` walks
+  `tests/live_*.rs` and fails on any test missing the attribute, and it walks
+  every test binary under `tests/` and fails on any that no workflow names.
+  A deliberate line, still — the guard makes you write it, it does not write
+  it for you.
+
+### Every test binary is enrolled or the build fails
+
+The rule is wider than the live tests and it holds for any `tests/*.rs`: a test
+binary that no command anywhere runs fails `offline_green` (#189). Three places
+count as running one, and a binary has to be named in one of them — `ci.yml`
+for an offline check, `live.yml` for the gh-live set, or a `pixi.toml` task the
+contract workflow calls for anything wanting a chosen `dl`. Four things to know
+before adding a test file:
+
+- **Compiling is not running.** `ci.yml`'s `--all-targets --no-run` step keeps
+  every binary building, so rot is caught either way — which is exactly why the
+  gap was invisible. A new `tests/foo.rs` compiled, ran nowhere, and looked
+  fine.
+- **A comment does not enrol anything.** The guard strips full-line comments
+  before looking, because the workflows here are mostly prose and the prose
+  names test binaries.
+- **A rename is caught from the other side too.** A `--test` argument naming no
+  file fails at the commit that moved the file, rather than in whichever
+  workflow was holding the stale line.
+- **`pixi.toml` counts only while a workflow still calls its task.** The
+  manifest has no trigger of its own; `devlaunch-contract.yml` calling
+  `contract` is the whole reason a binary named there is run, and that call is
+  asserted rather than assumed.
 
 ## The devlaunch contract
 
--- a/tests/offline_green.rs
+++ b/tests/offline_green.rs
@@ -15,9 +15,29 @@
 //! attribute, which no run of the tests can notice — the failure is precisely
 //! that they run when they should not, or never run at all.
 //!
+//! **Enrolment is the other half of the same promise (#189), and it lives here
+//! too.** The attribute decides whether a test runs in the default set; a line
+//! in a workflow decides whether the binary holding it is ever run at all, and
+//! that half was prose in AGENTS.md handed back to whoever wrote the test. It
+//! failed the same way and worse: a `tests/foo.rs` with no `live_` prefix wants
+//! no `#[ignore]`, compiles under `ci.yml`'s `--no-run` step, and is selected
+//! by no workflow — run by nothing and flagged by nothing, a whole binary of
+//! assertions rather than one test. "An assertion nothing runs is not an
+//! assertion" is the rule this repo closes everywhere else; this is that hole
+//! one meta-level up, and it is the same kind of invisible-to-a-test-run
+//! failure as the attribute, so it is read off the same kind of text.
+//!
 //! This binary is deliberately not named `live_*`. It is not a live test, and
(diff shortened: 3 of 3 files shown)


blooop (commented):

Two-axis review at merge-base `4135446` (diff `4135446...0095ba8`). Both axes ran as independent fresh-context subagents; each verified the guard by executing it against the PR head and by mutation (injecting the tree states it exists to reject).

## Standards

**Verified by execution.** The guard was run standalone at the PR head: all 8 tests pass against the real repo state. Mutation tests confirm the claimed failure modes are real and loud: an orphan `tests/foo.rs` fails the forward sweep, a stale `--test gone_away` in ci.yml fails the reverse sweep, and a directory-form `tests/live_x/main.rs` fails the dedicated blind-spot check. The pixi-delegation check correctly parses `pixi run -e <env> contract` in devlaunch-contract.yml and extracts both `contract` task definitions from pixi.toml (`--test-threads=1` confuses neither scanner).

**Repo standards compliance.** Exemplary. The file follows the documented `devcontainer_prebuild.rs` mould (source-text seam, offline, self-checked scanner limits), doc comments carry the why including rejected alternatives (`syn` considered and declined, with the limits then named and pinned), and AGENTS.md plus the ci.yml step comment are updated in the same PR to match the new behavior.

**Findings:**

1. **Non-blocking — false enrolment via non-comment prose** (`tests/offline_green.rs:333-350`, `binaries_named_in`). Comment-stripping handles `#` lines only, but YAML carries prose in other shapes: confirmed by mutation that a step title `- name: step that mentions --test foo in a title` silently enrols an orphan `tests/foo.rs` — a green light in the exact direction the guard exists to close. Same class: an `echo "... --test foo"` in a run block, or `pixi run contract` quoted inside an echo satisfying the delegation check. The doc comment "reads commands, not the prose around them" overclaims; the self-check pins only the `#`-comment case. Low likelihood, but worth either a named limit in the doc comment (house convention: every scanner limit is named) or restricting the scan to `run:` lines.
2. **Nit — `--test=name` equals form unrecognised** (`tests/offline_green.rs:344`). Cargo accepts it; the guard would report the enrolled binary as an orphan (confirmed by mutation — loud, fail-safe, but with a misleading "run by nothing" message). Unnamed as a limit; one line in the doc comment would do.
3. **Nit — hyphenated target names** (`tests/offline_green.rs:346-348`). `tests/foo-bar.rs` is a legal cargo target; the name-boundary scan stops at `-`, producing a false orphan. Fail-safe, none exist today.

**Fowler baseline.** Clean. The three `read_dir`-collect-sort walks are borderline Duplicated Code, but each filters differently and the house style prefers this explicitness; no other smell applies.

Axis verdict: **pass with comments** — no blocking findings; #1 is the one worth acting on.

## Spec

Spec: ticket #189. Verified against the diff and executed at PR head in a scratch worktree (clean tree: 8/8 pass; then three constructed violations, each red).

**1. Enumeration — implemented.** `test_targets()` (tests/offline_green.rs:265) walks `tests/`, admitting both `tests/<name>.rs` and `tests/<name>/main.rs`, excluding `tests/common/`. The directory-form claim is real: an injected `tests/zz_dir/main.rs` failed the run, both in the enrolment sweep and (for `live_*` dirs) via the ignore-walk blind-spot check (offline_green.rs:557).

**2. Enrolment — implemented, empirically confirmed.** All nine targets at head are genuinely enrolled: five in ci.yml (lines 107–156), three in live.yml:90–93, and `live_devlaunch` via pixi.toml `contract` tasks that devlaunch-contract.yml:100–112 calls. An injected unenrolled `tests/zz_unenrolled.rs` fails the forward sweep (offline_green.rs:393); an unignored `tests/live_zz.rs` fails the ignore walk.

- **Non-blocking:** the ticket says a non-live binary must be "named in a `cargo test --test <name>` line of `ci.yml`"; the PR accepts any of three runners, so a non-live `tests/foo.rs` named only in live.yml's block would count as enrolled while `-- --ignored` deselects its unignored tests — enrolled on paper, 0 tests executed. Obscure (requires a deliberate wrong-file edit), but the sweep reads "named", not "run".
- **Note:** "named in live.yml/devlaunch-contract.yml" is implemented differently — devlaunch-contract.yml contains no `--test` line; the PR correctly routes this through pixi.toml plus a delegation assertion (offline_green.rs:501). Also stricter than the ticket's letter: a `live_` prefix alone no longer suffices — the binary must appear in a runner, which is exactly the AGENTS.md:194 gap the ticket cites. Both faithful to intent.

**3. Self-check on the checker — present and meaningful** (offline_green.rs:414): parser pinned on constructed text (comments enrol nothing; `--test-threads` is not a binary) plus a direct negative probe.

**4. Beyond spec — all justified.** Reverse sweep (:382), `[[test]]` hand-declaration guard (:531), pixi-delegation check, directory-form checks: each closes a way the new guard could over-report enrolment, i.e. the ticket's own failure mode. AGENTS.md rewrites exactly the line the ticket quotes. ci.yml changes are comment + step-name only; the `run:` line is unchanged — no existing check weakened.

**5. Unimplemented elements — none.**

Axis verdict: **pass with comments** — no blocking findings; the live.yml/`--ignored` enrolment-without-execution loophole is the only substantive gap.

## Verdict

**Approve.** No blocking findings on either axis. Two non-blocking findings worth a follow-up (or a doc-comment limit line in this PR): Standards #1 (a `- name:` title or `echo` string can falsely enrol a binary — the scan reads all non-`#` lines, not just `run:` commands) and Spec's live.yml cross-enrolment loophole (a non-live binary named only in live.yml counts as enrolled while `--ignored` runs none of its tests). Both are fail-open only under unusual, deliberate text; every failure direction that exists in the tree today fails loud. Recorded verdict: Approve — GitHub review state may show as Comment where self-approval is refused.

PR #197, opened 2026-08-22
projects.json saves are atomic and single-seamed

### LINKED ISSUE #185: projects.json saves are atomic and single-seamed
## Question

Make `projects.json` saves atomic and route every writer through one seam.

`save` is a bare `std::fs::write` — truncate-then-write — at `projects.rs:255`, and there are three writers: the UI path (`picker.rs:469–472`, `546–567`), the discovery task (`refresh.rs:327–329`), and any second `wf` instance. A torn read parses as corrupt, `load_or_default` (`projects.rs:234–239`) falls back to an empty default, and that instance's next save rewrites the registry from empty — silently losing every checkout, seed, and session record.

Fix shape: write to a temp file in the same directory and rename over the target (atomic on the same filesystem); consider read-merge-write so two live instances stop last-writer-winning each other's registrations. While in this seam, fix the asymmetry the review flagged: the startup save is fatal (`cache.save(&cache_path)?` at `picker.rs:470/472` aborts the launcher on a full disk or read-only HOME) while the handover save is deliberately best-effort (`picker.rs:566`) — the cache's own doc says a failed save should be reported, not fatal.

blooop:
Closes #185.

`projects.json` was written with a bare `std::fs::write` — truncate, then write — and three writers share it: the startup registration, the discovery task's findings, and the handover stamp, plus any second `wf` the human has open. A reader landing in the truncate window parsed nothing, `load_or_default` reported "no projects", and that instance's next save wrote the empty registry back over every checkout, seed and session on the machine.

## What changed

**Saves are atomic.** The cache is written to a hidden scratch file beside it and renamed over the target — same directory, so the rename is atomic on the filesystem it lands on. A reader now sees the whole previous registry or the whole new one, and there is no third answer. The scratch name carries the writing process and a counter within it, so two savers racing cannot hand each other half a registry through a shared temp file, and a failed rename cleans up rather than littering a cache directory nobody sweeps. No `fsync`: durability across a crash is deliberately not bought for a cache whose whole cost of loss is re-opening each project once, and paying for it would put a disk flush on the path to the first frame.

**There is one write seam, and it re-reads.** `ProjectsCache::update(path, edit)` loads the registry as it is on disk *now*, applies the edit to that, and persists it; `save` is private behind it. The signature is the point — there is no longer any way to hand the cache a copy loaded minutes ago, which is how a registration used to disappear with nothing failing. It is not a lock and does not claim to be: it shortens the clobber window from "as long as this `wf` has been running" to one read plus one write. All three production writers go through it, and an edit that reports no change writes nothing, which is what the handover path needs at the moment the terminal is already on its way to the agent.

**A failed startup save no longer aborts the launcher.** `cache.save(&cache_path)?` at startup meant a full disk or a read-only `HOME` stopped `wf` from starting, while the handover save was deliberately best-effort — an asymmetry the review flagged, against a doc comment that says a failed save should be reported. `ProjectsCache::at_startup` now returns the registry plus the outcome of persisting it, and the launcher prints the failure and runs on the registry it just loaded. Printed before the alternate screen goes up, so it lands on the primary buffer and is still there when `wf` puts that buffer back.

## Tests

Red before green on each slice:

- a reader loop against a writer hammering saves: 293 of 717 reads saw a registry that was not there before, none after;
- a hard link standing in for a reader holding the old file, deterministically proving the file is replaced rather than rewritten in place;
- nothing left in the cache directory beside the cache itself;
- an edit through the seam landing on a registration another instance made after this one loaded its copy;
- an edit that changes nothing leaving the file byte-identical (hand-written compact JSON as the detector — a rewrite comes back pretty-printed);
- a startup whose cache directory cannot exist still yielding a usable, registered registry, with the failure reported rather than raised.

Full chain from AGENTS.md green: fmt, clippy `--all-targets` under `-D warnings`, `--lib --bins --examples` (438 + 24), `skill_docs`, `devcontainer_prebuild`, `toolchain_pin`, `offline_green`, `live_fetch -- common::`, and `cargo doc` under `-D warnings`.

`tests/live_discovery.rs` registers through the seam now, since `save` is no longer public.

## Summary by Sourcery

Make project-cache updates atomic and resilient to concurrent writers without preventing the launcher from starting when persistence fails.

Bug Fixes:
- Prevent corrupted or stale project registries from erasing checkout and session records during concurrent saves.
- Allow launcher startup to continue when the projects cache cannot be persisted, while reporting the save failure.

Enhancements:
- Centralize project-cache writes behind a read-modify-write update seam so startup, discovery, and handover updates preserve the latest on-disk registry.
- Replace in-place cache writes with atomic same-directory replacement and avoid unnecessary writes when an edit makes no changes.

Tests:
- Add coverage for atomic reads, replacement semantics, concurrent-update preservation, startup save failures, no-op edits, and temporary-file cleanup.

### DIFF
--- a/src/picker.rs
+++ b/src/picker.rs
@@ -461,15 +461,18 @@ pub async fn run_picker() -> Result<()> {
     let here = projects::discover_checkout(&cwd).await;
     let cache_path =
         projects::default_cache_path().context("cannot resolve the XDG cache directory")?;
-    let mut cache = ProjectsCache::load_or_default(&cache_path);
-    // Accretion needs a matching forget: a checkout that has been deleted must
-    // stop offering itself as somewhere an agent could run.
-    let pruned = cache.prune_missing();
-    if let Some((path, slug)) = &here {
-        cache.register(path.clone(), slug.clone());
-        cache.save(&cache_path)?;
-    } else if pruned {
-        cache.save(&cache_path)?;
+    let (cache, saved) = ProjectsCache::at_startup(
+        &cache_path,
+        here.as_ref()
+            .map(|(path, slug)| (path.as_path(), slug.as_str())),
+    );
+    // Reported, not fatal — the launcher runs on the registry it just loaded
+    // whether or not the machine can remember it, exactly as the handover write
+    // does. Printed *here*, before the alternate screen goes up, so it lands on
+    // the primary buffer: visible on the way in, and still there when `wf` puts
+    // that buffer back on the way out.
+    if let Err(err) = saved {
+        eprintln!("wf: could not save the projects cache: {err:#}");
     }
     let repos = cache.repos();
     // The head start (#28): the map numbers the last search found. Reading them
@@ -534,19 +537,19 @@ pub async fn run_picker() -> Result<()> {
             // it, and for everyone else launching is the only other act that
             // means "this is what I am working on".
             //
-            // Re-read before writing, exactly as the discovery task does: it
-            // writes the search's findings to this same file while the picker
-            // is up, so the copy loaded before the first frame is stale by
-            // now, and saving it would trade this stamp for next run's head
-            // start.
+            // Through the cache's write seam, which re-reads: the discovery
+            // task writes the search's findings to this same file while the
+            // picker is up, so the copy loaded before the first frame is stale
+            // by now, and writing it back would trade this stamp for next run's
+            // head start.
             //
             // Best-effort on purpose. A cache that will not write is not worth
             // refusing a launch over, and the only cost of losing this is one
             // project sitting lower in a list than it might have.
-            let mut cache = ProjectsCache::load_or_default(&cache_path);
-            let mut changed = cache.touch(launch.cwd());
-            // And record the conversation this launch is about to start, so a
-            // later run can offer the way back into it (#35).
+            //
+            // The session record rides along in the same edit — the
+            // conversation this launch is about to start, so a later run can
+            // offer the way back into it (#35).
             //
             // Written **here**, immediately before the terminal is restored
             // and the image replaced, because this is the last moment `wf`
@@ -555,16 +558,17 @@ pub async fn run_picker() -> Result<()> {
             // gets. A creation records nothing: it has no node to key on until
             // its skill files one.
             //
-            // Best-effort like the stamp above it, and for a smaller cost: a
-            // record that fails to write means the resume row is missing next
-            // time, not that anything is wrong with the launch.
-            if let Some(session) = launch.session() {
-                cache.record_session(session);
-                changed = true;
-            }
-            if changed {
-                let _ = cache.save(&cache_path);
-            }
+            // Best-effort like the stamp it travels with, and for a smaller
+            // cost: a record that fails to write means the resume row is
+            // missing next time, not that anything is wrong with the launch.
+            let _ = ProjectsCache::update(&cache_path, |cache| {
+                let mut changed = cache.touch(launch.cwd());
+                if let Some(session) = launch.session() {
+                    cache.record_session(session);
+                    changed = true;
+                }
+                changed
+            });
             // The prompts the selected agent is about to run. `wf skills
             // install` links its skills directory at a *copy* of the bundle,
             // and a copy is a thing that can fall behind a `pixi global update
--- a/src/projects.rs
+++ b/src/projects.rs
@@ -155,6 +155,38 @@ impl Session {
     }
 }
 
+/// A scratch path in the same directory as `path`, for a write that will be
+/// renamed over it.
+///
+/// Same directory on purpose: a rename is only atomic within one filesystem,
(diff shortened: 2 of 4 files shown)


blooop (commented):

Two-axis review of #197 at head 0cee569 (merge-base 4135446 with `main`). Axes were run as independent fresh-context reviewers and are not merged or reranked.

## Standards

Repo standards (AGENTS.md, guard tests) — checked, no violations:

- The picker.rs source-text denylist (`the_picker_names_neither_a_subprocess_nor_reap`: bare `reap`, `fs`, `Command`) is not tripped — the diff routes all IO through `ProjectsCache`; main.rs is untouched, so its line-list guard needs no update.
- tests/live_discovery.rs keeps its `#[ignore = "live: ..."]`, so `offline_green` stays satisfied; no new live test was added, so no workflow enrolment is owed.
- House style (why-comments carrying design decisions, alternatives-considered) is followed well; `save` going private with `update` as the sole write seam matches the repo's "make illegal states unsayable" posture.

Findings:

- **Non-blocking — temp file leaks on a failed write** (src/projects.rs, `save`): only the *rename* failure path removes the scratch file. If `std::fs::write(&temp, json)` itself fails after creating the file (ENOSPC — the very scenario `at_startup` now tolerates), a hidden `.projects.json.<pid>.<n>.tmp` is left in a directory the doc itself says "nobody ever cleans". The litter test (`a_save_leaves_nothing_behind_beside_the_cache`) covers only the success path, so the stated invariant is narrower than the doc claims. Wrap the write failure in the same `remove_file(&temp).ok()` cleanup.
- **Non-blocking — flaky by construction**: `a_reader_in_another_instance_never_sees_a_half_written_registry` asserts `reads > 0`, which fails if the reader thread is preempted for the writer's full 150 saves — unlikely but scheduling-dependent, in the CI-blocking lib suite. The deterministic hard-link test already pins the mechanism; consider a loop-until-first-read handshake or a fixed read count.
- **Nit — doc overclaim** (`update` doc): "there is no signature here that accepts a stale copy" — a closure can still capture a stale `ProjectsCache` and assign it over `cache`. The seam removes the easy mistake, not the possibility; the repo's guard-test docs are scrupulous about stating exactly that kind of limit, and this one should match.
- **Nit** — `at_startup(path, here: Option<(&Path, &str)>)` is a small data clump, and picker.rs's `map(|(path, slug)| ...)` shim exists only to feed it; it mirrors `discover_checkout`'s existing tuple, acceptable as-is.

Correctness of the atomicity itself: same-directory rename, pid-plus-counter scratch name, and the documented no-fsync trade are sound; Windows rename semantics are out of scope (no Windows build in recipe.yaml). Fowler baseline: no other hits.

## Spec

Spec: ticket #185. The diff meets the spec.

- **"write to a temp file in the same directory and rename over the target (atomic on the same filesystem)"** — Met. `save` (src/projects.rs:375-390) writes to `scratch_beside(path)` — built via `path.with_file_name(...)`, provably the same directory — then renames over the target, cleaning up the temp on rename failure. Temp names embed pid + a per-process counter, covering the picker's discovery-task/main-thread concurrency. No fsync, but the spec asks only for atomicity, and the trade is documented. Informational only.
- **"route every writer through one seam"** — Met. `save` is now private; the only public write path is `ProjectsCache::update` (src/projects.rs:298). All three writers named in the spec go through it: startup via `at_startup` (src/picker.rs:464), handover (src/picker.rs:564), discovery task (src/refresh.rs:332); a second `wf` instance runs this same code. Grepped all of src/ and tests/ for `fs::write`, `.save(`, and `projects.json`: tests/live_discovery.rs:34 now uses `update`; the other live tests only read; direct `save` calls remain only inside `projects.rs`'s own `mod tests`, which test the seam itself. No `[[bin]]`/`[[example]]` targets exist. No bypass found.
- **"the startup save is fatal ... a failed save should be reported, not fatal"** — Met. src/picker.rs:474-476 prints the failure to stderr and continues; no `?` on the save. `a_startup_that_cannot_write_still_gives_the_launcher_a_registry` (src/projects.rs:766-793) pins that `at_startup` returns a usable registry plus a reported `Err`.
- **"consider read-merge-write so two live instances stop last-writer-winning"** — Implemented as read-modify-write: `update` re-reads disk immediately before every write. This eliminates the described mode (writing back a copy "loaded minutes ago") but is not a lock — two processes whose read+write windows overlap can still lose one edit. The doc states this honestly. The spec said "consider", so this is satisfied; residual race noted, non-blocking.
- **Tests pin the behaviors** — Yes: concurrent writer/reader stress test asserting zero torn reads, deterministic hard-link rename-not-rewrite test, stale-copy merge test, non-fatal-startup test, no-write-when-unchanged test, no-temp-litter test.

No blocking findings.

## Verdict

**Approve.** No blocking findings on either axis. Two non-blocking Standards findings worth a follow-up commit (temp-file cleanup on a failed write; the vacuity assert in the torn-read stress test) and two nits — none gate the merge.

PR #198, opened 2026-08-22
Never prune a skills copy directory wf cannot prove it made

### LINKED ISSUE #187: The skills mirror is pruned only when wf made it
## Question

Never prune a skills mirror wf cannot prove it made.

`install` runs `create_dir_all(target.mirror)` (`skills.rs:511–513`), which silently *adopts* a pre-existing `~/.claude/wf-skills` — a directory or a symlink to one that a user or another tool made — and then `prune_mirror` (`skills.rs:685–702`) deletes every entry not in `BUNDLED`/`.source`. The comment "everything in here is wf's by construction" (`skills.rs:679–681`) is false on that adoption path, and it contradicts the README's promise that "a skill of yours can never match however dead it looks."

Related over-claim in the same module: `is_ours` accepts `parent == bundle` (`skills.rs:721`), so with `WF_SKILLS_DIR` pointed at a user's own directory, a symlink the *user* hand-made into it under a non-BUNDLED name gets swept (`skills.rs:763–767`).

Fix shape: prune only a mirror that carries a `.source` record wf wrote (the existing provenance mechanism — absent it, treat the directory as Unmanaged: report and leave it, like the module already does for real directories under `skills/`); tighten `is_ours` so a link is wf's only when wf's own records say so, not merely because it points into the bundle. Tests for both adoption shapes.

blooop:
Closes #187

`wf skills install` ran `create_dir_all` on `~/.claude/wf-skills` and then pruned every entry in it that was not a bundled skill. `create_dir_all` cannot tell a directory it made from one it merely found, so a `wf-skills` a user or another tool already had — a directory, or a link into a dotfile tree — was silently **adopted** by the first install and then emptied by it. The comment saying "everything in here is `wf`'s by construction" was false on exactly that path, and it contradicted the promise the README makes: a name is `wf`'s only when `wf` put it there.

The same over-claim lived one function along. `is_ours`, the whole permission argument for removing a link, counted "points into the bundle" as proof `wf` wrote it — and the bundle is whatever directory `WF_SKILLS_DIR` names, so with it aimed at a directory of your own, a symlink you hand-made into it under a name this build does not ship was swept.

## What changed

**Provenance is evidence, not an assumption.** The copy directory now has three states rather than exists-or-not: nothing there (an install creates it, so what it then holds is `wf`'s), a real directory carrying the `.source` record `install` writes (the only shape prune may touch), or anything else — reported and left exactly as found, the way a real directory under `skills/` already is. `MIRROR` and `SOURCE` arrived in one commit, so a copy `wf` made has always carried a record; absence of one is a real answer.

`install` therefore returns one verdict for the run — `Installed::Done(..)` or `Installed::CopyUnmanaged` — because the links are written *into* that copy, so there is no such thing as installing five skills correctly beside a copy somebody else owns. `wf skills install` prints it once against the path in question and exits non-zero, and the other agent's install still proceeds. The record moves to the front of the install for the same reason it is now load-bearing: an install interrupted between the `create_dir_all` and the record would otherwise leave a copy the next one could not recognise as its own.

**A link where the copy goes is never `wf`'s.** `create_dir_all` takes a link to a directory without complaint, so every copy landed in somebody's tree and the prune deleted out of it. `wf` only ever makes a real directory there, checked before the record is consulted — a link carrying a record is a link carrying somebody else's file. The launch-time `refresh` is gated on the same proof; it writes into the copy on every `enter`, so it had the same reach.

**`is_ours` keeps only the shapes nothing else produces**: the relative link into a copy `wf` can show it made, and a path ending `share/wf/skills`, which is `wf`'s own packaging layout. An adopted copy withdraws the first — the link merely names it, so the name is somebody's own choice — while a copy that is *gone* keeps it, because a relative link into a `wf-skills` that is not there is residue nothing but `wf` leaves, and clearing that is what sweep is for. `sweep` no longer needs the bundle at all.

The cost is exact and deliberate: a link a pre-#110 `wf` wrote into a **checkout** bundle is no longer reclaimed by a launch or swept. That leaves a dangling link alone rather than deleting a link out of a directory of yours, which is the side to err on.

## Tests

Three new tests, one per adoption shape the ticket named, each red before the fix:

- a copy directory `wf` did not make is reported and left alone — their prompt intact, no record written into it that would claim it next time, and no link into it
- a copy directory that is a *link* is never `wf`'s, even one whose target carries a record — and no launch writes through it either
- a link into a directory `wf` was merely pointed at is not swept; nor is a link into an adopted copy — while the same shape over a copy that is gone still sweeps

The existing rename/prune/repoint tests stay as the regression that the tightening did not overshoot. Full chain from AGENTS.md is green: fmt, clippy, `--lib --bins --examples`, `skill_docs`, `devcontainer_prebuild`, `toolchain_pin`, `offline_green`, `live_fetch -- common::`, `cargo doc`. Driven for real against a scratch `CLAUDE_CONFIG_DIR`: Claude blocked with the message and the directory untouched, Codex installed beside it, exit 1, and a clean install once the directory is moved aside.

## Summary by Sourcery

Protect user-managed skill directories and links by requiring verifiable wf ownership before installation, refresh, pruning, or cleanup.

Bug Fixes:
- Prevent installs and launch-time refreshes from adopting or modifying unmanaged skill copy directories.
- Stop pruning links merely because they point into the configured skills bundle unless their destination is independently attributable to wf.

Enhancements:
- Track copy-directory provenance with an install record and distinguish absent, owned, and unmanaged copies.
- Report unmanaged copy directories once, leave them untouched, and preserve cleanup for links into owned or removed wf copies.

Documentation:
- Document the ownership and provenance rules for skill copy directories and symlinks.

Tests:
- Add regression coverage for pre-existing copy directories, symlinked copy directories, and links into unmanaged bundles or copies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,8 +67,11 @@ Every name carries the `wf` prefix because both `~/.claude/skills` and
 you have. Unprefixed, `tdd` and `review` are names `wf` would *squat on* rather
 than merely occupy: while it held one, you could not have your own. `wf skills
 install` clears the links an older `wf` left under its old names, and touches
-nothing else — it removes a link only when the link points into a `wf` bundle,
-so a skill of yours can never match however dead it looks.
+nothing else — it removes a link only where the link points at something nothing
+but `wf` writes: the copy below, or a path ending `share/wf/skills`, which is
+`wf`'s own packaging layout. A skill of yours can never match however dead it
+looks, and neither can one of yours that `wf` was merely *pointed at*: a link
+into whatever directory `WF_SKILLS_DIR` names is not evidence `wf` wrote it.
 
 `wf skills install` **symlinks**, so a name under either skills directory is
 `wf`'s only when `wf` put it there and a real directory can go on meaning
@@ -76,6 +79,15 @@ so a skill of yours can never match however dead it looks.
 `~/.claude/wf-skills` and `~/.codex/wf-skills`, **copies** of the package's
 bundle kept beside their links, reached *relatively* —
 `wf-tdd -> ../wf-skills/wf-tdd`.
+
+The copy directory is held to the same rule as a name under `skills/`, and it
+has to be, because `wf` *prunes* it — anything in there that this build does not
+ship is a leftover of an older one. So it prunes only a `wf-skills` it can show
+it made, by the record it writes inside it when it creates it. Find one already
+there carrying no such record — a directory or a link into your dotfiles, made
+by you or by another tool — and `wf skills install` reports it and stops, having
+written nothing anywhere, and every launch leaves it alone too. Move it aside
+and run the install again.
 
 That indirection exists for containers. An isolated Claude launch
 ([Isolation](#isolation-claude-runs-in-the-repos-devcontainer)) mounts your
--- a/src/main.rs
+++ b/src/main.rs
@@ -274,7 +274,23 @@ fn run_skills(install: bool) -> Result<()> {
             // ships under a new spelling, and clearing the old link first keeps
             // the two steps from racing over the same directory entry.
             let swept = skills::sweep(&bundle, &target)?;
-            let done = skills::install(&bundle, &target)?;
+            // A copy directory `wf` cannot prove it made stops the install
+            // whole, so it is reported once — against the path in question —
+            // rather than six times over skills none of which was touched.
+            let done = match skills::install(&bundle, &target)? {
+                skills::Installed::Done(done) => done,
+                skills::Installed::CopyUnmanaged => {
+                    let _ = writeln!(
+                        out,
+                        "  NOT INSTALLED — {} already exists and carries no record wf \
+                         wrote, so wf did not make it. Nothing was touched. Move it aside \
+                         (chezmoi may own it) and run this again",
+                        target.mirror().display()
+                    );
+                    blocked = true;
+                    Vec::new()
+                }
+            };
             for link in &swept {
                 let _ = writeln!(
                     out,
--- a/src/main.rs
+++ b/src/main.rs
@@ -273,7 +273,7 @@ fn run_skills(install: bool) -> Result<()> {
             // Before linking, not after: a swept name could be one this build
             // ships under a new spelling, and clearing the old link first keeps
             // the two steps from racing over the same directory entry.
-            let swept = skills::sweep(&bundle, &target)?;
+            let swept = skills::sweep(&target)?;
             // A copy directory `wf` cannot prove it made stops the install
             // whole, so it is reported once — against the path in question —
             // rather than six times over skills none of which was touched.
--- a/src/skills.rs
+++ b/src/skills.rs
@@ -370,6 +370,46 @@ pub fn installed_from(target: &Target) -> Option<PathBuf> {
     (!recorded.is_empty()).then(|| PathBuf::from(recorded))
 }
 
+/// What sits where the copy of the bundle goes, and whether `wf` can prove it
+/// put it there.
+///
+/// The distinction is the whole permission argument for [`prune_mirror`], which
+/// deletes every entry it does not recognise. `create_dir_all` cannot tell a
+/// directory it just made from one it merely found, so a `~/.claude/wf-skills`
+/// a user or another tool already had was *adopted* by the first install and
+/// then emptied by it (#187) — and the module's own promise, that a name is
+/// `wf`'s only when `wf` put it there, was false on exactly that path.
+///
+/// Provenance is the record [`install`] already writes, so the answer is
+/// evidence rather than an assumption: [`MIRROR`] and [`SOURCE`] arrived in one
+/// commit, so a copy `wf` made has always carried one.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum Mirror {
+    /// Nothing is there. An install creates it, so what it then holds is
+    /// `wf`'s — and a link into it is `wf`'s to reclaim.
+    Absent,
+    /// A directory carrying `wf`'s own record of the bundle it was copied
+    /// from. The only shape [`install`] may prune.
+    Ours,
+    /// Something is there that carries no such record. Reported and left
+    /// exactly as found, the way a real directory under `skills/` already is:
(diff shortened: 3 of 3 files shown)


blooop (commented):

Reviewed in fresh context against two independent axes, neither aware of the other's findings and neither merged or reranked. Fixed point: merge-base `4135446`, `git diff 4135446...HEAD` (README.md, src/main.rs, src/skills.rs; +285/-33). Preflight: ref resolves, diff non-empty, all five checks green.

## Standards

Repo standards are met. AGENTS.md's rules for this module hold — `BUNDLED` untouched, no `Route` changes — and the source-text guard tests (`tests/live_devlaunch.rs`'s self-`include_str!` around reap, `tests/skill_docs.rs`'s pinned README paragraphs) pin material this diff does not touch, so nothing is tripped. The heavy why-comment house style is matched, including the honest rewrite of the now-false `prune_mirror` comment (`src/skills.rs:760-767`). Fowler baseline is otherwise clean: `Installed` is an enum return rather than a sentinel, the tests funnel through one `done()` helper instead of duplicating the match, and both signature changes (`install`, `sweep`) were propagated to every caller.

1. **Non-blocking — "Nothing was touched" can be false.** `run_skills` calls `skills::sweep(&target)?` *before* `skills::install` (`src/main.rs:276`), and sweep with an `Adopted` mirror still removes links whose parent ends `share/wf/skills` (`src/skills.rs:808-822`). When install then returns `CopyUnmanaged`, the message at `src/main.rs:283-290` claims "Nothing was touched" — and the `for link in &swept` loop immediately below (`src/main.rs:294`) can print those removals directly underneath it. `Installed::CopyUnmanaged`'s doc ("Nothing was written, linked or removed anywhere", `src/skills.rs:552-555`) is true of `install` alone; the user-facing sentence covers the whole run and the README repeats it ("having written nothing anywhere"). Fix either way: inspect the mirror before sweeping too, or soften the two sentences. No user data is at risk — the swept shape is `wf`'s own packaging layout — but a PR whose subject is not over-claiming should not ship an over-claim.

2. **Nit — the interrupted-install comment overclaims.** `src/skills.rs:587-591` says an install interrupted halfway leaves a directory the next one still recognises as its own. Moving the record to the front shrinks that window a great deal but does not close it: an interrupt between `create_dir_all` (`:584`) and the record write (`:593`) leaves a recordless empty directory `wf` itself made, which `Mirror::inspect` then reports as `Adopted` indefinitely. The error message's recovery path ("move it aside") covers it, so this is cosmetic — the comment, not the code.

3. **Nit — TOCTOU between `Mirror::inspect` and `create_dir_all`** (`src/skills.rs:581-586`): a symlink planted in that window is followed by `create_dir_all`, and the record and copies land in the linked tree — precisely the shape `inspect` exists to refuse. Single-user config directory, adversarial-only; a mkdir-then-verify would close it if it is ever worth the lines.

4. **Nit — three meanings of "mirror".** `MIRROR` (the name), `Target::mirror` (the path) and the new `enum Mirror` (its provenance state) now coexist, and the public arm is `CopyUnmanaged` while the internal one is `Adopted` (`src/skills.rs:387-398`, `546-556`). Mild Mysterious Name; the doc comments carry it, but `Provenance` would read cleaner.

## Spec

Spec is #187, "The skills mirror is pruned only when wf made it". **All five stated requirements are met**, verified by tracing the paths rather than trusting the description, and the new tests were checked to be genuinely red on the old logic: patching `Mirror::inspect` to always return `Ours` in a throwaway worktree fails three of the four, and the fourth pins the deleted `parent == bundle` arm, confirmed present at the merge-base (old `src/skills.rs:720`).

- *"prune only a mirror that carries a `.source` record wf wrote"* — `prune_mirror` (`:772`) is reachable only from `install`, which returns `CopyUnmanaged` before creating or writing anything when `Mirror::inspect` (`:410-419`) finds no record (`:581-583`). Met.
- *"absent it, treat the directory as Unmanaged: report and leave it"* — reported at `src/main.rs:282-292` with a non-zero exit, and the launch-time `refresh` is gated on the same proof (`src/skills.rs:702-706`), so no launch writes through it either. Met.
- Symlinked mirror — `inspect` uses `symlink_metadata` and classifies a non-directory as `Adopted` *before* consulting the record, so a link whose target carries a record is still refused (`:414-417`); pinned at `:1476`. Met.
- *"tighten `is_ours` so a link is wf's only when wf's own records say so, not merely because it points into the bundle"* — the `parent == bundle` arm is gone (`:808-822`); pinned at `:1508`. Met.
- *"Tests for both adoption shapes"* — `:1442` and `:1476`, plus two `is_ours` tests at `:1508` and `:1531`. Met.

Non-blocking:

1. **The same "Nothing was touched" wording**, found independently on this axis against *"report and leave it"*: the mirror itself is indeed left exactly as found — only the sentence and the README overstate, because sweep already ran.
2. **`is_ours` keeps two shape-based arms with no record behind them** (`:816-822`): the `share/wf/skills` suffix, and a relative `../wf-skills` link while the mirror is *Absent*. In the Absent case no record exists to consult, so *"only when wf's own records say so"* is met in spirit — nothing else plausibly leaves either shape — but not to the letter: a hand-made `foo -> ../wf-skills/foo` with no `wf-skills` present would be swept. Deliberate, pinned at `:1548-1552`, documented in the README, and it is the residue #104's rename actually left. Flagged for the record, not for change.

Edge cases probed and clean: record present but mirror is a symlink (Adopted), empty `.source` (Adopted — conservative), unreadable mirror (stat failure → Absent → `create_dir_all` errors before any prune).

## Verdict

**Comment** — no blocking findings. The defect the ticket names is fixed at the seam it names, both adoption shapes are refused with tests that are red on the old logic, and the accepted cost (a pre-#110 link into a *checkout* bundle no longer reclaimed or swept) is stated in the PR, the code comments and the README rather than left for a reader to discover.

Nothing here blocks a merge. One finding is worth fixing first because it is the same class of over-claim this PR exists to remove: `wf skills install` prints "Nothing was touched" and the README says "having written nothing anywhere", while `sweep` ran before `install` and may have printed link removals in the lines below. Two sentences, or an earlier mirror check. The remaining three Standards items are nits and the second Spec item is deliberate and documented.

Issue #184, opened 2026-08-22
The screen says when a map fetch was truncated


blooop:
## Question

Surface truncation in the map fetch instead of silently dropping data, and stop a label-heavy map from being un-mapped.

`fetch.rs:54–61` caps `subIssues(first: 100)`, `blockedBy(first: 50)`, and `labels(first: 20)` with no `pageInfo` and no indicator. Consequences: a 101-ticket map silently drops a ticket; 51 blockers can misclassify a node as unblocked; and — worst — a map issue with 21+ labels where `wayfinder:map` is not in the first 20 fails the re-verification at `fetch.rs:333` and is rejected as "no longer a map", while discovery keeps re-finding it, permanently.

Fix shape: query the map label membership directly (filter by label in the query, or raise the cap well past plausible) so label count can never un-map a map; fetch `pageInfo.hasNextPage` on the ticket-bearing connections and say so on the count line when a map arrived truncated, rather than rendering a silently partial tree.

Blocked by the reap-rollup ticket because `reap.rs:2694–2713` byte-compares the shared PR selection block — land the reap side first, then keep that drift test green here.

blooop:
**breadcrumb:** claimed by wf-mid run (2026-08-22); entering build stage via wf-tdd in a fresh worktree subagent. **agent-decided (simplicity):** #183 is at PR #199, approved-but-unmerged, and this ticket's drift test depends on its query text — so this build branches from `wayfinder/wayfinder-183` and its PR stacks on #199 (base = that branch) rather than duplicating the query edit off main. Close call: waiting for #199 to merge nearly won, but merging stays human and parking a buildable ticket on a human's schedule loses to a stacked PR that is rebased in one click.

blooop:
[...]

PR #199, opened 2026-08-22
Reap survives a truncated PR rollup

### LINKED ISSUE #183: Reap survives a truncated PR rollup
## Question

Make `wf reap` survive a truncated PR rollup instead of deleting on it.

`closedByPullRequestsReferences(first: 5, includeClosedPrs: true)` at `reap.rs:1156` fetches no `pageInfo` and never paginates. GitHub returns references oldest-first, so on a node with more than five linked PRs the *open* PR is exactly the one that falls off the page: the node reads as done-by-merge and `plan()` reaps its workspace while work is in flight. This silently violates the module's own "never partial" standard (`reap.rs:1099–1102`), and no test covers >5 PRs.

The fix the review recommends: ask for `pageInfo { hasNextPage }` and treat a truncated rollup as unreadable — the same keeping arm an unknown PR state already lands on — rather than paginating (reap's posture everywhere else is fail-toward-keeping, and a sixth PR is rare enough that a kept row saying why beats a second round trip). Add the >5-PR fixture test.

Beware: `reap.rs` byte-compares its PR selection block against `fetch.rs`'s (`reap.rs:2694–2713`), so this query edit moves the screen's query too — which is why the screen-truncation ticket is blocked on this one.

blooop:
Closes #183

## The hazard

`closedByPullRequestsReferences(first: 5, includeClosedPrs: true)` was asked for
a bounded page and never asked whether that page was all of it. GitHub answers
the connection **oldest-first**, so the reference a node with more than five
linked PRs loses is the *newest* — which on a `wf` node is the PR still open.

Read from that page alone, such a node is five merges and nothing in flight:
`NodeFact::DoneByMerge`, `Verdict::Reap`, and `wf reap` deletes a workspace
whose work is in flight. That is the one direction `reap` may not be wrong in,
and it violated the module's own "never partial" standard silently — no fixture
in the file had more than two linked PRs, so nothing was red.

## The fix

Per the ticket's recommendation: **ask, and treat a truncated rollup as
unreadable rather than paginating.**

- Both query texts select `pageInfo { hasNextPage }` — reap's copy and the map
  query's, because `both_queries_ask_for_a_linked_pr_in_exactly_the_same_words`
  holds them byte-equal. Nothing on the screen reads the answer yet; #184 is
  where that lands, and `fetch.rs` carries a comment saying why the field is
  there in the meantime.
- The page and its completeness travel as **one value** (`PrRollup::Whole` /
  `PrRollup::Truncated`), so no caller can hold a page without holding what it
  is a page of, and every one of the sixteen existing `node_fact` call sites now
  says which it means.
- A page that was not all of it reads as a seventh `NodeFact` arm,
  `RollupTruncated`: `plan` keeps it with a row saying why, `doomed` never names
  it, and `unfinished` stops the unattended cleanup on it.
- The remainder is **not** fetched. `reap` fails toward keeping everywhere else,
  and a sixth linked PR on one node is rare enough that a kept row costs a human
  one glance where a second round trip costs every run latency it almost never
  needs.

Two answers the page cannot overturn are deliberately unchanged, each with its
own test: a **closed ticket** is the ticket's own word, read before any PR is;
and a **live PR visible on the page** already dominates every arm below it, so a
reference behind the page can only agree with it. Saying "cannot tell" over an
open PR the binary can point at would be a worse row, not a safer one.

## Tests

Six new, and two mutations confirmed to bite:

- Removing the truncation arm from `node_fact` reddens four tests, including the
  wire-to-`doomed` one.
- Removing `pageInfo` from the reap selection reddens three, including
  `the_batch_parse_reads_no_field_the_batch_forgot_to_ask_for` — the shared
  fixture gained a >5-PR node (`i87`) whose fact is *derived from the selection*,
  so the field going missing from the query changes the answer rather than going
  unnoticed.

`a_full_rollup_page_keeps_the_workspace_all_the_way_from_the_wire` is the
ticket's requested >5-PR test, and it runs from the bytes `gh` would print to
`doomed`. Its two bodies differ in `hasNextPage` and nothing else, so one
reaping and the other keeping can only be the tracker's word about the page
being read.

The `Kept, always:` list in the README gains the new keep — a ticket with five
merged PRs is *described there* as finished work, so `wf` keeping it anyway
looks like a bug until the list says the page is the reason.

## Checks

The full offline chain from AGENTS.md is green under `RUSTFLAGS=-D warnings` and
`RUSTDOCFLAGS=-D warnings`, including the source-text guards over the reap path
and the byte-compare between the two PR selections. The new query text was also
run against live GitHub (`gh api graphql`) and the map query driven through
`examples/preview_screen` against a real map, so `pageInfo` is a field both
connections really answer.

## Summary by Sourcery

Keep workspaces when the linked pull-request rollup is incomplete instead of treating partial merge history as evidence that work is finished.

Bug Fixes:
- Prevent workspace reaping when GitHub returns a truncated linked-pull-request rollup.
- Treat incomplete PR evidence as a keep condition while preserving closed-ticket and visible in-flight PR decisions.

Enhancements:
- Track PR rollup completeness alongside its page and add an explicit truncated-rollup node fact.
- Request linked-PR pagination metadata in both GraphQL selections without fetching additional pages.

Documentation:
- Document that tickets with more linked PRs than wf reads are always kept.

Tests:
- Add coverage from GraphQL response parsing through reaping decisions for complete and truncated rollups.
- Expand exhaustive node-fact and query-selection tests to cover the new safety behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1161,12 +1161,23 @@ decides. Needs `dl` **0.0.21 or newer** for the JSON listing.
 Kept, always: workspaces `dl` did not create (they are not `wf`'s, whatever
 their branch looks like), branches that are not `wayfinder/<repo>-<n>` for that
 repo, tickets with an open or draft PR (in review is where review fixes happen),
-tickets someone has claimed that no PR has come of, and anything **running** — a
-ticket closing is no evidence that the session in the container ended.
+tickets someone has claimed that no PR has come of, tickets with more linked PRs
+than `wf` read, and anything **running** — a ticket closing is no evidence that
+the session in the container ended.
 
 The claim is the *last* thing read, not the first: it settles a ticket nothing
 has come of yet, and it does not outrank the PRs. A claimed ticket whose PR
 merged is finished work, and reap collects it like any other.
+
+**A ticket with more linked PRs than `wf` read is kept saying so.** The linked-PR
+rollup is asked for one bounded page and GitHub answers it oldest-first, so on a
+ticket with more references than the page holds the one that falls off is the
+newest — which on a `wf` ticket is the PR still open. Read from that page alone
+such a ticket is a row of merges and nothing in flight, which is exactly the
+reading that reaps. So `wf` asks whether the page was all of it, and a page that
+was not is not evidence at all: the row says there is more of this than it read
+and the workspace stays. It does not fetch the rest — a sixth PR on one ticket is
+rare, and a kept row costs a glance where a second round trip costs every run.
 
 **A `dl` that says nothing is not a `dl` saying nothing is at risk.** From
 devlaunch **0.0.24** the listing answers `unsaved` for every clone `dl` made,
--- a/src/fetch.rs
+++ b/src/fetch.rs
@@ -44,6 +44,13 @@ pub const MAP_LABEL: &str = "wayfinder:map";
 /// `pub(crate)` for one reason: [`reap`](crate::reap) selects a subset of these
 /// same fields into its own batched query, and the two are held to each other
 /// by a test rather than by a comment. Nothing outside that test may read it.
+///
+/// That tie is why the linked-PR rollup asks for `pageInfo { hasNextPage }`
+/// while nothing on the screen reads the answer yet: the reaper needs it (#183
+/// — a rollup it cannot see all of must not read as done-by-merge), the two
+/// selections are held byte-equal, so it lands here in the same breath. The
+/// screen's own use of it is its own change; an unread nullable field costs
+/// this query nothing and a divergent copy would cost it the guard.
 pub(crate) const MAP_QUERY: &str = "\
 query($owner: String!, $name: String!, $number: Int!) {
   repository(owner: $owner, name: $name) {
@@ -64,6 +71,7 @@ query($owner: String!, $name: String!, $number: Int!) {
               statusCheckRollup { state }
               repository { nameWithOwner }
             }
+            pageInfo { hasNextPage }
           }
         }
       }
--- a/src/liveness.rs
+++ b/src/liveness.rs
@@ -324,6 +324,11 @@ mod tests {
             NodeFact::Superseded { pr: 1 },
             NodeFact::InFlight { pr: 1 },
             NodeFact::Unstarted,
+            // Nor is a rollup `wf` did not see all of (#183): a stall is a
+            // claim with nothing come of it, and this is the tracker declining
+            // to say what came of it — a reason to keep the workspace, not a
+            // reason to tell anyone the run died.
+            NodeFact::RollupTruncated,
         ] {
             let known = facts([(node(REPO, 133), fact.clone())]);
             assert_eq!(
--- a/src/reap.rs
+++ b/src/reap.rs
@@ -29,6 +29,13 @@
 //! is why [`Verdict::Warn`] exists and why [`doomed`] is the single place that
 //! decides what actually goes.
 //!
+//! One reading is not on that lattice at all, and it is the reason [`PrRollup`]
+//! exists: the linked-PR page the tracker sends is bounded, and a node with
+//! more references than it holds loses the newest of them — the PR still open
+//! (#183). So the *completeness* of that page travels with it, and a page that
+//! was not all of it is not evidence about the node in either direction. It
+//! keeps, and the row says why.
+//!
 //! Two commands come out of that one decision, and they differ only in who is
 //! reading. [`run`] is `wf reap`: every workspace on the machine, a plan
 //! printed, a prompt. [`cleanup`] is what an autonomous run ends with (#151):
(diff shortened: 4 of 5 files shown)


blooop (commented):

Fresh-context review of #199 against ticket #183. Fixed point: merge-base `4135446`...`9c0fdf8` (`git diff 4135446...9c0fdf8`). Preflight: ref resolves, diff non-empty (5 files, +471/-39), all checks green on the head. Two axes were run as independent subagents and are reported unmerged.

## Standards

Repo-documented standards (AGENTS.md, module docs) come first here, and the diff follows them closely.

**House rules — respected.**

- The reap/fetch query byte-compare guard (`both_queries_ask_for_a_linked_pr_in_exactly_the_same_words`, `src/reap.rs:3034`) is honored: `pageInfo { hasNextPage }` was added to `fetch::MAP_QUERY` (`src/fetch.rs:74`) and `NODE_FACT_SELECTION` (`src/reap.rs:1246`) in identical words, and `src/fetch.rs:47-53` documents why the screen carries a field it does not yet read.
- The selection-prune guard was extended so node `i87`'s fact actually *depends* on the new field (`src/reap.rs:3226-3232`) — the query-text test is not the only thing holding it, which is the difference between a guard and a comment.
- "Illegal states unrepresentable — do not add a bool that can disagree with a sum type" is followed exactly. The wire bool (`PageInfo.has_next_page`, `src/reap.rs:1305-1310`) is confined to the parse boundary and converted immediately into `PrRollup::Whole`/`Truncated` (`src/reap.rs:1342-1347`), so no caller can hold a page without holding what it is a page of. `node_fact` taking `PrRollup<'_>` rather than `&[PrOutcome]` plus a flag is the right constructive move.
- Why-comments carry the decisions and the rejected alternatives (`Keep` not `Warn` at `src/reap.rs:764-770`; `false` as the safe serde default at `src/reap.rs:1299-1305`; the remainder deliberately not paginated at `src/reap.rs:461-465`).

**Fowler baseline.** No Mysterious Names (`RollupTruncated`, `PrRollup`, `rollup_all_of_it` all self-explain). The new `NodeFact` arm ripples through five match sites (`src/reap.rs:742`, `:951`, `:3686`; `src/reclaim.rs:398`; liveness via the test at `src/liveness.rs:322-331`), but each is an exhaustive match the repo maintains deliberately as a forcing function — that is the designed mechanism, not Shotgun Surgery. An unread `pageInfo` on the screen query would normally read as Speculative Generality; the byte-equality guard makes divergence the more expensive option and the doc comment says so, so it is justified. `page()` is a two-line accessor, not a Middle Man. No data clumps, message chains, feature envy, or refused bequest.

**Finding**

- *nit* — `src/reap.rs:775`: the Keep reason ends `— keeping`, but the row renderer already prefixes every Keep row with the `keep` label (`src/reap.rs:1475`), so it prints as `keep  <id>  (… — keeping)`. No other Keep reason carries the suffix (`"still running — stop it first"`, `"{name} is still open"`). The string is also asserted verbatim at `src/reap.rs:2860`, so a future rewording touches two places.

Axis verdict: **pass** — the change follows the repo's guard-test and constructive-modeling rules to the letter; the single finding is cosmetic.

## Spec

Spec is ticket #183. Every finding below quotes it. All 436 lib tests pass on the head.

1. **"ask for `pageInfo { hasNextPage }`" — satisfied.** Present in `NODE_FACT_SELECTION` (`src/reap.rs:1246`) and `MAP_QUERY` (`src/fetch.rs:74`), with `the_batch_asks_whether_the_pr_rollup_was_all_of_it` asserting the query text and the prune test proving the field is load-bearing for node `i87`'s fact.

2. **"treat a truncated rollup as unreadable … rather than paginating" — satisfied.** `first: 5` is unchanged and no second request exists. `parse_node_facts` builds `PrRollup::Truncated`, `node_fact` returns `NodeFact::RollupTruncated`, and `plan()` maps it to `Verdict::Keep` with a row saying why — matching "a kept row saying why beats a second round trip". *Non-blocking note:* the spec's letter says "the same keeping arm an unknown PR state already lands on" (i.e. `InFlight`); the PR adds a distinct arm instead. Behaviourally identical (keep in `plan`, stop in `cleanup`), and the distinct arm delivers the spec's own "kept row saying why" better — a faithful improvement, not a deviation. The check is ordered after ticket-`Closed` and after an on-page live PR, both of which truncation genuinely cannot overturn; no ordering reaches a delete-on-truncation.

3. **"Add the >5-PR fixture test." — satisfied.** Fixture `i87` carries five merged PRs plus `"pageInfo": {"hasNextPage": true}`, the only faithful wire shape for >5 references over a `first: 5` page. `a_full_rollup_page_keeps_the_workspace_all_the_way_from_the_wire` runs wire bytes → `parse_node_facts` → `plan` → `doomed`, asserts `doomed` is empty (both with and without `-f`) and asserts the keep reason, paired against the same body with `hasNextPage: false` which still reaps — isolating the one word.

4. **"this query edit moves the screen's query too" — satisfied.** The byte-equality guard (`src/reap.rs:3049-3050`) passes with both queries carrying the new field.

5. **"This silently violates the module's own 'never partial' standard" — no consumer gap found.** Facts are derived in exactly one place (`node_fact`, sole caller `parse_node_facts`), so no site can still read a truncated page as done-by-merge. `plan` keeps; `doomed` collects only `Reap`; the autonomous `cleanup` stops via `unfinished` (fate-table test asserts `Stops`); liveness matches only `Claimed`, so truncation is not reported as a stall (test added); `reclaim` goes through `plan`, so `Keep` never surfaces, and its exhaustive-arm test now covers the seventh arm. `fetch.rs` deliberately does not read the field yet — the spec itself defers that ("the screen-truncation ticket is blocked on this one").

Axis verdict: **pass** — every requirement is implemented and tested end-to-end.

## Verdict

**Approve.** No blocking findings on either axis. The one non-blocking item is a cosmetic redundancy in a Keep row string (`src/reap.rs:775`), safe to fold into a later touch of that file or to leave alone.

(Posted as a comment rather than an approval: GitHub refuses `--approve` on a PR authored by the same account, which is the normal case for an agent reviewing an agent's work. The written verdict above is the gate.)

PR #200, opened 2026-08-22
fix: reap re-checks running right before it removes

### LINKED ISSUE #186: Reap re-checks running right before it removes
## Question

Close the TOCTOU on interactive reap's "still running" guard.

"Running" is read from the `dl --ls --json` snapshot (`reap.rs:659`), then `run()` prints the plan and blocks on the y/N prompt for arbitrarily long (`reap.rs:1361–1372`) before calling `remove()`. A session started between the listing and the confirmation gets its container torn down: `dl` re-checks *unsaved work* at `rm` time (that refusal is load-bearing, `reap.rs:1262–1266`) but nothing re-checks *running* — a live, fully-pushed session is destroyable. The autonomous `--finished` path narrows the window but has the same shape.

Fix shape: re-take the listing (or at minimum the running bit for the doomed rows) immediately after the prompt answers and before the first removal; a row that became running between plan and confirmation is kept and named, like any other keeping arm. Mind the guard tests: `reap.rs` and `main.rs` carry source-text denylists around this path.

blooop:
Closes #186 (part of map #182).

## What

Both deleting paths re-take the `dl --ls --json` listing immediately before the first removal, and re-apply the running guard to the doomed set:

- **`wf reap`**: after the y/N prompt answers (the window the ticket names — the prompt blocks for arbitrarily long, and the plan's "still running" guard read a listing from before it). A doomed row whose container came up in the window is spared and named as a keep row (`started running since the plan — stop it first`), and the run ends green.
- **`wf reap --finished`**: after the decision, before the first removal — the same window one size smaller (tracker round-trip plus the decision). The spared row joins the `kept` report as an ordinary keep, consistent with `decide`'s existing treatment of a running workspace of a finished ticket.

A listing that cannot be re-taken fails the whole command: `wf` gone blind between plan and removal must not delete on the reading it had while it could still see.

Only the running bit is re-read, deliberately — the tracker facts are not re-taken. The plan the human approved is the plan that runs; the container check protects the one thing on this path that cannot be recovered after the fact. A doomed row absent from the fresh listing still goes: "it does not exist" is `dl`'s answer to give at `rm` time.

## Tests (written first, red before green)

- Unit, at the partition seam (`recheck_running`): a doomed row now running is spared and named; one still down goes; one gone from the fresh listing goes to `dl`.
- End to end, via recorded probe runs: `probe::record_relisted` gives the shimmed `dl` a second listing served from its second fixture-serving call onward, so a run can watch the machine move between two reads. Both `wf reap -y` and the autonomous cleanup are pinned to remove nothing, name the kept row, and end green when the doomed workspace turned Running between reads — both tests failed before the wiring with `dl <wf-129-closed> <rm>` in the recording.

Guard tests: `main.rs` untouched; the reap/fetch query byte-compare untouched; full local chain green (`fmt`, `clippy`, lib/bins/examples, `skill_docs`, `devcontainer_prebuild`, `toolchain_pin`, `offline_green`, `live_fetch common`, `doc`) with `-D warnings`.

## Summary by Sourcery

Recheck workspace running states immediately before removal to protect sessions that start after reap planning.

Bug Fixes:
- Prevent `wf reap` and finished-workspace cleanup from deleting containers that started running after the initial plan was read.
- Fail the command when the workspace listing cannot be refreshed before removals.

Enhancements:
- Share running-state revalidation across interactive and autonomous deletion paths while preserving the approved tracker plan.
- Report newly running workspaces as kept items and continue successfully without removing them.

Tests:
- Add unit coverage for running, stopped, and missing workspaces during revalidation.
- Add recorded end-to-end coverage for both interactive reap and autonomous cleanup when a workspace starts running between listings.

### DIFF
--- a/src/probe.rs
+++ b/src/probe.rs
@@ -218,7 +218,24 @@ impl Recording {
 /// If the child could not be run, or failed. A failed child is a failed probe:
 /// its assertions are the test's.
 pub fn record(test: &str, dl_stdout: &str, gh_stdout: &str) -> Recording {
-    record_as_dl(test, dl_stdout, gh_stdout, SHIMMED_DL)
+    record_full(test, dl_stdout, None, gh_stdout, SHIMMED_DL)
+}
+
+/// [`record`], with a *second* listing the shimmed `dl` answers from its
+/// second fixture-serving call onward.
+///
+/// This exists for exactly one kind of claim: what the code under test does
+/// when the machine **moves between two reads** — the TOCTOU that #186 closes,
+/// where a container starts between the listing a plan was made from and the
+/// removal that plan authorises. A single-fixture shim cannot produce that
+/// world: it answers every call identically, so a re-take of the listing is
+/// indistinguishable from never re-taking it.
+///
+/// `--version` answers do not consume the first fixture — the shim answers
+/// those before it reaches the fixture at all — so `dl_then` arrives on the
+/// second call that actually asked for output, whichever argv that is.
+pub fn record_relisted(test: &str, dl_first: &str, dl_then: &str, gh_stdout: &str) -> Recording {
+    record_full(test, dl_first, Some(dl_then), gh_stdout, SHIMMED_DL)
 }
 
 /// [`record`], with the release the shimmed `dl` claims to be.
@@ -238,8 +255,21 @@ pub fn record(test: &str, dl_stdout: &str, gh_stdout: &str) -> Recording {
 ///
 /// [`DEVLAUNCH_FLOOR`]: crate::launch
 pub fn record_as_dl(test: &str, dl_stdout: &str, gh_stdout: &str, dl_version: &str) -> Recording {
+    record_full(test, dl_stdout, None, gh_stdout, dl_version)
+}
+
+fn record_full(
+    test: &str,
+    dl_stdout: &str,
+    dl_then: Option<&str>,
+    gh_stdout: &str,
+    dl_version: &str,
+) -> Recording {
     let dir = scratch(test);
     std::fs::write(dir.join("dl.out"), dl_stdout).expect("the dl fixture");
+    if let Some(then) = dl_then {
+        std::fs::write(dir.join("dl.out.again"), then).expect("the second dl fixture");
+    }
     std::fs::write(dir.join("gh.out"), gh_stdout).expect("the gh fixture");
     let log = dir.join("argv.log");
     std::fs::write(&log, "").expect("the log");
@@ -647,6 +677,14 @@ fn shim(dir: &Path, name: &str, dl_version: &str) {
     //
     // The answer is [`SHIMMED_DL`] for the same reason the fixtures name a
     // release: a probe is only evidence about a machine it could be.
+    // The `.again` block is what lets a probe watch the machine *move between
+    // two reads* (see [`record_relisted`]): when a second fixture exists, the
+    // first fixture-serving call leaves a `.seen` marker beside it and every
+    // later one answers from `.again` instead. The marker lives in the scratch
+    // dir, not the child's `HOME`, so `touched_no_files` does not read the
+    // shim's own bookkeeping as the run disturbing the machine. With no
+    // `.again` file the branch is never entered and the shim answers every
+    // call identically, as it always has.
     let body = r#"#!/bin/sh
 line=$({
   printf '%s' 'PROGRAM'
@@ -654,6 +692,10 @@ line=$({
 } | tr '\n' ' ')
 printf '%s\n' "$line" >> "$LOGVAR"
 if [ "$1" = "--version" ]; then printf '%s\n' 'PROGRAM VERSION'; exit 0; fi
+if [ -e 'FIXTURE.again' ]; then
+  if [ -e 'FIXTURE.seen' ]; then cat 'FIXTURE.again'; exit 0; fi
+  : > 'FIXTURE.seen'
+fi
 cat 'FIXTURE'
 "#
     .replace("PROGRAM", name)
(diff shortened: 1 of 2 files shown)


blooop (commented):

Reviewed with fresh eyes at `b766769` against merge-base `4135446` (`git diff 4135446...b766769` — `src/probe.rs`, `src/reap.rs`, +255/-1). Preflight: ref resolves, diff non-empty, all five checks green. Two axes run as independent parallel reviewers; findings deliberately not merged or reranked across them.

## Standards

Nothing blocking. The change follows the house style closely: the *why* sits next to the code, the guard is one function used by both deleting paths, and the new keep row is reported in each path's own format.

Refuted before reporting, so they are not re-raised: the generic on `recheck_running` is not Speculative Generality (two real callers, genuinely different element types — `&Verdict` and `Cleared`); the wording split between `"started running since the plan"` and `plan`'s `"still running — stop it first"` (`src/reap.rs:709`) is deliberate and carries the window; inline `replacen` fixture editing is the established pattern; `self::workspaces()` is forced by the shadowing local binding; and no new probe children, so AGENTS.md's probe-count arithmetic still holds.

Nits, all optional:

1. **Duplicated Code** — `src/reap.rs:3662-3671` and `:3845-3854` are ten byte-identical lines building the same `DL_LISTING` variant. `src/probe.rs`'s own `DL_LISTING` doc states the rule ("a fixture written twice is a fixture that can disagree with itself about what the machine looks like"), and `DL_LISTING_UNSAVED` / `DL_LISTING_LEGACY` are the house answer: a named variant in `probe.rs`, derived once.
2. **Mysterious Name / missing doc** — `src/probe.rs:261` `record_full` is the only member of the `record*` family with no doc comment, and "full" names its arity rather than its job. Every other private fn in that file is documented.
3. **Documentation drift** — `README.md:1116-1126` shows the `wf reap` plan and prompt with no keep row printed *after* the y/N, and `run`'s own doc comment still describes the plan as the place what-is-kept is said. `src/reap.rs:2908` (`the_readme_states_the_keep_rule_the_derivation_implements`) is this repo's own record that AGENTS.md makes the README part of `wf`'s interface. The *rule* is already documented ("Kept, always: … anything **running**", `README.md:1164`) — it is only the new, post-prompt *timing* that is undocumented, which is why this is a nit and not a Spec finding.
4. **Minor duplication in `run`** — `src/reap.rs:1433` re-spells the plan's row format inline where `cleanup` reuses its existing `kept` loop, and `"nothing to reap"` now has two emit sites in one function.
5. **Untested claim** — `src/reap.rs:1424-1429` asserts that a listing which cannot be re-taken fails the whole command. The behaviour is real (the Spec axis reproduced it), but no test guards it, unlike the neighbouring safety claims. Also `src/reap.rs:1519` takes the fresh listing unconditionally, so an empty `going` still costs a `dl --ls --json`; `run` guards that case.

## Spec

**The spec is met**, on both paths, verified by call order in the source rather than from the PR description.

- *"re-take the listing (or at minimum the running bit for the doomed rows) immediately after the prompt answers and before the first removal"* — interactive: the y/N read ends at `src/reap.rs:1411-1422`, the re-listing is `:1430-1431`, the first `remove()` is `:1445-1446`. Nothing deletes in between.
- *"The autonomous `--finished` path narrows the window but has the same shape."* — `decide()` at `:1503`, re-listing at `:1519-1520`, first `remove()` at `:1536-1537`.
- Only two `remove()` call sites exist in the crate (`:1446`, `:1537`) and both sit behind the guard; `main.rs` and `picker.rs` reach deletion only through `reap::run` / `reap::cleanup`.
- *"a row that became running between plan and confirmation is kept and named, like any other keeping arm"* — `recheck_running` (`src/reap.rs:427`) emits `Verdict::Keep { reason: "started running since the plan — stop it first" }`; `run` prints it in the plan's own `keep` format (`:1432-1434`), `cleanup` folds it into `kept` ahead of the removal report (`:1521`). Both end green.
- Fail-safe on a blind re-listing, checked empirically rather than taken from the PR body: with an unparseable second listing, both paths abort with zero `dl … rm` in the argv log.
- The tests are not vacuous: with both guard blocks deleted in a scratch worktree, `a_workspace_that_started_running_while_the_prompt_waited_is_spared` and `a_finished_workspace_that_started_running_before_collection_is_kept` both fail with `this path must not be able to destroy a workspace, and it ran: dl <wf-129-closed> <rm>`. Restored; `cargo test --lib` is 435/435 at `b766769`, so the source-text denylists the ticket warns about are intact.

**Residual, narrowing rather than a violation.** `recheck_running` spares only `is_running()`, i.e. `state == "Running"` (`src/reap.rs:121`). A container *mid-start* reads `Busy`, which this file's own doc calls "a container mid-start or mid-stop" (`:135`) and which answers `false` to both `is_running` and `is_down` — so a `dl up` still in `Busy` at re-check time is doomed, and a container that came up *during the window* is exactly the population most likely to be mid-start. This mirrors the pre-existing plan guard (`:709`) and sits within the ticket's own floor ("at minimum the running bit"), so it is inherited, not introduced. Likewise the re-check runs once before the loop, so a session starting mid-loop is unguarded. Both are seconds-wide against the unbounded prompt the ticket names — worth a line on map #182 rather than a change here.

## Verdict

**Comment — no blocking findings on either axis. Mergeable as it stands; the node may advance.**

Blocking findings: none. Standards nits 1–5 are optional polish and Spec's residual is pre-existing; neither axis asks for a change before merge. The two items worth carrying forward as notes on map #182, not as work on this PR: the `Busy` mid-start state falling outside both running predicates, and the post-prompt keep row's absence from the README's `wf reap` walkthrough.

Issue #190, opened 2026-08-22
A red live or scheduled run files an issue


blooop:
## Question

Make a red non-blocking run summon somebody: alert on failures of `live.yml` and the weekly devlaunch-contract re-solve.

The live tier is deliberately non-blocking (main-push only; two tests assert the tracker's present contents, one asserts wall-clock budgets), and the weekly contract re-solve is the only leg that can discover a new devlaunch release. Both depend entirely on a human noticing a red run — for a non-blocking tier, noticing is the whole game, and nothing files an issue or notifies today.

Fix shape: a small `workflow_run`-triggered (or job-level `if: failure()`) step that files/updates a single tracking issue on this repo when `live.yml` or the scheduled `devlaunch-contract.yml` run goes red — idempotent (one open issue per workflow, updated not duplicated), closing itself or commenting on the next green. Keep permissions job-scoped (`issues: write` on that job only), matching the repo's minimal-permissions posture.

blooop:
**breadcrumb:** claimed by wf-mid run (2026-08-22); entering build stage via wf-tdd in a fresh worktree subagent.

blooop:
**breadcrumb:** building via wf-tdd. Restating the contract: a new non-blocking alert workflow, triggered by `workflow_run: completed` of the live leg and the devlaunch contract leg, that reconciles one tracking issue per watched workflow — files it on a red run, comments instead of duplicating while it stays red, closes it on the next green of the same class, and does nothing for a cancelled run. It reacts only to the runs that are non-blocking by design: live runs on main, and the contract workflow's scheduled re-solve leg. `issues: write` is granted to that one job only.

Seams, pre-agreed by the ticket's fix shape and the repo's files-as-behavior pattern:
1. The alert workflow's configuration text (trigger type, the watched workflow names derived from the two source workflows' own `name:` fields so a rename cannot silently disconnect the subscription, the permission grants, the non-blocking-leg guards) — asserted by a new offline test binary in the style of the devcontainer prebuild contract.
2. The reconciliation logic itself, as a script the workflow invokes, tested by running it against a recording `gh` stub — the same shim ethos the devlaunch contract tests use — asserting create-vs-comment-vs-close-vs-nothing at the gh boundary.

blooop:
[...]

PR #201, opened 2026-08-22
Counts and cursor agree on a DAG diamond

### LINKED ISSUE #188: Counts and cursor agree on a DAG diamond
## Question

Make the count line and the cursor agree with the tickets, not the rows, when the DAG has a diamond.

The leverage lens deliberately draws one ticket under *every* root that unblocks it. Two consumers don't compensate:

- `counts()` at `app.rs:587` uses `visible().len()` (drawn rows) as the numerator and `scoped().len()` (tickets) as the denominator — a 4-ticket map with a diamond renders 5 rows and the count line reads `5/4`; a sifted query matching 2 tickets claims 3. A test currently pins the wrong behavior: `ui.rs:1986` asserts `3/5`.
- Cursor identity-pinning takes the *first* stop matching a `StopKey` (`point_at`, `app.rs:697`; same in `preserve_cursor`, `refresh.rs:438`), so a cursor placed on the second rendering of a duplicated ticket teleports to the first after a `replace_clusters` or lens toggle — defeating the "still the row they chose" guarantee those functions exist for.

Fix shape: count distinct tickets in the numerator (the rollups already count nodes-not-rows — reuse that discipline); disambiguate duplicate stops (e.g. an occurrence index in the key) so pinning is stable. Fix the `ui.rs:1986` pin to assert the corrected count. The deeper redesign of row identity is deliberately out of this ticket — see the map's Not yet specified.

blooop:
Closes #188.

The leverage lens deliberately draws one ticket under every root that unblocks it. Two consumers didn't compensate; both now count and pin **tickets, not drawn rows**.

## Count line

`counts()` used `visible().len()` — drawn rows — as the numerator over a ticket denominator, so a diamond inflated "shown": a sifted query matching 2 tickets claimed 3, and the unsifted fixture read a coincidentally-round 5/5 off five rows of four distinct shown tickets. The numerator now dedups rows by `(map, index)` — the same nodes-not-rows discipline the rollups keep. The rendered pins move with it: the sifted `3/5` becomes `2/5`, the unsifted `5/5`s become `4/5`.

## Cursor identity

`StopKey`'s ticket arm gains an occurrence index — which drawing of the ticket the stop is — computed over the whole stop list, since a key derived from one stop alone cannot know it is the second drawing. `point_at` and the swap-time resolve match the exact drawing first, and degrade to the first drawing of the same ticket when that drawing left the screen (a lens toggle to forest, a root that closed) — so a cursor parked on the second rendering no longer teleports to the first on every `replace_clusters`, and losing a drawing never means losing the ticket. Only a ticket gone entirely falls back to position, as before.

The deeper redesign of row identity stays out per the map's *Not yet specified*.

## Tests

- `a_ticket_a_diamond_draws_twice_is_counted_once` — red first against the row-count numerator
- `a_cursor_on_the_second_drawing_of_a_diamond_ticket_survives_a_swap` — red first against the occurrence-blind key
- `a_drawing_that_vanishes_degrades_to_the_ticket_not_the_top`, `a_lens_toggle_keeps_the_cursor_on_the_ticket_when_its_drawing_leaves` — guard the degrade path
- existing count-line pins corrected to the ticket-true numbers

Full local chain green (fmt, clippy, lib/bins/examples, skill_docs, devcontainer_prebuild, toolchain_pin, offline_green, live_fetch common, doc) with `-D warnings`.

## Summary by Sourcery

Make ticket counts and cursor preservation consistent when the leverage lens renders the same ticket under multiple DAG roots.

Bug Fixes:
- Correct ticket count displays so duplicate DAG-diamond drawings are counted once.
- Preserve cursor identity for duplicate ticket drawings across refreshes, while falling back to another drawing of the same ticket when needed.

Enhancements:
- Align count and cursor behavior around ticket identity rather than rendered row identity.

Tests:
- Add coverage for deduplicated diamond counts and cursor preservation and fallback behavior across swaps and lens changes.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -3551,6 +3551,7 @@ mod tests {
                     ticket(PROJECT, 9, "Main screen design", true, true, vec![]),
                     ticket(PROJECT, 14, "Breadcrumb markers", true, false, vec![6, 9]),
                 ],
+                truncated: false,
             },
         );
         app_on(PROJECT, clusters)
--- a/src/app.rs
+++ b/src/app.rs
@@ -168,11 +168,30 @@ pub struct RowKey {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum StopKey {
     Map(MapId),
-    Ticket(RowKey),
+    /// A ticket row, plus **which drawing of it** — the leverage lens
+    /// deliberately draws one ticket under every root that unblocks it, so on
+    /// a DAG diamond the [`RowKey`] alone names two stops. Without the
+    /// occurrence, pinning "the row they chose" resolved to the *first*
+    /// matching drawing, teleporting a cursor parked on the second one on
+    /// every refetch (#188).
+    Ticket(RowKey, usize),
     Group(GroupId),
     /// A whole project, by repo slug — already index-free, like the
     /// map and group keys.
     Project(String),
+}
+
+impl StopKey {
+    /// The first drawing of the same stop — the identity to fall back to when
+    /// the exact drawing left the screen but the ticket did not (a lens
+    /// toggle, a root that closed). Only tickets are ever drawn twice, so the
+    /// other arms are already their own first occurrence.
+    fn first_occurrence(&self) -> StopKey {
+        match self {
+            StopKey::Ticket(row, _) => StopKey::Ticket(row.clone(), 0),
+            other => other.clone(),
+        }
+    }
 }
 
 /// Where the cursor is, and — the part that matters — **whether anyone put it
@@ -233,7 +252,17 @@ impl Pinned {
     fn resolve(&self, new_order: &[StopKey]) -> usize {
         match self {
             Pinned::Identity(key, fallback) => {
-                crate::refresh::preserve_cursor(Some(key), *fallback, new_order)
+                // The exact drawing when the new order still has it; else the
+                // first drawing of the same ticket — a swap that removed the
+                // root a duplicate hung under took the drawing, not the
+                // ticket, and the cursor's promise is the ticket. Only when
+                // both are gone does the positional fallback apply.
+                let key = if new_order.contains(key) {
+                    key.clone()
+                } else {
+                    key.first_occurrence()
+                };
+                crate::refresh::preserve_cursor(Some(&key), *fallback, new_order)
             }
             Pinned::Position(fallback) => {
                 crate::refresh::preserve_cursor(None, *fallback, new_order)
@@ -614,15 +643,33 @@ impl App {
         }
     }
 
-    /// A stop's durable identity: a ticket by (map, number), a map or a group
-    /// by its own id — both of which already name no indices.
-    fn stop_key(&self, stop: &Stop) -> StopKey {
-        match stop {
-            Stop::Map(id) => StopKey::Map(id.clone()),
-            Stop::Ticket(row) => StopKey::Ticket(self.row_key(row)),
-            Stop::Group(id) => StopKey::Group(id.clone()),
-            Stop::Project(repo) => StopKey::Project(repo.clone()),
+    /// Every stop's durable identity, in on-screen order: a ticket by
+    /// (map, number) plus which drawing of it this is, a map or a group by its
+    /// own id — which already name no indices.
+    ///
+    /// Computed over the whole list rather than stop-by-stop because the
+    /// occurrence *is* list context: a key derived from one stop alone cannot
+    /// know it is the second drawing of a diamond-duplicated ticket, and a key
+    /// that cannot say so pins two stops to one identity (#188).
+    fn stop_keys(&self) -> Vec<StopKey> {
+        let mut keys: Vec<StopKey> = Vec::new();
+        for at in self.stops() {
+            let key = match &at.stop {
+                Stop::Map(id) => StopKey::Map(id.clone()),
+                Stop::Ticket(row) => {
+                    let row_key = self.row_key(row);
+                    let occurrence = keys
+                        .iter()
+                        .filter(|k| matches!(k, StopKey::Ticket(prev, _) if *prev == row_key))
+                        .count();
+                    StopKey::Ticket(row_key, occurrence)
+                }
+                Stop::Group(id) => StopKey::Group(id.clone()),
+                Stop::Project(repo) => StopKey::Project(repo.clone()),
+            };
+            keys.push(key);
         }
+        keys
     }
 
     /// Cursor position clamped into the stop list. An untouched cursor is not a
@@ -672,9 +719,10 @@ impl App {
         })
     }
 
-    /// The cursor's stable identity.
+    /// The cursor's stable identity — read off the whole key list, since which
+    /// drawing the cursor is on is a fact about the list, not the stop.
     fn cursor_key(&self) -> Option<StopKey> {
-        self.cursor_stop().map(|stop| self.stop_key(&stop))
+        self.stop_keys().into_iter().nth(self.cursor_pos())
     }
 
     /// The stop to hold on to while the list is rebuilt underneath the cursor —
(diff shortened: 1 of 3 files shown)


blooop (commented):

Reviewed in fresh context against two independent axes, neither aware of the other's findings and neither merged nor reranked. Fixed point: merge-base `4135446`, `git diff 4135446...HEAD` (`.github/scripts/red_run_alert.sh`, `.github/workflows/red-run-alert.yml`, `.github/workflows/ci.yml`, `AGENTS.md`, `tests/red_run_alert.rs`; +566/-3). Preflight: ref resolves, diff non-empty, all checks green. Both axes ran the new binary (10/10) and drove the script for real; Standards mutation-tested the production files in a throwaway worktree.

## Standards

The repo's conventions are matched with care: the files-as-behavior pattern is followed, the heavy why-comment house style is upheld, the permission split matches `package.yml`/`devcontainer.yml`, `set -euo pipefail` and quoting are correct, and the enrolment in `ci.yml` and AGENTS.md's chain honours "an assertion no workflow runs is not an assertion" — including the `Four`→`Five` count. Refuted before reporting: `head -n 1` + `pipefail` SIGPIPE (needs >64KB from `gh`, unreachable), `checkout@v7`, and the `if:` guard's shape.

1. **Blocking — the env block's *values* are unasserted, so the binary's headline claim does not hold.** `tests/red_run_alert.rs:392-420` collects `handed` and `read` as *key names only* and compares those; the PR body says it "holds the job's env block and the script's reads to exactly match in both directions". Verified: rewriting `WORKFLOW_NAME: ${{ github.event.workflow_run.name }}` (`red-run-alert.yml:73`) to `${{ github.workflow }}` keeps all 10 tests green — and in a `workflow_run` run that expression is the *alert's own* name, so both watched legs collapse into one summons titled "Red run: red-run alert", precisely what `each_watched_workflow_has_a_summons_of_its_own` says is foreclosed. `html_url` → `url` survives too. The sibling convention is to assert the value (`devcontainer_prebuild.rs` compares the `PUBLISHED_IMAGE` literal in both config *and* workflow).
2. **The `gh` lookup's flag surface is untested.** `red_run_alert.sh:28-31`: mutating `--state open --limit 100` → `--state all --limit 1` stays green, because the stub ignores argv past `$1 $2` and only the `--jq` predicate is asserted. `--state all` is a *silent* break — a closed summons matches, so a re-reddened leg comments on a closed issue and never files again.
3. **`--limit 100` plus client-side filtering makes idempotence depend on open-issue count.** `red_run_alert.sh:28-31`. Thirteen open today, but this is a tracker repo by design; once the summons falls out of the newest 100, every red run files a duplicate. A `--search` narrow *plus* the same exact-title `--jq` keeps the exactness the comment argues for without the bound.
4. **The one load-bearing block with no guard, and its claim is wrong.** `red-run-alert.yml:33-38`: removing `concurrency` wholesale stays green, and GitHub cancels a *previously pending* run in a group, so three completions drop the middle reconciliation rather than queueing it.
5. **The new shell file is not enrolled in the repo's shell lint.** `pixi.toml:66` lints only `scripts/without-dl.sh`, run every PR (`devlaunch-contract.yml:87`) under the comment "this repo denies warnings in every other language it is written in." It passes `shellcheck --severity=style` today, so this is a one-line enrolment — but enrolment is exactly the line this repo draws deliberately.
6. **Duplicated Code (Fowler).** `top_level_block` is byte-identical to `devcontainer_prebuild.rs:214-237` (24 lines), the comment-stripper is a rename of `publishing_workflow`, `repo_file` is now its third copy, and `only_the_reconciling_job_may_write_issues` is a structural clone of `only_the_publishing_job_can_write_a_package`. `tests/common/mod.rs` is the standing precedent.
7. Nits: `ci.yml:146`'s "offline like its three siblings above" gains a sibling in this diff and is not updated, while the equivalent AGENTS.md count is — the repo has fixed this exact bug class twice (`0608358`, `efed34e`); the number may already have been stale, so treat as low confidence. `handed` is sorted but not deduped while `read` is (`red_run_alert.rs:401` vs `414`), so a second step setting the same var fails confusingly.

## Spec

Spec is #190, "A red live or scheduled run files an issue". The core shape is met and was traced rather than trusted: permissions are exactly as demanded — *"Keep permissions job-scoped (`issues: write` on that job only)"* — with workflow-level `contents: read` and the single write on `reconcile` (`red-run-alert.yml:44-49`); *"closing itself or commenting on the next green"* holds (`red_run_alert.sh:46-49`); the scheduled contract leg is targeted precisely (`event == 'schedule'` cannot be forged); `workflow_run.name` really is the workflow name, so the name-derived subscription holds; and there is no scope creep. Three gaps.

1. **Blocking — only `conclusion == failure` counts as red.** `red_run_alert.sh:31`, against *"files/updates a single tracking issue … when `live.yml` or the scheduled `devlaunch-contract.yml` run goes red"*. Run with `CONCLUSION=timed_out` or `startup_failure` the script falls to the `else` and does nothing at all — no create, no comment — yet both render as a red X. `startup_failure` is what a broken `live.yml` produces, and `live.yml` declares no `timeout-minutes` (verified), so the leg the spec describes as *"one asserts wall-clock budgets"* lands on GitHub's timeout path rather than a test failure. Red detection wants to be "not `success`, and not `cancelled`/`skipped`" — the script's own comment enumerates only those two exclusions, which is the correct list stated as the wrong condition.
2. **Blocking — the live guard checks the branch but not the event, so a fork PR can withdraw the summons.** `red-run-alert.yml:53-54` requires only `name == 'live' && head_branch == 'main'`, against *"for a non-blocking tier, noticing is the whole game"*. A fork PR whose head branch is `main` (a fork's default) and which adds `pull_request:` to `live.yml` produces a run named `live` with `head_branch == 'main'`; the guard passes, main's script runs with `issues: write`, and the fork chooses the conclusion — a `success` **closes** a genuinely red summons. Verified: the repo is public and `fork-pr-contributor-approval` is `first_time_contributors`, so a prior contributor needs no approval. One clause fixes it: `workflow_run.event == 'push'`, or `head_repository.full_name == github.repository`.
3. **Idempotence rests on an exact title a human will plausibly break.** `red_run_alert.sh:22,28-31`, against *"idempotent (one open issue per workflow, updated not duplicated)"*. Editing the title to add triage context — the natural thing to do on this very issue — makes the next red run file a second one, with both left open. A label survives edits; an exact title does not. Deliberate and argued in the comment, so flagged rather than demanded.

Minor: the alert is itself an unwatched non-blocking leg — a `gh` error goes red with nobody summoned. And the concurrency group can drop a pending reconciliation (found independently here too).

**Operational note, not a defect in this diff:** `live.yml` has concluded `failure` on *every* push to `main` since 2026-08-15 (verified via `gh run list`). This alert is therefore correct to fire the moment it merges — but it will file the summons immediately and add one "Still red" comment per push to main until the underlying live failure is fixed, and a manual close is re-filed within hours. Worth knowing before merging, and arguably worth a ticket on map #182.

## Verdict

**Request changes** — stated in text because GitHub refuses `--request-changes` on a PR authored by the same account, so this is posted as a comment.

Three blocking findings, each cheap to fix:

1. **Spec 1** — `timed_out`/`startup_failure` are red and summon nobody, in a PR whose whole subject is noticing red. Invert the condition to "not `success`, not `cancelled`/`skipped`".
2. **Spec 2** — the live guard is forgeable from a fork PR, which can close a real summons. Add `event == 'push'` or a head-repository check.
3. **Standards 1** — the env contract test compares key names only, so the wiring the PR advertises as held in both directions is not held; `${{ github.workflow }}` collapses both legs into one summons with all tests green. Assert the expression values, as `devcontainer_prebuild.rs` does.

Standards 2 (`--state`/`--limit` unasserted) is close behind: it is the same class as Standards 1 and one assertion away.

Everything else — the `--limit 100` bound, the concurrency claim, the shellcheck enrolment, the shared test-helper duplication, and the two nits — is non-blocking and can ride the same follow-up. The design is right: a script rather than an inline block so the decision can be driven under test, and the subscription pinned to the watched files' own `name:` fields, are both good calls that the fixes above do not disturb.

PR #202, opened 2026-08-22
A red live or scheduled run files an issue

### LINKED ISSUE #190: A red live or scheduled run files an issue

blooop:
Closes #190. Part of map #182.

## What

Nothing summoned anybody when a non-blocking run went red: `live.yml` (post-merge, asserts the tracker's present contents) and `devlaunch-contract.yml`'s weekly re-solve (the only leg that can discover a new devlaunch release) both depended entirely on a human noticing.

- **`.github/workflows/red-run-alert.yml`** — triggered by `workflow_run: completed` of both workflows, subscribed by their declared names. Reacts only to the runs nothing else stands behind: live runs on `main`, and the contract workflow's scheduled leg (every other contract leg blocks a pull request already). Workflow-level permissions are read-only; `issues: write` sits on the one reconciling job. A single fixed concurrency group serialises reconciliations so two completions cannot race one-issue-per-workflow into two.
- **`.github/scripts/red_run_alert.sh`** — the idempotent half: one open tracking issue per watched workflow (exact-title match), filed on red, commented rather than duplicated while the leg stays red, closed with a pointer to the run that cleared it on the next green, untouched by cancelled runs.
- **`tests/red_run_alert.rs`** — the fifth offline files-as-behavior binary. Half of it holds the workflow text: the `workflow_run` subscription is compared to the watched files' own `name:` fields (the subscription is by display name, which GitHub never validates — a rename would disconnect the alert silently), the write grant stays job-scoped, the guards keep it to the non-blocking legs, and the job's env block is held to exactly match the script's reads in both directions. The other half drives the script for real against a recording `gh` stub and asserts the create-vs-comment-vs-close decisions at the boundary they cross.
- Enrolled in `ci.yml` and AGENTS.md's check chain — an assertion no workflow runs is not an assertion.

## Why a script, not an inline run block

So the reconciliation can be executed under test. The recording-`gh` shim follows the same ethos `tests/live_devlaunch.rs` applies to `devpod`: the real decision code runs, and what is asserted is the calls that crossed the boundary.

## Summary by Sourcery

Automatically track red non-blocking workflow runs with idempotent GitHub issue alerts and guard the integration with offline contract tests.

New Features:
- Add automated issue alerts for red non-blocking live and scheduled workflow runs, with per-workflow tracking issues that are created, updated, and closed as runs change state.

Bug Fixes:
- Ensure failures such as timeouts and startup failures, rather than only ordinary failed conclusions, summon maintainers while cancelled and skipped runs remain neutral.

Enhancements:
- Restrict alerts to trusted runs that are not already covered by pull-request gating and serialize reconciliation to prevent duplicate issues.

CI:
- Run the red-run alert contract and behavior tests in CI and include them in the documented verification chain.

Documentation:
- Document the additional offline workflow-behavior test in AGENTS.md.

Tests:
- Add offline tests covering workflow subscriptions, permissions, run guards, environment wiring, concurrency, and issue reconciliation behavior.

### DIFF
--- a/.github/scripts/red_run_alert.sh
+++ b/.github/scripts/red_run_alert.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+# Reconcile the tracking issue for one completed run of a watched workflow
+# (#190). The workflow (`red-run-alert.yml`) decides *which* runs count and
+# hands this script the conclusion; this script's whole job is the idempotent
+# half — one open issue per watched workflow, filed on red, never duplicated,
+# closed on the next green of the same class.
+#
+# A script rather than an inline `run:` block so it can be driven for real:
+# `tests/red_run_alert.rs` runs it against a recording `gh` stub and asserts
+# the create-vs-comment-vs-close decisions at the boundary they cross.
+#
+# Reads: WORKFLOW_NAME, CONCLUSION, RUN_URL, REPO — plus GH_TOKEN, which `gh`
+# itself consumes.
+set -euo pipefail
+
+# The issue's identity. An exact-match title rather than a label, because the
+# search below can then be read literally: rename the workflow and the old
+# issue simply stops matching, which is the correct outcome — a rename also
+# re-points the `workflow_run` subscription (the shape `tests/red_run_alert.rs`
+# holds both ends to).
+title="Red run: ${WORKFLOW_NAME}"
+
+# Exact match on the title, not a substring search: `gh issue list --search`
+# tokenises, and "Red run: live" would find "Red run: live tests, take two".
+# The `--jq` filter is the real gh's own (embedded gojq), so no jq needs to be
+# installed on the runner.
+open_issue="$(gh issue list --repo "${REPO}" --state open --limit 100 \
+  --json number,title \
+  --jq ".[] | select(.title == \"${title}\") | .number" | head -n 1)"
+
+if [ "${CONCLUSION}" = failure ]; then
+  if [ -z "${open_issue}" ]; then
+    gh issue create --repo "${REPO}" --title "${title}" \
+      --body "A non-blocking run of \`${WORKFLOW_NAME}\` went red: ${RUN_URL}
+
+Nothing blocks on this leg, so this issue is the summons. It is updated (not duplicated) while the leg stays red, and closes itself on the next green run of the same class. Filed by \`.github/workflows/red-run-alert.yml\` (#190)."
+  else
+    gh issue comment "${open_issue}" --repo "${REPO}" \
+      --body "Still red: ${RUN_URL}"
+  fi
+else
+  # Green, and only green: the trigger's `types: [completed]` also delivers
+  # cancelled and skipped runs, and those are neither evidence of red nor of
+  # green — closing a summons on a cancelled run would withdraw it on no
+  # evidence at all.
+  if [ "${CONCLUSION}" = success ] && [ -n "${open_issue}" ]; then
+    gh issue close "${open_issue}" --repo "${REPO}" \
+      --comment "Green again: ${RUN_URL}"
+  fi
+fi
--- a/.github/scripts/red_run_alert.sh
+++ b/.github/scripts/red_run_alert.sh
@@ -28,7 +28,27 @@ open_issue="$(gh issue list --repo "${REPO}" --state open --limit 100 \
   --json number,title \
   --jq ".[] | select(.title == \"${title}\") | .number" | head -n 1)"
 
-if [ "${CONCLUSION}" = failure ]; then
+# Red is defined by exclusion — everything that is not one of the three named
+# below — rather than as the single string `failure`. Three conclusions are not
+# red: a green run, and the two non-verdicts, cancelled and skipped, which are
+# evidence of neither red nor green. Every other conclusion GitHub can hand us
+# draws a red X on the Actions tab, and two of those are live on the watched
+# legs: `live.yml` declares no `timeout-minutes`, so the leg that asserts
+# wall-clock budgets overruns into `timed_out` rather than into a failing
+# assertion, and a workflow that stops parsing concludes `startup_failure`
+# before there is a step to fail. Matching on `failure` alone left both of
+# those summoning nobody, which is this workflow's whole subject reintroduced
+# one string comparison in.
+#
+# Naming the exclusions is also safe in a way that naming the inclusions is
+# not: a conclusion GitHub adds later summons somebody by default, and for a
+# leg whose entire point is noticing, that is the right way to be wrong.
+case "${CONCLUSION}" in
+  success | cancelled | skipped) red=false ;;
+  *) red=true ;;
+esac
+
+if [ "${red}" = true ]; then
   if [ -z "${open_issue}" ]; then
     gh issue create --repo "${REPO}" --title "${title}" \
       --body "A non-blocking run of \`${WORKFLOW_NAME}\` went red: ${RUN_URL}
@@ -39,10 +59,10 @@ Nothing blocks on this leg, so this issue is the summons. It is updated (not dup
       --body "Still red: ${RUN_URL}"
   fi
 else
-  # Green, and only green: the trigger's `types: [completed]` also delivers
-  # cancelled and skipped runs, and those are neither evidence of red nor of
-  # green — closing a summons on a cancelled run would withdraw it on no
-  # evidence at all.
+  # Green, and only green, withdraws the summons. The `else` above also holds
+  # the two non-verdicts, and closing on a cancelled run would withdraw a
+  # summons on no evidence at all — so the close is guarded by `success`
+  # itself rather than by "not red".
   if [ "${CONCLUSION}" = success ] && [ -n "${open_issue}" ]; then
     gh issue close "${open_issue}" --repo "${REPO}" \
       --comment "Green again: ${RUN_URL}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,7 @@ jobs:
       - name: The live tests are all gated
         run: cargo test --locked --test offline_green
 
-      # The live fixture's own diagnostic — offline like its three siblings
+      # The live fixture's own diagnostic — offline like its five siblings
       # above — which decides what a failing live test tells
       # you about a missing GH_TOKEN. It is split from the environment it reads
       # precisely so it can be asserted — and an assertion no workflow runs is
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,6 +123,16 @@ jobs:
       - name: The Rust pin has one home
         run: cargo test --locked --test toolchain_pin
 
+      # The fourth config-as-behavior binary: the red-run alert subscribes to
+      # `live.yml` and `devlaunch-contract.yml` by their *display names*, a
+      # string GitHub never validates, so a rename of either workflow would
+      # disconnect the alert with no symptom anywhere — except here, where the
+      # subscription is compared to the watched files' own `name:` fields. The
+      # same binary drives the reconciliation script against a recording `gh`,
+      # offline like its siblings.
+      - name: The red-run alert stays wired to what it watches
+        run: cargo test --locked --test red_run_alert
+
       # The guard on the exclusion every step around it depends on: the live
       # binaries are safe to leave out of the unit-test step only because each
       # of their tests carries `#[ignore]`, and a new one that forgets it is
--- a/.github/workflows/red-run-alert.yml
+++ b/.github/workflows/red-run-alert.yml
@@ -28,10 +28,20 @@ permissions:
   contents: read
 
 concurrency:
-  # Serialised, never cancelled: the reconciliation below is a search-then-act
-  # on the tracker, and two completions racing through it is exactly how "one
-  # open issue per workflow" becomes two. One fixed group — not keyed on the
-  # ref or the event — so every reconciliation queues behind the last.
+  # One fixed group — not keyed on the ref or the event — because the
+  # reconciliation below is a search-then-act on a single tracker issue, and
+  # two completions racing through it is exactly how "one open issue per
+  # workflow" becomes two. `cancel-in-progress: false` keeps the reconciliation
+  # that is already running.
+  #
+  # What this does *not* buy, stated exactly, because the obvious reading is
+  # wrong: reconciliations do not all queue up behind each other. GitHub holds
+  # at most one pending run per group, so a third completion arriving while one
+  # runs and one waits cancels the *waiting* one. That is survivable here, and
+  # only here: the run being cancelled is never the newest, and reconciliation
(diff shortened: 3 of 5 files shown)


blooop (commented):

Reviewed in fresh context against two independent axes, neither aware of the other's findings and neither merged nor reranked. Fixed point: merge-base `4135446`, `git diff 4135446...HEAD` (`.github/scripts/red_run_alert.sh`, `.github/workflows/red-run-alert.yml`, `.github/workflows/ci.yml`, `AGENTS.md`, `tests/red_run_alert.rs`; +566/-3). Preflight: ref resolves, diff non-empty, all checks green. Both axes ran the new binary (10/10) and drove the script for real; Standards mutation-tested the production files in a throwaway worktree.

## Standards

The repo's conventions are matched with care: the files-as-behavior pattern is followed, the heavy why-comment house style is upheld, the permission split matches `package.yml`/`devcontainer.yml`, `set -euo pipefail` and quoting are correct, and the enrolment in `ci.yml` and AGENTS.md's chain honours "an assertion no workflow runs is not an assertion" — including the `Four`→`Five` count. Refuted before reporting: `head -n 1` + `pipefail` SIGPIPE (needs >64KB from `gh`, unreachable), `checkout@v7`, and the `if:` guard's shape.

1. **Blocking — the env block's *values* are unasserted, so the binary's headline claim does not hold.** `tests/red_run_alert.rs:392-420` collects `handed` and `read` as *key names only* and compares those; the PR body says it "holds the job's env block and the script's reads to exactly match in both directions". Verified: rewriting `WORKFLOW_NAME: ${{ github.event.workflow_run.name }}` (`red-run-alert.yml:73`) to `${{ github.workflow }}` keeps all 10 tests green — and in a `workflow_run` run that expression is the *alert's own* name, so both watched legs collapse into one summons titled "Red run: red-run alert", precisely what `each_watched_workflow_has_a_summons_of_its_own` says is foreclosed. `html_url` → `url` survives too. The sibling convention is to assert the value (`devcontainer_prebuild.rs` compares the `PUBLISHED_IMAGE` literal in both config *and* workflow).
2. **The `gh` lookup's flag surface is untested.** `red_run_alert.sh:28-31`: mutating `--state open --limit 100` → `--state all --limit 1` stays green, because the stub ignores argv past `$1 $2` and only the `--jq` predicate is asserted. `--state all` is a *silent* break — a closed summons matches, so a re-reddened leg comments on a closed issue and never files again.
3. **`--limit 100` plus client-side filtering makes idempotence depend on open-issue count.** `red_run_alert.sh:28-31`. Thirteen open today, but this is a tracker repo by design; once the summons falls out of the newest 100, every red run files a duplicate. A `--search` narrow *plus* the same exact-title `--jq` keeps the exactness the comment argues for without the bound.
4. **The one load-bearing block with no guard, and its claim is wrong.** `red-run-alert.yml:33-38`: removing `concurrency` wholesale stays green, and GitHub cancels a *previously pending* run in a group, so three completions drop the middle reconciliation rather than queueing it.
5. **The new shell file is not enrolled in the repo's shell lint.** `pixi.toml:66` lints only `scripts/without-dl.sh`, run every PR (`devlaunch-contract.yml:87`) under the comment "this repo denies warnings in every other language it is written in." It passes `shellcheck --severity=style` today, so this is a one-line enrolment — but enrolment is exactly the line this repo draws deliberately.
6. **Duplicated Code (Fowler).** `top_level_block` is byte-identical to `devcontainer_prebuild.rs:214-237` (24 lines), the comment-stripper is a rename of `publishing_workflow`, `repo_file` is now its third copy, and `only_the_reconciling_job_may_write_issues` is a structural clone of `only_the_publishing_job_can_write_a_package`. `tests/common/mod.rs` is the standing precedent.
7. Nits: `ci.yml:146`'s "offline like its three siblings above" gains a sibling in this diff and is not updated, while the equivalent AGENTS.md count is — the repo has fixed this exact bug class twice (`0608358`, `efed34e`); the number may already have been stale, so treat as low confidence. `handed` is sorted but not deduped while `read` is (`red_run_alert.rs:401` vs `414`), so a second step setting the same var fails confusingly.

## Spec

Spec is #190, "A red live or scheduled run files an issue". The core shape is met and was traced rather than trusted: permissions are exactly as demanded — *"Keep permissions job-scoped (`issues: write` on that job only)"* — with workflow-level `contents: read` and the single write on `reconcile` (`red-run-alert.yml:44-49`); *"closing itself or commenting on the next green"* holds (`red_run_alert.sh:46-49`); the scheduled contract leg is targeted precisely (`event == 'schedule'` cannot be forged); `workflow_run.name` really is the workflow name, so the name-derived subscription holds; and there is no scope creep. Three gaps.

1. **Blocking — only `conclusion == failure` counts as red.** `red_run_alert.sh:31`, against *"files/updates a single tracking issue … when `live.yml` or the scheduled `devlaunch-contract.yml` run goes red"*. Run with `CONCLUSION=timed_out` or `startup_failure` the script falls to the `else` and does nothing at all — no create, no comment — yet both render as a red X. `startup_failure` is what a broken `live.yml` produces, and `live.yml` declares no `timeout-minutes` (verified), so the leg the spec describes as *"one asserts wall-clock budgets"* lands on GitHub's timeout path rather than a test failure. Red detection wants to be "not `success`, and not `cancelled`/`skipped`" — the script's own comment enumerates only those two exclusions, which is the correct list stated as the wrong condition.
2. **Blocking — the live guard checks the branch but not the event, so a fork PR can withdraw the summons.** `red-run-alert.yml:53-54` requires only `name == 'live' && head_branch == 'main'`, against *"for a non-blocking tier, noticing is the whole game"*. A fork PR whose head branch is `main` (a fork's default) and which adds `pull_request:` to `live.yml` produces a run named `live` with `head_branch == 'main'`; the guard passes, main's script runs with `issues: write`, and the fork chooses the conclusion — a `success` **closes** a genuinely red summons. Verified: the repo is public and `fork-pr-contributor-approval` is `first_time_contributors`, so a prior contributor needs no approval. One clause fixes it: `workflow_run.event == 'push'`, or `head_repository.full_name == github.repository`.
3. **Idempotence rests on an exact title a human will plausibly break.** `red_run_alert.sh:22,28-31`, against *"idempotent (one open issue per workflow, updated not duplicated)"*. Editing the title to add triage context — the natural thing to do on this very issue — makes the next red run file a second one, with both left open. A label survives edits; an exact title does not. Deliberate and argued in the comment, so flagged rather than demanded.

Minor: the alert is itself an unwatched non-blocking leg — a `gh` error goes red with nobody summoned. And the concurrency group can drop a pending reconciliation (found independently here too).

**Operational note, not a defect in this diff:** `live.yml` has concluded `failure` on *every* push to `main` since 2026-08-15 (verified via `gh run list`). This alert is therefore correct to fire the moment it merges — but it will file the summons immediately and add one "Still red" comment per push to main until the underlying live failure is fixed, and a manual close is re-filed within hours. Worth knowing before merging, and arguably worth a ticket on map #182.

## Verdict

**Request changes** — stated in text because GitHub refuses `--request-changes` on a PR authored by the same account, so this is posted as a comment.

Three blocking findings, each cheap to fix:

1. **Spec 1** — `timed_out`/`startup_failure` are red and summon nobody, in a PR whose whole subject is noticing red. Invert the condition to "not `success`, not `cancelled`/`skipped`".
2. **Spec 2** — the live guard is forgeable from a fork PR, which can close a real summons. Add `event == 'push'` or a head-repository check.
3. **Standards 1** — the env contract test compares key names only, so the wiring the PR advertises as held in both directions is not held; `${{ github.workflow }}` collapses both legs into one summons with all tests green. Assert the expression values, as `devcontainer_prebuild.rs` does.

Standards 2 (`--state`/`--limit` unasserted) is close behind: it is the same class as Standards 1 and one assertion away.

Everything else — the `--limit 100` bound, the concurrency claim, the shellcheck enrolment, the shared test-helper duplication, and the two nits — is non-blocking and can ride the same follow-up. The design is right: a script rather than an inline block so the decision can be driven under test, and the subscription pinned to the watched files' own `name:` fields, are both good calls that the fixes above do not disturb.

blooop (commented):

Second review of this PR, in a fresh context that had not seen the first report or the revision commits. Fixed point `4135446...c98d5f6`; all checks green at `c98d5f6`; spec is #190. The two axes below were run independently and are not merged or reranked.

## Standards

**The job's `if:` guard is asserted as four loose substrings, so the test survives the exact mutations that defeat it — including one that disables the alert entirely.** `.github/workflows/red-run-alert.yml:69-74`, `tests/red_run_alert.rs:147-190`. The assertions check for the *presence* of `head_repository.full_name == github.repository`, `head_branch == 'main'` and `event == 'schedule'`; nothing checks how they are assembled. Verified by running, in a throwaway worktree at `c98d5f6` (baseline 15/15 green):

- Changing the `&&` on line 71 to `||` re-opens the fork hole this commit says it closed — a fork's forged `success` reaches the one job holding `issues: write` and **closes** a genuinely red summons. **15/15 still green.**
- Changing the leg `||` on line 73 to `&&` makes the condition `name == 'live' && name == 'devlaunch contract'`, never true, so the alert fires for nothing, ever. **15/15 still green.**

The second is why this blocks rather than merely notes: the suite cannot distinguish a working alert from a permanently inert one, and this workflow's whole subject is a leg where nothing else does the noticing. By the standard `ci.yml` states in this same diff — "an assertion no workflow runs is not an assertion" — a guard that survives its own negation is not a guard. The precedent for the fix is one screen down in the same file: `every_variable_describes_the_watched_run_and_not_the_alerts_own` asserts the env *expressions* as literals precisely because a name set is identical whether the values are right. Assert the `if:` the same way, as a normalized whole. **Blocking** — noting the shipped workflow text is correct today, so the scope of the fix is one test.

**The fixed concurrency group can drop a real reconciliation, and its comment argues otherwise.** `red-run-alert.yml:30-46`. `concurrency` is workflow-level while the filter is a job-level `if:`, so every completion of both watched workflows enters the group — including the contract workflow's `pull_request` completions, whose job merely skips. GitHub keeps one pending run per group and cancels the *waiting* one, so a no-op completion can cancel a pending genuine red reconciliation. The block's defence ("the newest completion always gets its say... converges on the current state") does not hold, because a run whose job skips converges nothing. The traffic is real: contract completions land 4-20s apart routinely. `tests/red_run_alert.rs:441` asserts `!group.contains("${{")`, which forbids the obvious fix — and keying on the watched run's name would not suffice anyway, since the contract PR completions share that name; the group belongs on the job, where a skipped job never takes the slot. Self-heals on the leg's next run, hence **non-blocking** — except for the weekly leg, which gets 52 chances a year.

**README's copy of the check chain went stale in this diff.** `README.md:145-155` lists the chain and did not gain `cargo test --locked --test red_run_alert`, while `AGENTS.md:116-119` did. `AGENTS.md:3` calls README "the real documentation" and nothing guards this block, so CI cannot catch it. (README's separate "Four things under `tests/`" paragraph at :212 was already stale before this branch — it omits `toolchain_pin` — so that half is not this PR's to fix.) **Non-blocking.**

**The new script sits outside the repo's shellcheck lint.** `pixi.toml`'s `lint` task names `scripts/without-dl.sh` only, and justifies shellcheck with "this repo denies warnings in every other language it is written in"; `.github/scripts/red_run_alert.sh` is now the most behaviour-carrying shell in the repo. It is `--shell=sh` today and the new script is bash, so this is a second invocation rather than one more word — a fair reason it was deferred, not a reason it is covered. No actual warning claimed; shellcheck was unavailable here. **Non-blocking.**

Nits: `top_level_block` is byte-identical to `tests/devcontainer_prebuild.rs:214`, and `repo_file` appears a third time at `tests/toolchain_pin.rs:22` (Fowler Duplicated Code, test-only). The commit message's parting claim that `ci.yml`'s sibling count "is stale by one more after it" is itself wrong — `ci.yml:146` now reads "five siblings" and five offline binaries do precede it; the base's "three" was the stale one. No other Fowler smell worth flagging.

## Spec

The diff meets #190's stated behavior. Each clause checked against the code rather than against the PR description:

- *"alert on failures of `live.yml` and the weekly devlaunch-contract re-solve"* — the `workflow_run` subscription's `"live"` / `"devlaunch contract"` match the watched files' own `name:` fields exactly, and the test derives them from those files so a rename fails CI. `devlaunch-contract.yml:26-27` does carry the `schedule:` leg, `:80-82` is what re-solves under `if: github.event_name == 'schedule'`, and the guard's `workflow_run.event == 'schedule'` matches how that leg identifies itself.
- *"idempotent (one open issue per workflow, updated not duplicated)"* — identity is the exact title `Red run: <workflow name>`, matched by jq equality rather than a substring search, so the two legs cannot collide and neither fans out.
- *"closing itself or commenting on the next green"* — the close is guarded on `success` alone, so the two non-verdicts withdraw nothing; red is defined by exclusion, so `timed_out` and `startup_failure` summon too.
- *"Keep permissions job-scoped (`issues: write` on that job only)"* — held, with `contents: read` at workflow level and the job's own block restating the checkout's read.

One Spec finding, which both axes reached independently without either knowing the other had: the concurrency race above bears on *"files/updates a single tracking issue on this repo when `live.yml` or the scheduled `devlaunch-contract.yml` run goes red"*, because the run that gets cancelled may be the only one carrying a red payload. Same severity call — **non-blocking**.

Nit, against *"idempotent"*: `.github/scripts/red_run_alert.sh:27` looks the summons up with `--limit 100`, newest-first, so a tracking issue older than 100 open issues stops matching and a second is filed. 13 open issues today; a wayfinder map does file ~10 tickets at a time. **Nit.**

## Verdict

1. **The `if:` guard's boolean assembly is unasserted** (`tests/red_run_alert.rs:147-190`). Two one-character mutations were run and both left 15/15 green: `&&`→`||` on `red-run-alert.yml:71` re-opens the fork hole this commit closed, and `||`→`&&` on line 73 makes the alert permanently inert. Assert the expression as a normalized literal, as this PR already does for the env block.

Everything else is non-blocking and recorded rather than requested: the workflow-level concurrency group's cross-leg drop (found by both axes), README's chain, the shellcheck enrolment, the `--limit 100` bound, and the test-helper duplication.

Out of scope by standing decision and deliberately not counted against this PR: `live.yml` being red on `main` today. An alert that fires on a genuinely red tier is the feature working.

PR #203, opened 2026-08-22
Highlighting and widths survive non-ASCII titles

### LINKED ISSUE #191: Highlighting and widths survive non-ASCII titles
## Question

Make match highlighting and popup widths survive non-ASCII titles.

Two families of unicode confusion, both cosmetic, both real on GitHub-arbitrary titles:

- **Grapheme vs char indices.** nucleo-matcher 0.3.1 segments graphemes (`Utf32Str::new` → `chars::graphemes`, verified in its source), so match indices are *grapheme* indices — but `matched()` checks tightness against `hay.chars()` (`filter.rs:168`), `hit()` splits at `chars().count()` (`filter.rs:177`), and `ui.rs:211` underlines by `chars().enumerate()`. Any title with a multi-codepoint grapheme (combining accent, emoji ZWJ, flag) skews every index after it: wrong tightness decisions, misplaced underlines. No panic — grapheme count ≤ char count.
- **Three width metrics in one renderer.** The launch-picker popup mixes display width `l.width()` with `title.chars().count()` (`ui.rs:838–839`) so CJK/emoji titles under-measure and the border clips the title; `key.len() as u16 + 30` at `ui.rs:909` uses *byte* length as a column width (benign only while repo names stay ASCII); the count line uses `chars().count()` — a third convention.

Fix shape: convert nucleo's indices to char indices at the one boundary where matches are produced (walk the haystack's graphemes once), and standardize the popup/count-line measurements on unicode display width. Tests with an emoji-bearing title and a CJK repo name.

blooop:
Closes #191.

Two families of unicode confusion in the picker, both real on GitHub-arbitrary titles:

- **Grapheme vs char indices.** Nucleo reports *grapheme* indices (its haystack keeps one codepoint per grapheme), while the tightness rule and the screen's underline walk chars — so any title with a flag, ZWJ emoji or combining accent skewed every index after it. Worse than cosmetic: for a decomposed accent, nucleo's own `Utf32Str::new` falls back to byte-indexed ASCII matching, and the byte indices overran the char slice in the tightness check — an out-of-bounds **panic** on a matchable title (pinned by a now-passing test). The filter now segments the haystack itself with the same grapheme walk nucleo uses, judges tightness in grapheme space, and converts to char indices at the one boundary where matches are produced — a matched grapheme lighting every char it spans.

- **Three width metrics in one renderer.** The launch picker measured its title in chars while its rows were measured in display columns (a CJK title clipped its own popup border); the checkout picker used the key's *byte* length plus a magic margin; the count line reserved chars for segments that occupy columns, so a CJK notice lost its tail to the reclaim note. All of it now measures in display columns via ratatui's own span metric — one `popup_width` for the two pickers, one `cols` for the count line.

New dependency edge: `unicode-segmentation` (already in the tree — it is nucleo-matcher's own dependency, same version).

Built test-first per the ticket's seams: a flag-emoji title, a decomposed-accent title, a whole-grapheme highlight, a CJK-heavy launch-picker title, a non-ASCII checkout key, and a CJK notice beside the reclaim note.

## Summary by Sourcery

Make picker matching and terminal layout robust for arbitrary Unicode titles and labels.

Bug Fixes:
- Preserve correct match highlighting for titles containing multi-codepoint graphemes, combining accents, and other non-ASCII text without panics or shifted indices.
- Prevent non-ASCII titles and notices from being clipped in pickers and count-line output.

Enhancements:
- Unify text sizing around terminal display columns for popup widths, line budgets, truncation, and reclaim/liveness messages.

Build:
- Add Unicode segmentation and display-width dependencies for consistent matching and terminal sizing.

Tests:
- Add coverage for emoji and decomposed-accent matching, CRLF titles, CJK picker titles, non-ASCII checkout keys, and wide count-line content.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -679,6 +679,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
+ "unicode-segmentation",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -680,6 +680,7 @@ dependencies = [
  "serde_json",
  "tokio",
  "unicode-segmentation",
+ "unicode-width 0.2.0",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,3 +87,7 @@ tokio = { version = "1", features = [
 # nucleo's own dependency): the filter has to segment identically to map the
 # grapheme indices nucleo reports back onto the chars the screen underlines by.
 unicode-segmentation = "1"
+# The same table ratatui measures spans with (it is ratatui's own dependency):
+# the modules that fit text to a column budget are not UI modules and must not
+# reach for a `Span` to ask how wide a string is. See `cols`.
+unicode-width = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,3 +83,7 @@ tokio = { version = "1", features = [
     "sync",
     "time",
 ] }
+# The same grapheme walk nucleo-matcher builds its haystacks with (it is
+# nucleo's own dependency): the filter has to segment identically to map the
+# grapheme indices nucleo reports back onto the chars the screen underlines by.
+unicode-segmentation = "1"
--- a/src/filter.rs
+++ b/src/filter.rs
@@ -23,6 +23,7 @@ use std::fmt;
 
 use nucleo_matcher::pattern::{CaseMatching, Normalization, Pattern};
 use nucleo_matcher::{Config, Matcher, Utf32Str};
+use unicode_segmentation::UnicodeSegmentation;
 
 use crate::model::{Map, Ticket};
 
@@ -64,6 +65,14 @@ fn haystack(ticket: &Ticket) -> String {
 ///
 /// A pattern of nothing but negations (`!bread`) lands on no characters at all,
 /// and vacuously passes: there is nothing there to be loose about.
+///
+/// Judged in *grapheme* space: `chars` is one codepoint per grapheme — the
+/// haystack exactly as nucleo matched it — and `indices` are nucleo's own.
+/// That is deliberate rather than incidental: judged against the raw chars, a
+/// decomposed accent (a combining mark is not alphanumeric) would open a word
+/// start in the middle of "naïve", and the loose match this rule exists to
+/// refuse would be back for exactly the titles this module was just taught to
+/// survive.
 fn tight(chars: &[char], indices: &[u32]) -> bool {
     let matched = |i: usize| indices.binary_search(&(i as u32)).is_ok();
     indices.iter().all(|&i| {
@@ -144,28 +153,70 @@ impl Query {
     /// project list that matched loosely would be the second matching
     /// vocabulary this rule exists to prevent.
     pub fn text(&mut self, text: &str) -> Option<(u32, Vec<usize>)> {
-        let (score, indices) = self.matched(text)?;
-        Some((score, indices.into_iter().map(|i| i as usize).collect()))
+        self.matched(text)
     }
 
-    /// The pattern against one haystack: the score and the char indices it
+    /// The pattern against one haystack: the score and the *char* indices it
     /// landed on, sorted, unique and tight — or `None`. The one place a match
     /// is decided, so [`Query::hit`] and [`Query::text`] cannot come to differ
-    /// about what counts as one.
-    fn matched(&mut self, hay: &str) -> Option<(u32, Vec<u32>)> {
+    /// about what counts as one — and the one boundary where nucleo's indices
+    /// are translated into the screen's, so nowhere else has to know they were
+    /// ever anything but char indices.
+    ///
+    /// Nucleo matches one codepoint per grapheme (its `Utf32Str::new` keeps
+    /// only the first), so the indices it reports are *grapheme* indices. The
+    /// haystack is segmented here rather than by `Utf32Str::new` for a second
+    /// reason: that constructor falls back to matching the raw *bytes*
+    /// whenever every grapheme's first codepoint is ASCII — a decomposed
+    /// accent takes exactly that path — and byte indices past the accent point
+    /// past the end of the chars they are supposed to name, which was an
+    /// out-of-bounds panic in the tightness check, not merely a skewed
+    /// underline.
+    fn matched(&mut self, hay: &str) -> Option<(u32, Vec<usize>)> {
+        self.buf.clear();
+        self.buf.extend(
+            hay.graphemes(true)
+                .map(|g| g.chars().next().expect("graphemes are non-empty")),
+        );
+        // The ASCII arm is nucleo's fast path, taken on nucleo's own terms: an
+        // ASCII haystack segments to itself, so both arms match the same text.
+        let utf32 = if hay.is_ascii() {
+            Utf32Str::Ascii(hay.as_bytes())
+        } else {
+            Utf32Str::Unicode(&self.buf)
+        };
         let mut indices = Vec::new();
-        let score = self.pattern.indices(
-            Utf32Str::new(hay, &mut self.buf),
-            &mut self.matcher,
-            &mut indices,
-        )?;
+        let score = self.pattern.indices(utf32, &mut self.matcher, &mut indices)?;
         // Nucleo makes no promise about order and can repeat an index across
         // atoms of a multi-atom pattern; both the tightness rule and the screen
         // walk these in step with the characters, so they have to be sorted and
         // unique before either does.
         indices.sort_unstable();
         indices.dedup();
-        tight(&hay.chars().collect::<Vec<char>>(), &indices).then_some((score, indices))
+        if !tight(&self.buf, &indices) {
+            return None;
+        }
+        // Grapheme indices → char indices, walking the graphemes once. A
+        // matched grapheme lights every char it spans — an underline over half
+        // a flag emoji is not a thing the screen can draw — and everything
+        // after a multi-codepoint grapheme lands where the chars actually are.
+        let lit = if hay.is_ascii() {
+            indices.into_iter().map(|i| i as usize).collect()
+        } else {
+            let mut lit = Vec::with_capacity(indices.len());
+            let mut matched = indices.into_iter().peekable();
+            let mut char_at = 0;
+            for (g, grapheme) in hay.graphemes(true).enumerate() {
+                let len = grapheme.chars().count();
+                if matched.peek() == Some(&(g as u32)) {
+                    matched.next();
+                    lit.extend(char_at..char_at + len);
+                }
+                char_at += len;
+            }
+            lit
+        };
+        Some((score, lit))
     }
 
     /// The match, with the characters it landed on — and `None` when it landed
@@ -179,7 +230,7 @@ impl Query {
             score,
             ..Hit::default()
         };
-        for i in indices.into_iter().map(|i| i as usize) {
+        for i in indices {
             if i < repo_len {
                 hit.in_repo.push(i);
             } else if i > repo_len {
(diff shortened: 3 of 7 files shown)


blooop (commented):

Reviewed at the merge-base fixed point `git diff 4135446...73ef2f2` (Cargo.toml, Cargo.lock, src/filter.rs, src/ui.rs). Preflight: ref resolves, diff non-empty, all six checks green. Spec: #191. Two axes run independently in parallel contexts and reported unmerged, below.

## Standards

**Blocking**

**1. The diff introduces a new out-of-bounds panic of exactly the class it exists to remove** (`src/filter.rs:176-198`, reproduced by the reviewer, not by a subagent). `self.buf` is now grapheme-segmented *unconditionally*, but the ASCII arm hands nucleo `Utf32Str::Ascii(hay.as_bytes())` — **byte** indices — and `tight(&self.buf, &indices)` then walks those byte indices into the grapheme-segmented slice. `"\r\n"` is one grapheme cluster and two bytes, so any ASCII haystack containing a CRLF makes `buf` shorter than the index space, and `tight`'s unguarded `chars[i]` (`filter.rs:83`) goes off the end. Titles arrive raw from GraphQL — no newline sanitising anywhere in `fetch.rs`/`model.rs`.

Repro (probe test on the branch head):

```rust
let t = ticket("blooop/wayfinder", 6, "a\r\nzz");
Query::new("zz").unwrap().hit(&t);
// thread panicked at src/filter.rs:83:49:
// index out of bounds: the len is 17 but the index is 17
```

The same probe **passes on `4135446`** (`Some(Hit { in_row: [6, 7] })`) — pre-PR, ASCII gave byte indices into a char slice of equal length. So this is a regression, not a pre-existing hole.

One decision instead of two fixes it, and drops the wasted grapheme walk on the dominant ASCII path at the same time (`buf` is rebuilt per row, per frame):

```rust
let ascii = hay.is_ascii();
self.buf.clear();
if ascii {
    self.buf.extend(hay.chars());
} else {
    self.buf.extend(hay.graphemes(true).map(|g| g.chars().next().expect("graphemes are non-empty")));
}
let utf32 = if ascii { Utf32Str::Ascii(hay.as_bytes()) } else { Utf32Str::Unicode(&self.buf) };
// ... and `let lit = if ascii {`
```

Verified in a scratch worktree: the probe passes and the full suite stays green (437 lib + all integration tests, 0 failed). Worth pinning with a CRLF-title test, since this is the second time the same `chars[i]` has been reached out of bounds.

**2. `src/ui.rs:740-748` — the new function took `centered`'s doc comment.** `popup_width` was inserted directly beneath the `/// A centered box ...` line with no blank line, so that line is now `popup_width`'s rustdoc *summary* — the line rustdoc puts in the module index — and `centered` (:757) is left undocumented. In a repo whose house style is that the doc comments carry the design decisions, a summary describing a different function is a real regression. One blank line plus moving the orphan down.

**Non-blocking**

**3. `src/ui.rs:1002-1009` — the comment overclaims, and the new convention stops one call deep.** "Everything is measured in display columns" is true only of `draw`'s own arithmetic: the two segments that actually *fit text to the budget* still size themselves in chars — `src/liveness.rs:222,226` (`line.chars().count() <= width`) and `src/reclaim.rs:203,216,227`. A wide-char name is still handed a column budget it spends in chars. Also, the comment says the stall names "carry ticket titles"; they don't — `Liveness::hint` names `Node::name`, i.e. `short_repo#number` (`src/reap.rs:544`). Either push `cols()` down into those two, or narrow the comment to what changed. (The Spec axis reached this independently and demonstrated it; see there.)

**Nits**

**4. `src/ui.rs:754` — `.unwrap_or(40)` is unreachable.** `chain(once(title_width))` guarantees an element, so `max()` is never `None`. Dead at both old call sites too, but the extraction was the moment to delete the fallback rather than centralise it.

**5. `src/ui.rs:1155-1163` — `wide()` is only correct for all-wide strings.** It space-separates *every* char regardless of width, so it silently misreads any mixed-width string a future test hands it. Both current callers pass all-CJK strings, so it works today.

**Clean elsewhere.** The grapheme→char translation is confined to the single `matched` seam as its doc claims; whole-grapheme lighting composes correctly with `lit_spans`' run coalescing (`ui.rs:202-226`), so a flag stays one span; `unicode-segmentation` adds no new `[[package]]` to `Cargo.lock` and the direct dep carries its rationale. New tests follow the repo's naming and why-first commenting.

## Spec

**The diff meets the ticket, with one residual.** Each half verified by reverting it in a scratch copy of the branch.

**"convert nucleo's indices to char indices at the one boundary where matches are produced" — met.** `pattern.indices` appears exactly once in the tree (`src/filter.rs:189`); `text()` and `hit()` both route through `matched()`, and every consumer is char-indexed and consistent (`ui.rs:211` `lit_spans`, `filter.rs:230` `repo_len`). No surviving byte/char indexing of a haystack against nucleo indices.

**The panic is real; the ticket is wrong.** The ticket asserts *"No panic — grapheme count ≤ char count"*. Reverting `matched()` reproduces `index out of bounds: the len is 23 but the index is 23` at `filter.rs:83`. Cause confirmed in nucleo's source (`utf32_str.rs:44-55`): `Utf32Str::new` returns `Ascii(str.as_bytes())` whenever every grapheme's *first* codepoint is ASCII, so a decomposed accent yields byte indices — and that slice also violates the `Ascii` variant's own documented "must only contain valid ASCII" invariant. The extra scope is justified and lands on the same one boundary; nothing unrelated rides along.

**"standardize the popup/count-line measurements on unicode display width" — the three named sites are all fixed.** `l.width()`/`title.chars().count()` (was ui.rs:838-839) and `key.len() as u16 + 30` (was ui.rs:909) both go through the single `popup_width`, and those are the only two popups in the file; the count line uses `cols`.

**Finding (non-blocking): the count line's char metric survives one call deep.** Against *"the count line uses `chars().count()` — a third convention"* and *"standardize the popup/count-line measurements on unicode display width"*: the caller now budgets in columns (`ui.rs:1027`), but `Liveness::hint` still picks which variant fits with `line.chars().count() <= width` (`liveness.rs:222,226`). Stall names are `short_repo#number` (`reap.rs:544`) — and this PR's own new test posits `blooop/测试仓库` as realistic input. Demonstrated in scratch: two stalls on `blooop/测试仓库仓库` plus a reclaimable set renders, at width 55, `· 2 stalled: 测试仓库仓库#9, +1 more · 2 recla` — the exact fragment that `neither_count_line_segment_can_clip_the_other_into_a_fragment` (`ui.rs:1671`) exists to forbid. `reclaim.rs:203-238` has the same shape, though `dl` ids are likelier to stay ASCII.

**Nit: one of the six tests is a pin, not red-first.** The breadcrumb claims "Six red-first tests"; five are. `the_checkout_picker_title_survives_a_non_ascii_key` **passes** against the old `key.len() as u16 + 30`, because UTF-8 bytes over-count columns for every wide char — which makes the ticket's *"benign only while repo names stay ASCII"* wrong too. The change is still right (it kills a magic margin), but the claim overstates it.

**Tests required by the spec** ("Tests with an emoji-bearing title and a CJK repo name") — present, and the other four are genuinely red-first.

## Verdict

**Request changes.**

Blocking:

1. **The new CRLF panic in `matched()`/`tight()`** (`src/filter.rs:176-198`) — a regression against `4135446`, in the same `chars[i]` this PR was written to stop overrunning. Proven fix above, full suite green with it.
2. **`popup_width` swallowed `centered`'s doc summary** (`src/ui.rs:740-757`) — one blank line, and give `centered` its line back.

Not blocking, but both axes found #3 independently and it is the honest completion of this ticket's second half: either push `cols()` into `Liveness::hint` and `Reclaimable::hint`, or narrow the `draw` comment so it stops claiming a convention that ends at the call boundary. The two overstated claims (the ticket's "no panic", the breadcrumb's "six red-first") are worth correcting in the record, not in the code.

Everything else is clean, and the central grapheme→char translation is the right shape in the right place.

blooop (commented):

Two-axis review of `wayfinder/wayfinder-191` against merge-base `4135446` (`git diff 4135446...HEAD`), spec = #191 **as corrected by its own trail** (the body's "No panic — grapheme count ≤ char count" is false; the correcting breadcrumb supersedes it). Preflight: ref resolves, diff non-empty, all checks green at `568925f`. The axes were run in parallel and are reported unmerged.

## Standards

**1. `clip` measures in a metric `cols` does not use, and the count-line segment overruns its budget again — blocking.**

`src/reclaim.rs` `clip` charges each char on its own (`crate::cols(ch.encode_utf8(&mut buf))`) and sums. `crate::cols` is unicode-width 0.2's `UnicodeWidthStr::width`, which is *not* that sum: it is a right-fold carrying `WidthInfo`, with a **signed** per-char contribution (`unicode-width-0.2.0/src/tables.rs:398-408`). U+FE0F retroactively widens the char before it from 1 column to 2 — after `clip` has already charged 1 for it and 0 for the selector. So `clip` admits a pair that costs 2 columns for the price of 1.

Reproduced end-to-end on an id carrying `▶\u{FE0F}` — `Reclaimable::hint` returns a segment **wider than the width it was handed**:

| width given | columns returned |
|---|---|
| 81 | 84 |
| 88 | 91 |

It overruns at every width 81–88. This is a **regression, not pre-existing**: the identical probe over widths `0..=140` with the identical ids is clean at `4135446`, where the ladder counted chars and `▶\u{FE0F}`'s two chars happened to equal its two columns.

It also falsifies three claims this PR adds — `clip`'s "As much of `chars` as fits `budget` columns", `abbreviate`'s "shortened in the middle to `max` columns", and `Reclaimable::hint`'s "`cols` is the only metric in here" — and `README.md`'s "The segment never overruns the width it is given at any terminal size." The new property test at `src/reclaim.rs:743` misses it because it only tries CJK, where per-char summing happens to coincide with `str_width`.

Fix shape: never sum per-char — grow a candidate prefix and measure the whole with `cols`, or use `unicode-truncate`, already in the lock via ratatui 0.29 and both grapheme- and width-aware.

**2. `the_checkout_picker_title_survives_a_non_ascii_key` (`src/ui.rs:2269`) pins nothing — non-blocking.** Its comment says it stops a cheaper metric coming back, but the `/data/k1/repo` body rows outweigh the title: making `popup_width` ignore the title entirely (`body + 4`) leaves this test green while its sibling at `ui.rs:2355` goes red. It needs a key that dominates the body.

**3. `src/filter.rs:175`'s doc overstates — non-blocking.** "Neither arm may be left to `Utf32Str::new`, and the two fail in opposite directions." For an all-ASCII haystack `Utf32Str::new` returns `Ascii(bytes)` without segmenting (`nucleo-matcher-0.3.1/src/utf32_str.rs:45`), so the CRLF direction was never a failure of `new` — it was a failure of filling `buf` with graphemes unconditionally, introduced in `f996cd4` and fixed in `df3bd3f` inside this PR. Confirmed at the merge-base: an all-ASCII `a\r\nzz` title does **not** panic there, while `Cafe\u{301} menu` does, at `filter.rs:74`. The doc credits `new` with a bug that was this branch's own.

**4. Minor — all non-blocking.** `src/filter.rs:189-233` switches on `ascii` three times; the first two merge (the `&mut self.buf` borrow ends before `&self.buf`). `src/ui.rs:1013` `let cols = crate::cols;` is a local alias while `popup_width` fully qualifies — one `use crate::cols;` serves both. `src/ui.rs:1711` is near-verbatim `ui.rs:1678`, and `liveness.rs:480` near-verbatim `liveness.rs:455`; parameterising over the repo name would also stop the originals asserting in the abandoned convention (`liveness.rs:460`, `reclaim.rs:629`). `src/ui.rs:1163` `wide()` is correct only when *every* char is double-width, not for "a wide string" as named.

**Refuted and dropped.** `matched`'s `buf`↔index-space invariant survived a 40k-case fuzz (CRLF, combining marks, ZWJ, regional indicators, VS, jamo, NUL, BOM): no panic, indices in range, graphemes lit whole, sorted and unique. `cols` really is ratatui's metric (`Span::width` → `UnicodeWidthStr::width`; ratatui 0.29 → unicode-width 0.2.0) and the lock gains no package. `cols` in `lib.rs` is justified — `liveness`/`reclaim` import no ratatui. `popup_width`'s dropped `unwrap_or(40)` was dead. `abbreviate`'s reversed clip can strand a zero-width mark on the ellipsis (`wf-…\u{301}-127`), but the char-count code did the same at a different budget — pre-existing, dropped.

## Spec

**The diff meets the corrected spec. No blocking findings.**

**Half 1 — "convert nucleo's indices to char indices at the one boundary where matches are produced."** Met. `matched` (`src/filter.rs:196-234`) is the only nucleo call site in the tree, and it now returns char indices, so `tight` (`filter.rs:77`), the `hit()` split at `short_repo().chars().count()` (`filter.rs:246`) and the underline at `ui.rs:211` share one index space — `ui.rs:211` correctly left untouched. The corrected panic claim is closed in **both** directions: an independent integration probe (26 titles × 6 repos × 23 queries, asserting in-bounds indices, whole-grapheme lighting in both halves, and that a literal query lands on its own substring) panics at `4135446` — `index out of bounds: the len is 23 but the index is 23` at `filter.rs:74`, the decomposed-accent byte-index arm — and passes at HEAD, including `a\r\nzz`, `zz\r\n`, `a\u{301}\r\nzz`, ZWJ families, flags, ZWSP and CJK.

**Half 2 — "standardize the popup/count-line measurements on unicode display width."** All three named sites converted: `popup_width` (`ui.rs:753`) for both pickers (`ui.rs:852`, `ui.rs:918`), and the count line (`ui.rs:1013-1035`). No surviving char- or byte-as-width in non-test code on those paths. `cols` resolves unicode-width 0.2.0, the version ratatui 0.29 already pulls, so the "same table" claim holds and no lock package is added.

**Tests it asks for** — "an emoji-bearing title and a CJK repo name": both present (`filter.rs:427`, `filter.rs:443`; `ui.rs:1711`, `liveness.rs:479`), and red-first confirmed by reverting each fix in a scratch copy: 3 filter tests red without `matched`, 3 ui tests red without `cols`, 2 segment tests red on char metrics.

- **Non-blocking** — `the_checkout_picker_title_survives_a_non_ascii_key` (`ui.rs:2269`) is a pin, not a test of the fix: restoring `key.len() as u16 + 30` alone leaves all 440 tests green. The trail already concedes this ("is five, not six").
- **Non-blocking** — two older assertions still guard the fits-its-budget invariant in the abandoned convention: `liveness.rs:459` and `reclaim.rs:629`.

Refuted and dropped: the reclaim fragment at widths 16–23 (byte-identical for ASCII and CJK, pre-existing, documented at `reclaim.rs:174`); an apparent ZWJ title clip (the oracle, not the code); the removed `unwrap_or(40)` (both pickers push rows unconditionally). The `reclaim`/`liveness`/`abbreviate` conversions and `pub fn cols` reach past the three named lines but sit inside the quoted second half — not scope creep.

## Verdict

- **Standards 1** — `reclaim::clip` sums per-char widths, which is not what `crate::cols` computes, so `Reclaimable::hint` overruns the column budget it is handed (81 given → 84 returned) on an id carrying a variation selector. Proven a regression against `4135446`, and it is the same class of defect this ticket exists to remove, in the very half of it this PR added.

Spec passes on its own axis; nothing else here blocks. The remaining items are non-blocking and can ride with the fix or be left: the two tests that pin nothing (`ui.rs:2269`, and the char-convention ladders at `liveness.rs:459` / `reclaim.rs:629`), the `filter.rs:175` doc that credits `Utf32Str::new` with this branch's own bug, and the minor duplication and naming notes.

Issue #132, opened 2026-08-09
Reap's remaining sharp edges: an unknown issue state deletes, and two mutations the suite can't see


blooop:
## Question

The fresh two-axis review of [PR #130](https://github.com/blooop/wayfinder/pull/130) approved the slice with zero blocking findings, but raised six should-fixes that were deliberately not held against the merge. They are collected here so they are worked once, deliberately, rather than rediscovered one at a time. The largest is a hazard on a **deleting** arm.

### 1. An unrecognised issue state falls into the deleting arm

`src/reap.rs` derives `NodeFact::Closed` — which routes to `Reap`, i.e. deletion — for an issue state it does not recognise. `main`'s prior `issue_is_closed` was a positive `== "closed"` test, so an unknown state **kept** the workspace.

This is the exact mirror of the defect #129's amendment exists to kill (an unreadable value sliding into a definite arm), except pointed at the arm that *deletes* rather than the one that warns. It is currently a hazard rather than a live bug — GitHub's `IssueState` is `OPEN|CLOSED` today, and the derivation order is what [#128](https://github.com/blooop/wayfinder/issues/128) specifies — but the asymmetry is the point: the PR-reading projection already resolves "unreadable" to the arm that is safe whatever the new value means (in-flight, keeps, prints nothing), and the issue-state reading should be held to the same standard. Decide and pin which arm an unknown issue state belongs to.

### 2. Two real regressions survive the suite

Both were introduced by mutation and stayed green at 281/281:
- swapping the `DoneByMerge`/`Superseded` precedence;
- gutting `NODE_FACT_SELECTION`.

The behaviour in the diff is correct; the **pinning is thin**. A test that cannot tell these two mutations from the real thing is not guarding the precedence rule or the selection.

### 3. The GraphQL PR selection is duplicated verbatim

The same selection appears in both `src/fetch.rs` and `src/reap.rs`, with **no test binding either query to the types it feeds**. Making `parse_pr` `pub(crate)` moved the *type* into one vocabulary but left the *query text* in two — so the two can drift silently, which is the failure the one-vocabulary constraint was meant to prevent.

### 4. `README.md` misstates the keep rule

It lists "tickets someone has claimed" under **Kept, always**. A claimed ticket whose PR merged is `DoneByMerge` and is deleted. The docs are part of `wf`'s interface (`AGENTS.md`), so this is a real defect, not a typo.

### Seams to test at

[...]

Issue #137, opened 2026-08-09
wf notices what reap would claim, without being asked


blooop:
**Type:** build
**Map:** #121

User steer (2026-08-09): cleanup should happen automatically, because `wf` — unlike `dl` — knows when a ticket is done.

`wf` already agrees in principle. `src/reap.rs` states the division as **`dl` owns the containers, `wf` owns the tickets**: dl publishes what it knows and refuses to destroy work that exists nowhere else, and wf, which minted the branch names from its own tickets, decides what is finished. What is missing is not the decision — `reap::plan` and `reap::doomed` already make it — but the *trigger*: today it fires only when a person remembers to type `wf reap`.

This ticket automates the **noticing**, and deliberately not the deleting. See #138 for the deleting half and the three things that block it.

## Why the reading, and not the deletion, is what ships first

Automatic deletion is blocked three times over right now, and none of the three is speculative:

1. **[#132](https://github.com/blooop/wayfinder/issues/132) item 1** — an unrecognised issue state falls into `NodeFact::Closed`, which routes to `Reap`, i.e. **deletion**. `main`'s prior positive `== "closed"` test kept such a workspace. A hazard rather than a live bug while GitHub's `IssueState` stays `OPEN|CLOSED`, but it is pointed at the deleting arm and it is unfixed.
2. **[#132](https://github.com/blooop/wayfinder/issues/132) item 2** — two real regressions survive the suite at 281/281 green, one of them a swap of the `DoneByMerge`/`Superseded` precedence. The deletion decision's pinning is thin by the repo's own finding.
3. **blooop/devlaunch[#171](https://github.com/blooop/devlaunch/issues/171)** — `reap::plan` keeps a workspace when `workspace.unsaved` is `Some`, and that field is **dl's**, read here rather than argued with, exactly as this module's docstring says. Reproduced against merged devlaunch `main`: dl's `holds_unsaved_work` runs git with no `--git-dir` and no ceiling, so on a clone whose `.git` is unusable git's discovery walks up and answers about an **ancestor** repository — returning "nothing would be lost" for a clone holding untracked work. wf's keep-guard therefore inherits a reading that can be confidently wrong in the unsafe direction.

A human running `wf reap` today sees the verdict list and confirms. Automating the deletion removes precisely that reader, on top of a keep-guard that can fail open. Automating the *reading* removes the chore without removing the reader.

[...]

PR #204, opened 2026-08-22
The screen says when a map fetch was truncated

### LINKED ISSUE #184: The screen says when a map fetch was truncated

blooop:
Closes #184.

**Stacked on #199** (`wayfinder/wayfinder-183`) and should merge after it — this branch depends on the query text #183 changed (the shared linked-PR selection block that `reap` byte-compares against the map query). Retargeting to `main` happens automatically when #199 merges.

## What arrived silently partial, and where it now shows

The map fetch capped `subIssues(first: 100)`, `blockedBy(first: 50)`, and `labels(first: 20)` with no `pageInfo` and no indicator, so:

- a 101st ticket was dropped without a trace;
- a 51st blocking edge could misclassify a node as unblocked;
- worst, a map wearing 21+ labels with `wayfinder:map` off the fetched page failed re-verification as "no longer a map" on every refresh, while discovery's label-scoped search kept re-finding it — permanently.

## The three seams, test-first

1. **The parse boundary** (`parse_map`): the two ticket-bearing connections — the sub-issue page and each ticket's blocker page — now report `pageInfo.hasNextPage`, folded into `Map::truncated`. Absent `pageInfo` reads as complete, the same defaulting rule the PR selection follows. Partial still beats absent: a truncated map parses and renders.
2. **The query text**: `MAP_QUERY` asks for every page boundary the parse reads, guarded against the query source the way #132 guards reap's batch — fixture-fed parse tests cannot notice a live query that stopped asking. The guard is depth-aware so the linked-PR block's own `pageInfo` (#183's, deliberately untouched here) cannot answer for a connection that never asked. The map's own label page runs to the GraphQL page maximum (100) and carries `pageInfo`; only a *complete* page without the label may reject a cached number — a truncated one keeps the map, so label count can never un-map a map.
3. **The count line**: a truncated map is named there (`· repo#n truncated`; several collapse to `· N maps truncated`), symmetric with the failed-map segment — the body draws a normal-looking tree either way, so the count line is the only place left to say the tree is partial.

The shared PR selection block is byte-identical to #199's; `both_queries_ask_for_a_linked_pr_in_exactly_the_same_words` stays green.

Full AGENTS.md check chain green locally (fmt, clippy, tests, doc — all under `-D warnings`).

## Summary by Sourcery

Track and visibly report partial map fetches without rejecting maps based on incomplete label data.

New Features:
- Expose when fetched maps are partial due to truncated sub-issue or blocker data.
- Display truncation status on the map count line, naming a single map or collapsing multiple maps into a count.

Bug Fixes:
- Prevent incomplete label pages from incorrectly invalidating cached maps that still have the map label.
- Preserve and surface pagination signals so missing tickets or blocking edges are not silently treated as complete data.

Enhancements:
- Expand map queries to request pagination metadata for ticket-bearing connections and fetch the map's labels up to the GraphQL page maximum.
- Add query-shape checks ensuring all pagination boundaries consumed by map parsing remain selected.

Tests:
- Add coverage for truncated sub-issue and blocker pages, missing pagination metadata, label-heavy maps, query pagination selections, and truncation rendering.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -1407,6 +1407,7 @@ mod tests {
             Map {
                 title: "Map: wf".to_string(),
                 last_activity: None,
+                truncated: false,
                 tickets: vec![
                     ticket(
                         "blooop/wayfinder",
@@ -1448,6 +1449,7 @@ mod tests {
             Map {
                 title: "Map: dotfiles".to_string(),
                 last_activity: None,
+                truncated: false,
                 tickets: vec![ticket(
                     "blooop/dotfiles",
                     103,
@@ -1468,6 +1470,7 @@ mod tests {
             Map {
                 title: format!("Map: {repo}"),
                 last_activity: stamp.map(|s| Activity::parse(s).expect("fixture stamp parses")),
+                truncated: false,
                 tickets: vec![ticket(repo, 1, "only ticket", open, false, vec![])],
             },
         )
@@ -1657,6 +1660,7 @@ mod tests {
                     ticket(PROJECT, 1, "takeable", true, false, vec![]),
                     ticket(PROJECT, 2, "finished", false, false, vec![]),
                 ],
+                truncated: false,
             },
         )])
     }
@@ -1718,6 +1722,7 @@ mod tests {
             Map {
                 title: "Map: late arrival".to_string(),
                 last_activity: None,
+                truncated: false,
                 tickets: vec![ticket(PROJECT, 200, "zzz sleeper", true, false, vec![])],
             },
         );
@@ -1754,6 +1759,7 @@ mod tests {
                     Activity::parse("2026-08-07T00:00:00Z").expect("fixture stamp parses"),
                 ),
                 tickets: vec![ticket(PROJECT, 200, "breadwinner", true, false, vec![])],
+                truncated: false,
             },
         );
         app.replace_clusters(fresher);
@@ -1869,6 +1875,7 @@ mod tests {
             Map {
                 title: "Map: wf".to_string(),
                 last_activity: None,
+                truncated: false,
                 tickets: vec![
                     ticket("blooop/wayfinder", 6, "root", true, false, vec![]),
                     ticket("blooop/wayfinder", 7, "dep a", true, false, vec![6]),
@@ -1901,6 +1908,7 @@ mod tests {
             Map {
                 title: "Map: endgame".to_string(),
                 last_activity: None,
+                truncated: false,
                 tickets: vec![
                     ticket("blooop/bencher", 1, "done", false, false, vec![]),
                     ticket("blooop/bencher", 10, "root, chained", true, false, vec![]),
@@ -2038,6 +2046,7 @@ mod tests {
             Map {
                 title: "Map: endgame".to_string(),
                 last_activity: None,
+                truncated: false,
                 tickets: vec![
                     ticket("blooop/bencher", 1069, "root", true, false, vec![]),
                     ticket(
@@ -2443,6 +2452,7 @@ mod tests {
                 Map {
                     title: format!("Map: {map_number}"),
                     last_activity: None,
+                    truncated: false,
                     tickets: vec![ticket(
                         "blooop/wayfinder",
                         6,
@@ -3255,6 +3265,7 @@ mod tests {
                 Map {
                     title: "Map: the spine".to_string(),
                     last_activity: None,
+                    truncated: false,
                     tickets: vec![t],
                 },
             );
@@ -3358,6 +3369,7 @@ mod tests {
             Map {
                 title: "Map: wf".to_string(),
                 last_activity: None,
+                truncated: false,
                 tickets: vec![ticket("blooop/wayfinder", 6, "t6", true, false, vec![])],
             },
         );
@@ -3366,6 +3378,7 @@ mod tests {
             Map {
                 title: "Map: selection view".to_string(),
                 last_activity: None,
+                truncated: false,
                 tickets: vec![ticket("blooop/wayfinder", 50, "t50", true, false, vec![])],
             },
         );
@@ -3374,6 +3387,7 @@ mod tests {
             Map {
                 title: "Map: dotfiles".to_string(),
                 last_activity: None,
+                truncated: false,
                 tickets: vec![ticket("blooop/dotfiles", 103, "t103", true, false, vec![])],
             },
         );
@@ -3394,6 +3408,7 @@ mod tests {
                 Map {
                     title: "Map: dotfiles".to_string(),
                     last_activity: None,
+                    truncated: false,
                     tickets: vec![ticket(
                         &format!("{owner}/dotfiles"),
                         5,
--- a/src/fetch.rs
+++ b/src/fetch.rs
@@ -729,7 +729,10 @@ mod tests {
 
         // And the label page being cut short is not the *tree* being cut
         // short: nothing ticket-bearing was truncated here.
-        assert!(!map.truncated, "a truncated label page is not a partial tree");
+        assert!(
+            !map.truncated,
+            "a truncated label page is not a partial tree"
+        );
     }
 
     /// One sub-issue carrying the full PR-badge matrix (#52).
@@ -871,8 +874,8 @@ mod tests {
                 "subIssues more: {subs}, blockers more: {blockers} — the map must say so"
             );
         }
-        let map = parse_map(truncated_response(false, false).as_bytes(), &wf_map_id())
-            .expect("parse");
+        let map =
+            parse_map(truncated_response(false, false).as_bytes(), &wf_map_id()).expect("parse");
         assert!(!map.truncated, "full pages are not a truncation");
     }
 
--- a/src/fetch.rs
+++ b/src/fetch.rs
@@ -888,64 +888,150 @@ mod tests {
         assert!(!map.truncated);
     }
 
-    /// The brace-balanced block a named connection selects in `query` —
-    /// `field(…) { … }` from the first `{` to its close. Panics if the field
-    /// is missing or its braces never balance, because a guard that returns
-    /// `None` quietly is a guard that stops guarding when the query is
-    /// reworded.
-    fn connection_block<'q>(query: &'q str, field: &str) -> &'q str {
-        let from = query
-            .find(field)
-            .unwrap_or_else(|| panic!("the query no longer selects {field}"));
-        let mut depth = 0usize;
-        for (offset, ch) in query[from..].char_indices() {
+    /// One field of a GraphQL selection: the arguments it was called with, and
+    /// the fields selected under it — which is to say, the shape an answer to
+    /// that selection has.
+    #[derive(Default)]
+    struct Selected {
+        arguments: String,
+        fields: std::collections::BTreeMap<String, Selected>,
+    }
+
+    impl Selected {
+        /// The field reached by following `path` from here, or `None` if the
+        /// query does not select it *there*.
+        ///
+        /// Position is the whole point. The sub-issue block contains the
+        /// linked-PR block, whose own `pageInfo` (#183) would answer on behalf
+        /// of a connection that never asked if the question were "does this
+        /// name appear anywhere below".
+        fn at(&self, path: &[&str]) -> Option<&Selected> {
+            match path {
+                [] => Some(self),
+                [step, rest @ ..] => self.fields.get(*step)?.at(rest),
+            }
+        }
+
+        /// The field at `path`, created empty if the walk has not reached it
+        /// yet — how the parse below fills the tree in as it descends.
+        fn under(&mut self, path: &[String]) -> &mut Selected {
+            let mut at = self;
+            for step in path {
+                at = at.fields.entry(step.clone()).or_default();
+            }
+            at
+        }
+    }
+
+    /// Read the subset of GraphQL selection syntax [`MAP_QUERY`] is written in
+    /// — names, optional `(arguments)`, optional nested `{ blocks }` — as the
+    /// tree of fields it selects.
+    ///
+    /// Derived rather than matched, and that is the point of it. A guard that
+    /// searches the query text for a spelling is satisfied by any text carrying
+    /// that spelling: `pageInfo { endCursor }` passed a guard looking for
+    /// `pageInfo` while answering none of the questions the parse asks of it.
+    /// Walking the tree asks the parse's own question instead — is this field
+    /// selected at this path — which no rewording can answer by accident.
+    ///
+    /// Panics on unbalanced braces, because a guard that gives up quietly on a
+    /// query it cannot read is a guard that stops guarding the moment the query
+    /// is reworded.
+    fn selected(text: &str) -> Selected {
+        let mut root = Selected::default();
+        let mut path: Vec<String> = Vec::new();
+        let mut last: Option<String> = None;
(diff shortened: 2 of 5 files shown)


blooop (commented):

Reviewed at the fixed point `git diff 9c0fdf8...970fb0c` (three-dot against this PR's own base, `wayfinder/wayfinder-183` — not main). Preflight: ref resolves, diff non-empty (325 insertions across 5 files), all five checks green. Spec = #184. Two axes below, run independently and **not** merged or reranked.

## Standards

- **`src/fetch.rs:922` — the new query guard does not guard what it claims. (blocking)** `asks_its_own_page_info` matches the *substring* `"pageInfo"` at depth 1; it never asserts the `hasNextPage` the parse actually reads. Both `Paged::page_info` and `PageInfo::has_next_page` are `#[serde(default)]`, so a reworded selection deserialises silently as "no claim of more". Proven by mutation: rewriting the three `pageInfo { hasNextPage }` selections in `MAP_QUERY` to `pageInfo { endCursor }` leaves all 443 lib tests green — including `the_query_asks_for_every_page_boundary_the_parse_reads` and `the_maps_own_label_page_is_as_deep_as_the_tracker_allows`. Every map then reads as complete forever, which is exactly the pre-#184 behaviour these tests exist to forbid. Fix: assert the depth-1 block contains `pageInfo { hasNextPage }`, or derive the check from reap's `parse_selection`/`prune` (`src/reap.rs:2723`/`:2768`), which would catch this automatically.
- **`src/fetch.rs:896` — duplicated brace-walker. (non-blocking, Duplicated Code)** `connection_block` is `src/reap.rs:2661`'s `pr_selection` generalised, near line-for-line including the `checked_sub` trick and its panic wording. `src/probe.rs` is the repo's documented home for test scaffolding shared by more than one module, and its own header records `code_only` being written twice byte-for-byte before it landed there. Lift the extractor there and have `pr_selection` call it.
- **`src/ui.rs:598` — `truncated_note` bypasses `App::scoped_clusters` (`src/app.rs:477`). (non-blocking)** It iterates `app.clusters` raw, so it names maps outside the current scope. Proven render with two projects, scope on `blooop/wayfinder`: `  5/5  · blooop/elsewhere#9 truncated` — a map with no row on screen and outside the `5/5` it sits beside. `failure_note`'s population (`app.failed`) has no scope; clusters do. (The Spec axis independently judged this consistent with `failure_note` rather than a divergence — see below. Left unreconciled by design.)
- **`src/ui.rs:1677` — the count-line clipping guard was not extended. (non-blocking)** Add `truncated` to its word list and set the flag on its fixture and it goes red: widths 43-54 render `· 2 r` … `· 2 reclaimabl`, the #137 fragment. Pre-existing in kind (`app.failed` fragments at 40-51 on the base), so not introduced here — but this PR adds a third named segment ahead of the only segment that clips instead of vanishing, without touching the guard or the `reserved` accounting.
- **Checked, no finding:** `Paged` vs `Nodes` (split is documented; readers of `.nodes` unaffected); `Map::truncated` as a bool (no sum type it can disagree with, not derivable from `tickets` — clears the AGENTS.md "no bool beside a sum type" rule); `truncated |=` inside `.map()` (eager `collect`, so correct); `asks_its_own_page_info`'s raw `-=` (cannot underflow on a balanced block); the ~30 mechanical `truncated: false` fixture edits (no `Map` builder exists to avoid them, and nothing in the repo requires one).

## Spec

**Spec satisfied** on every enumerated consequence. All three of the ticket's named failures are discharged, and the drift test the ticket names as its precondition is green.

- *"a map issue with 21+ labels where `wayfinder:map` is not in the first 20 fails the re-verification at `fetch.rs:333` and is rejected as 'no longer a map'"* — `src/fetch.rs:70` raises the map's own label page to the GraphQL maximum and `src/fetch.rs:370-371` lets only a **complete** page reject. Exhaustively: ≤100 labels → complete page, correct verdict; >100 → `hasNextPage` true → rejection impossible. No residual rejection path. The opposite stall (a map that genuinely lost the label but wears 100+) is refuted: `replace_clusters` (`src/app.rs:836`) rebuilds clusters from the label-scoped search, so an unlabelled map drops out of `clusters` on the next discovery pass regardless of what the per-map fetch believes. `first: 100` plus truncation-tolerance is a faithful reading of *"filter by label in the query, or raise the cap well past plausible."*
- *"fetch `pageInfo.hasNextPage` on the ticket-bearing connections"* — both are covered (`src/fetch.rs:66,74`), read at `src/fetch.rs:384,390`, and guarded against query drift by the depth-aware `asks_its_own_page_info` (`src/fetch.rs:906`), which correctly refuses to let #183's nested linked-PR `pageInfo` answer for `subIssues`.
- *"say so on the count line when a map arrived truncated"* — reachable, not test-only: `src/fetch.rs:426-431` is the sole non-test `Map` construction in the tree, and `src/ui.rs:598` is wired into `draw`'s fixed segment list at `src/ui.rs:1000`. Its global (unscoped-by-repo) reach matches `failure_note`, which is also global (`app.failed`, `src/app.rs:281`) — so on this axis, not a divergence.
- *"land the reap side first, then keep that drift test green here"* — `reap::tests::both_queries_ask_for_a_linked_pr_in_exactly_the_same_words` passes; the shared PR selection block is byte-identical and untouched by the diff. Full `cargo test --locked --lib`: 443 passed, 0 failed.

Two non-blocking gaps the spec's *fix shape* arguably reaches but its enumeration does not:

- `src/fetch.rs:65` — sub-issue `labels(first: 10)` stays unguarded. An 11-label ticket whose type label lands on page 2 silently becomes `TicketType::Untyped` (`src/model.rs:324`), changing the launch route. Against *"rather than rendering a silently partial tree"* — but the ticket names only `labels(first: 20)`, the map's own, so this sits outside its literal scope. **Non-blocking.**
- `src/fetch.rs:168` — the query asks `closedByPullRequestsReferences … pageInfo` but the map parse keeps it as `Nodes`, discarding the answer, so a 6th linked PR is still dropped without a trace. PR links are not ticket-bearing, so this is consistent with the ticket rather than in violation. **Non-blocking.**

## Verdict

1. **`src/fetch.rs:922`, `asks_its_own_page_info` asserts the wrong thing.** Escalated to blocking over the Standards axis's own "non-blocking" call, for one reason: seam 2 is a deliverable this PR claims in its own body — *"`MAP_QUERY` asks for every page boundary the parse reads, guarded against the query source the way #132 guards reap's batch"* — and the mutation above shows it does not. The guard admits `pageInfo { endCursor }`, and combined with the two `#[serde(default)]`s that reword restores the exact silent-completeness bug #184 exists to kill, with every test still green. Nothing user-facing is broken today; what is broken is the thing standing between today and the regression. The fix is a one-line strengthening of the assertion, or better, deriving it from `reap::parse_selection`.

Everything else above is a follow-up, not a gate. The Spec axis passes cleanly on all three enumerated consequences and the #199 drift test holds. Two items are worth carrying to #182's **Not yet specified** rather than folding in here: the unguarded sub-issue `labels(first: 10)` → `TicketType::Untyped` misroute, and the count-line clipping guard now that a third named segment sits ahead of the one segment that fragments instead of vanishing.

blooop (commented):

Fresh-eyes re-review at the fixed point `git diff 9c0fdf8...7cb1bec` (three-dot against this PR's own base, `wayfinder/wayfinder-183` — **not** main; #199 is still open). Preflight: ref resolves, diff non-empty (412 insertions across 5 files), all five checks green at head `7cb1bec`. Spec = #184. Two axes below, run independently in parallel, **not** merged or reranked.

The earlier report and the revision commit were read as leads only; every conclusion below — including "the previous blocking finding is fixed" — was re-derived from source and from mutations run here.

## Standards

**Nothing blocking.** The behavioural core is correct and guarded by tests that actually bite. Sixteen mutations were run across the diff and every one that matters went red: neutering either `truncated |=` fold (`src/fetch.rs:384`, `:390`), dropping the `&& !issue.labels.page_info.has_next_page` qualifier (`src/fetch.rs:371`), removing `pageInfo { hasNextPage }` from any of the three connections, shrinking `labels(first: 100)`, a serde-rename typo on either `pageInfo` or `hasNextPage`, dropping the `#[serde(default)]` on `page_info`, `more + 1` → `more`, and unwiring `truncated_note` from the count line.

**The previous round's blocking finding is genuinely closed.** `selected()` (`src/fetch.rs:940`) parses `MAP_QUERY` into the tree it claims — the full 43-path tree was dumped and checked, including the `nodes`/`pageInfo` sibling split and the three `pageInfo.hasNextPage` paths. The claimed catch now bites: rewording any selection to `pageInfo { endCursor }` goes **red**, as does a fragment spread and a shrunk label page. Addressing by path also does what it says about #183's nested linked-PR `pageInfo` — it sits at a different path and cannot answer for a connection that never asked.

**Refuted, recorded as refuted.** The prior round left `truncated_note`'s unscoped `app.clusters` iteration (`src/ui.rs:598`) unreconciled. It is not a divergence: `src/ui.rs:2464` (`a_partial_failure_is_visible_and_survives_the_next_keypress`) deliberately names `blooop/dotfiles#4` while scoped on `blooop/wayfinder`, so the new segment matches the established reach of the segment it is modelled on. Also refuted: the `more + 1` arithmetic (mutation red), and the absent `# Panics` section — `truncated_note` uses `next()`/`count()` rather than `failure_note`'s `expect("len checked")`, so having none is right.

**Non-blocking 1 — Duplicated Code, the one worth fixing.** `Selected` + `selected()` (`src/fetch.rs:895-981`) is a near-verbatim copy of `Selected` + `parse_selection()` (`src/reap.rs:3061-3104`): same grammar, same three match arms, same `expect` strings, same `assert!(path.is_empty(), "unbalanced braces…")`. The fetch copy is a strict superset (argument capture, `at()`, `under()`). This is the drift that `both_queries_ask_for_a_linked_pr_in_exactly_the_same_words` (#132) exists to forbid, one screen up in the same file — *"a copy with nothing holding it can lose a field silently"* — now applied to the guard machinery instead of the query text. `src/probe.rs`'s own header records `code_only` being written twice byte-for-byte before it landed there, which is the documented home for exactly this. Fix proved feasible here: hoisting the fetch version to `#[cfg(test)] pub(crate)`, deleting reap's, and rewriting reap's three `asked.0["x"].0["y"]` index-panics as `at()` lookups leaves 443/443 green, deletes ~45 lines, and gives reap the path lookup it hand-rolls.

**Non-blocking 2 — the guard's claim is stronger than the guard.** `src/fetch.rs:935` says *"which no rewording can answer by accident."* Aliasing is that rewording: `pages: pageInfo { hasNextPage }` and `pageInfo { more: hasNextPage }` both stay **green** while `Paged::page_info` silently defaults — the exact pre-#184 "complete forever" bug. The parser reads `alias: field` as two sibling names, so the real name is always present at the asserted path. Not blocking: nobody has a reason to alias a page boundary, and `reap.rs`'s parser has the same blind spot pre-existing. Softening the sentence would satisfy this.

**Non-blocking 3 — README.** `README.md:495-547` documents every other persistent count-line segment in prose (`· N idle maps hidden`, stalls, reclaimable, *"a fetch that fails says so on the count line"*). The new segment lands with no README line, against AGENTS.md's *"README.md is the real documentation … why the screen is shaped the way it is."*

One clause each, not gated: `Map::truncated: bool` folds "tickets missing" and "edges missing" into one word the UI cannot tell apart, and 30 `truncated: false,` literals across four files are an argument that `Map` wants a `Default`; the `Paged`/`Nodes` split is thin but its doc comment argues the case honestly.

## Spec

**Clean.** No blocking or non-blocking spec findings survived probing. Every finding below quotes #184.

- *"query the map label membership directly (filter by label in the query, or raise the cap well past plausible) so label count can never un-map a map."* The PR took the second branch (`labels(first: 100)`, the GraphQL page maximum) **and** made a truncated page unable to reject (`src/fetch.rs:370-373`). Probed independently of the PR's own test: a fixture with 100 non-map labels and `hasNextPage: true` parses as a map, and the control — a *complete* page without the label — still rejects. The **>100-label case is therefore also closed**: any page that could hide the label is by definition truncated, and truncation forgives. Within the spec's stated latitude.
- *"fetch `pageInfo.hasNextPage` on the ticket-bearing connections"* — probed each path separately against `MAP_RESPONSE`: `subIssues.pageInfo.hasNextPage` alone sets `map.truncated`; one sub-issue's `blockedBy.pageInfo.hasNextPage` alone also sets it (`src/fetch.rs:384`, `:390`). The ticket asks for *reporting*, not pagination, so the still-capped `first: 100` / `first: 50` conform.
- *"say so on the count line when a map arrived truncated, rather than rendering a silently partial tree"* — checked in three realistic frames at the default 90 columns rather than the PR's fixture: a truncated map **in another project** renders as `other/repo#9 truncated`; a truncated map that is **idle and hidden from the body** (`· 1 idle map hidden` confirmed alongside) is still named; and the segment survives a crowded line carrying a failed map and an idle count. Not a test-only segment.
- *"land the reap side first, then keep that drift test green here."* `reap::tests::both_queries_ask_for_a_linked_pr_in_exactly_the_same_words` passes; `src/reap.rs` is untouched by the diff; and the brace-balanced `closedByPullRequestsReferences` block is **byte-identical** between base `9c0fdf8` and head `7cb1bec`. 443 tests pass.
- **Scope**: nothing in the diff the ticket did not ask for — the `model.rs` / `app.rs` / `view.rs` changes are the new field and its fixture updates — and nothing the ticket asked for is missing.

## Verdict

**Approve.** *(Posted as a comment: GitHub refuses `--approve` on a PR authored by the same account. The written verdict is what the gate reads.)*

No blocking findings on either axis. The single blocking finding from the previous round is closed and re-proved closed by mutation from a context that never saw the fix, and the axes' earlier disagreement about `truncated_note`'s reach is now resolved — against the finding, on the precedent at `src/ui.rs:2464`.

Three non-blocking follow-ups, in the order I would take them, none of them a gate on this PR:

1. Unify `fetch::selected` with `reap::parse_selection` into one shared test helper (`src/probe.rs` is the documented home). This is a #132-shaped drift risk in the guard machinery, and the unification is a net deletion.
2. Soften `src/fetch.rs:935`'s *"no rewording can answer by accident"* — aliasing does, in both parsers.
3. Give the new count-line segment a README line, like every other persistent segment has.

Two items were held out of scope by direction and are already on #182's **Not yet specified**: the sub-issue `labels(first: 10)` → `TicketType::Untyped` misroute, and extending the count-line clipping guard to this diff's third named segment.

PR #205, opened 2026-08-22
Robustness sweep: blocking probe, exec bit, backoff, stale prose

### LINKED ISSUE #192: Robustness sweep: blocking probe, exec bit, backoff, stale prose
## Question

Robustness sweep — four small, independent fixes plus the comment drift the review caught:

1. **Blocking subprocess on the async runtime.** `devlaunch_on_path()` uses std `Command::output()` (~90 ms Python start, `launch.rs:1392–1406`) and is reached from async `reap::workspaces` via `devlaunch_answers_unsaved()` (`reap.rs:1016`) — on a tokio worker inside the picker's background survey. Memoized, so one stall; still, wrap in `spawn_blocking` or use tokio's process.
2. **`resolve_on_path` diverges from execvp.** `launch.rs:1967–1977` takes the first `is_file()` PATH candidate without checking the execute bit; execvp would skip it and keep searching. A stray non-executable `claude` early on PATH fails a launch the OS would have made.
3. **Discovery retries forever with no backoff.** `refresh.rs:323–338` respawns a `gh` process every 4s for the whole session even when `gh` is absent or unauthenticated. Add backoff or a give-up after the failure is clearly structural.
4. **Comment drift in load-bearing prose.** `main.rs:588`'s failure message says "five" over a nine-entry guard list (same stale "five" at `main.rs:449`); `model.rs:280` says "none of the four types" where there are five; `picker.rs:551` says the session record is written "immediately before the terminal is restored" when it is written after `ratatui::restore()`. In a repo whose comments are load-bearing, a guard's red message that miscounts its own list misdirects the reader.

Also fold in the `link_state` cosmetic: a failed `read_link` maps to `Link::Stale(PathBuf::new())` (`skills.rs:393`), producing a message that names an empty path.

blooop:
Closes #192. Five independent fixes from the 0.19.2 review's robustness sweep, one test-first slice each:

1. **Blocking probe off the runtime worker.** `devlaunch_on_path()` waits ~90ms for a Python interpreter with std's `Command::output()`, and `reap::workspaces` reached it from a tokio worker inside the picker's background survey. The answer is now awaited through an off-thread seam (`spawn_blocking`), tested by the observable: a current-thread runtime's timers keep firing while a blocking call made through the seam runs.

2. **PATH resolution matches execvp.** `resolve_on_path` took the first `is_file()` hit, so a stray non-executable `claude` early on PATH failed a launch the OS would have made. Resolution now requires an execute bit (the same approximation `which(1)` makes), with the rules split from the one line that reads the environment so tests never mutate PATH under a parallel suite.

3. **Discovery backs off.** `spawn_discovery` respawned a `gh` process every 4s for the whole session even when the failure was structural. The wait now doubles per consecutive miss and settles at sixteen times the old cadence — a blip still recovers on the first quick retry, a mid-session `gh auth login` is noticed within the minute, and a broken `gh` costs one process a minute instead of fifteen. Pure schedule function with literal expected values, worded clear of refresh.rs's source-text denylist.

4. **Prose drift.** The reap guard's red message said "five" over a nine-entry list (and its sibling comment made the same stale count); model.rs said "four types" where `wf` knows five; picker.rs claimed the session record is written before the terminal is restored when it is written just after. Counts are now count-free so a list edit cannot re-stale them.

5. **No empty path in the skills report.** A failed `read_link` was carried as `Stale(PathBuf::new())` and printed "stale — links to " with nothing after it. `Stale` now carries `Option<PathBuf>`; the report says the target could not be read, the sweep declines to touch a link it cannot prove it wrote, and install still replaces it.

Full local chain green (fmt, clippy, lib/bins/examples, skill_docs, devcontainer_prebuild, toolchain_pin, offline_green, live_fetch common, doc) under `-D warnings`.

## Summary by Sourcery

Improve launch, discovery, and skills robustness while keeping runtime responsiveness, retry behavior, and status reporting reliable.

New Features:
- Add asynchronous offloading for blocking interpreter probes so runtime workers remain responsive.
- Make PATH resolution skip non-executable files and improve stale-link reporting when symlink targets cannot be read.

Bug Fixes:
- Prevent discovery from repeatedly respawning failed processes at a high frequency by applying bounded retry backoff.
- Correct stale counts and ordering claims in user-facing and maintenance prose.

Enhancements:
- Preserve safe replacement of unreadable stale links while avoiding unsupported ownership claims during refresh.

Tests:
- Add coverage for non-executable PATH candidates, runtime responsiveness during blocking work, bounded discovery backoff, and unreadable stale-link status output.

### DIFF
--- a/src/launch.rs
+++ b/src/launch.rs
@@ -1955,7 +1955,12 @@ impl Launch {
     }
 }
 
-/// Find `program` on `$PATH`, skipping empty entries.
+/// Find `program` on `$PATH`, skipping empty entries and any candidate the OS
+/// would not execute.
+///
+/// The execute bit matters because this stands in for execvp, which keeps
+/// searching past a file it cannot run — so a stray non-executable `claude`
+/// early on PATH must not fail a launch the OS would have made.
 ///
 /// A name containing a separator is a path already and is taken as given —
 /// that is the caller naming a file, not `$PATH` resolution.
@@ -1965,15 +1970,33 @@ impl Launch {
 /// wants only a path to run `--version` on and treats a miss as "no `dl` here",
 /// which [`Isolation::detect`] answers with [`Isolation::Host`].
 fn resolve_on_path(program: &str) -> Result<PathBuf, anyhow::Error> {
+    resolve_in(program, &std::env::var_os("PATH").unwrap_or_default())
+}
+
+/// [`resolve_on_path`] against a given `PATH` value — the part with rules worth
+/// testing, split from the one line that reads the environment so the tests
+/// never have to mutate it under a parallel suite.
+fn resolve_in(program: &str, path: &std::ffi::OsStr) -> Result<PathBuf, anyhow::Error> {
     if program.contains('/') {
         return Ok(PathBuf::from(program));
     }
-    let path = std::env::var_os("PATH").unwrap_or_default();
-    std::env::split_paths(&path)
+    std::env::split_paths(path)
         .filter(|dir| !dir.as_os_str().is_empty())
         .map(|dir| dir.join(program))
-        .find(|candidate| candidate.is_file())
+        .find(|candidate| executable(candidate))
         .ok_or_else(|| anyhow::anyhow!("`{program}` is not on PATH — is it installed?"))
+}
+
+/// Would execvp accept this candidate? A regular file with any execute bit
+/// set. Mode bits rather than `access(2)`'s exact answer, which is the same
+/// approximation `which` makes — close enough for a search whose miss is
+/// reported, and whose false positive still fails loudly at the `exec` itself.
+fn executable(candidate: &Path) -> bool {
+    use std::os::unix::fs::PermissionsExt;
+    candidate
+        .metadata()
+        .map(|meta| meta.is_file() && meta.permissions().mode() & 0o111 != 0)
+        .unwrap_or(false)
 }
 
 /// A launch prompt with the JSON of every `ctx:` block replaced by `…` (#124).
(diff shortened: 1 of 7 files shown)


blooop (commented):

Fresh-eyes review in a context that never saw this code written. Fixed point: `git diff 4135446...50bfef0` — 7 files, +241/−50. All checks green at 50bfef0. Both axes ran as independent parallel subagents against a clean PR-head checkout and are reported unmerged and unranked against each other.

## Standards

**Nothing blocking.** The two extractions are honest rather than indirection: `resolve_in` genuinely removes an environment mutation from the tests, and `state_line` makes the new arm testable without arranging an unreadable symlink — neither creates a second switch on the same type. `executable()`'s `0o111` check is correct (follows symlinks, right answer for root, the same approximation `which(1)` makes, documented as such). `retry_delay`'s arithmetic and its clamp check out — 4/8/16/32/64/64/64 confirmed, `retry_delay(0)` unreachable because the `Ok` arm returns — and its tests pin the schedule with literals rather than restating the expression. `Stale(Option<_>)` changes no `refresh` behaviour: `is_ours(Path::new(""))` was already `false` via `parent() == None`, so the unreadable case already fell to the `continue`. All matches are exhaustive.

1. **`src/refresh.rs:46` — non-blocking, introduced by this diff.** `RETRY_INTERVAL`'s own doc still reads "How long the map search waits before trying again." After this change it is only the *first* wait and the base of a doubling schedule; the backoff is documented on `spawn_discovery` and `retry_delay` only. In a PR whose subject is load-bearing prose being true, the constant is where a reader looks the cadence up.

2. **`src/skills.rs:559` + `src/main.rs:293` — non-blocking, introduced by this diff.** The sentinel leaves `Link::Stale`/`State::Stale` but reappears in `Outcome::Linked { was: Option<PathBuf> }`: `Stale(None)` now installs as `Linked { was: None }`, which `wf skills install` prints as `"linked"` — a claim that no link existed, where one existed and was just deleted. The status line got the honest wording and the install line got a false one; it previously printed `relinked (was )` — empty, but not untrue. (`Healed` is unaffected: `refresh`'s `Stale(_)` arm `continue`s.) This is the same message-honesty class as the map's already-open **`wf skills install` reporting honesty** item, which is where it most naturally belongs rather than as a re-open of this ticket.

3. **`src/skills.rs:77` and `src/skills.rs:513` — non-blocking, pre-existing.** Two live stale counts against `BUNDLED: [&str; 6]` (`skills.rs:58`), in the file this PR edits and in exactly finding 4's defect class: line 77 says a copy under `skills/` "would register **five** more skills" (six), and line 513 says one bad skill never costs "the other **four**" their install (five). Not named by the ticket, so out of its literal scope — but cheap to fold in if the branch is touched again, and they would re-stale the same way the "five" just fixed did.

4. **`src/launch.rs:1440` — nit.** `off_runtime` is a generic one-caller wrapper over `spawn_blocking`; the type parameters serve one production call plus its own test. `.expect("the blocking call does not panic")` also turns any `JoinError` into a panic on a `reap` path whose every other failure is a `Result` — `unwrap_or(false)` would match `answers_unsaved`'s documented permissive-on-unknown posture.

5. **`src/skills.rs:650` — nit.** "the sweep has no permission to touch it" is written inside `refresh`. It borrows `sweep`'s ownership test, but `sweep` is a different function (`skills.rs:753`), and the PR description repeats the same conflation.

## Spec

Ticket #192 is a sweep of five independent fixes, so each is checked on its own rather than against the headline.

| # | Ticket requirement | Verdict |
|---|---|---|
| 1 | "wrap in `spawn_blocking` or use tokio's process" | partially landed |
| 2 | "takes the first `is_file()` PATH candidate without checking the execute bit; execvp would skip it and keep searching" | landed |
| 3 | "Add backoff or a give-up after the failure is clearly structural." | landed |
| 4 | the four named prose sites | landed — all four re-verified true |
| 5 | "a failed `read_link` maps to `Link::Stale(PathBuf::new())` … producing a message that names an empty path" | landed |

- **2** — `resolve_in` now finds on `executable(c)`. The "keep searching" half is proved, not just the failure half: `path_resolution_skips_a_candidate_the_os_would_not_execute` puts a `0o644` `agent` in the *first* PATH entry and a `0o755` one in the second and asserts the second is returned. Verified red-before-green: the old `is_file()` predicate returns the decoy.
- **3** — `retry_delay(f) = RETRY_INTERVAL * 2^(f−1)` capped at ×16; the sleep is only reached with `failures >= 1`, so the first retry is the old cadence. Backoff satisfies the "or".
- **4** — `main.rs:588` and `:449` are count-free over a list independently counted at nine; `model.rs:280` "four types" → "the types `wf` knows" (five plus `Untyped`; `model.rs:253`/`:255` already said five/six correctly); `picker.rs:551`'s new ordering claim is true — `ratatui::restore()` at `picker.rs:526`, `cache.record_session` at `:559`, `launch.exec` last.
- **5** — user-visible status line is now "stale — the link's target could not be read"; no empty path. (See Standards 2 for the sibling install line.)

**Non-blocking — finding 1 is only fixed on the path the ticket named.** The spec says "`devlaunch_on_path()` uses std `Command::output()` … and is reached from async `reap::workspaces` via `devlaunch_answers_unsaved()`", and asks to "wrap in `spawn_blocking`". Only that seam moved off-thread. `devlaunch_on_path()` is still called synchronously from `Isolation::detect` (`launch.rs:1151`) and `Launch::describe` (`launch.rs:1740`), both reached from `picker.app.handle_key(key)` (`picker.rs:404`) — a sync call on the async loop — via `launch::plan`/`plan_create`/`prewarm` (`app.rs:1000`/`1056`). If an `enter` beats the background survey to the probe, the ~90 ms subprocess runs on the runtime thread anyway; if it races it, `OnceLock::get_or_init` blocks the runtime thread on the survey's probe instead. Wrapping the probe itself would need `Isolation::detect` and `describe` to become async — materially bigger than this slice — so wrapping the outer helper is the achievable fix on the path the ticket cited, and the residual belongs on the map rather than in this PR.

**Nit** — `refresh.rs:350` claims a mid-session `gh auth login` "is still found within the minute" while the ceiling the code produces is 64 s.

**No scope creep.** Both refactors exist only to make the ticket's rules testable. Skills semantics are otherwise preserved: `install` already removed the link for every `Stale` (`PathBuf::new()` was still `Some`), and `refresh` already skipped the unreadable case.

## Verdict

**Comment.** No blocking findings on either axis. All five of the ticket's items landed, each verified against the code rather than the PR description, and the four new tests pass in isolation (verified red-before-green where the predicate allows it).

Two items are worth a decision before merge, neither a gate:
- Standards 1 and 5 and the Spec nit are three lines of prose introduced by this diff that are now inaccurate — in a ticket whose fourth finding is exactly that. Cheapest to fix here.
- Standards 2 is a real regression, but it is the map's already-open **`wf skills install` reporting honesty** item, deliberately kept out of #198; routing it there is more honest than re-opening this ticket for it.
- Standards 3 (two stale `BUNDLED` counts) is pre-existing and out of the ticket's literal scope; fold in only if the branch is touched anyway.

